Issue #56, opened 2015-09-09
cpu_only mode unnecssary copy


tnowotny:
At the moment one needs to make a differently named copy from the .cu file(s) so that the compiler is not confused by the suffix.
So, in cpu_only mode, one, for example, needs to create something like classol_sim_cpu_only.cc from classol_sim.cu to get this done. Note, it has to be different in suffix and base name. This should be improved in the future so that "make cpu_only" suffices to sort this out.

Sayantan2048:
Hello sir, the cpu_only mode currntly depends on CUDA API. For that reason, users without cuda toolkit cannot run genn using cpu_only. As part of my GSoC proposl, should we make a CPU only mode that runs natively on CPU without CUDA/OpenCL? Also native CPU mode opens more possiblites for optimizations on hardware like Intel MIC.

tnowotny:
Hi Sayantan,
CPU only mode has always been intended to be CUDA independent. In my tests that was successful but I may have missed something. Where is the remaining CUDA dependence when CPU_only mode is compiled?

Sayantan2048:
Hello Sir,
I removed CUDA_PATH and then I ran the following userproject from project directory:
`make clean && make
./generate_run 0 100 1000 20 100 0.0025 outname MBody_delayedSyn 0 FLOAT 0
`
I got the following error message i.e the first error reported by compiler:
`/home/sayantan/genn/lib/include/global.h:32:26: fatal error: cuda_runtime.h: No such file or directory`

I went through the control flow and found that the code generation is CUDA dependent for CPU and GPU. For examle chooseDevice() function in generateALL cannot be compiled without CUDA host APIs. 

However, on a remote machine with properly installed CUDA and CUDA_PATH unharmed, I was able to run the above model on CPU. I looked into the generated neuronFnct.cc and runner.cc and found that runner.cc had some dependence on CUDA host API functions.

So, to sum up, it looks like both code generation and generated code are CUDA host API dependent even for CPU.

tnowotny:
Hi - I am sorry for the confusion but what you get when using the generate_run examples and choosing execution on the CPU is never the CPU_ONLY mode that this issue refers to. I know it's a bit confusing but the philosophy of GeNN has always been that both CPU and GPU code are available at compile time and the user is free to program their user side code to use either or both (which the program invoked by generate_run does).
[...]

PR #65, opened 2016-03-13
do not try linking CUDA libraries when CUDA_PATH is not set.

Sayantan2048:
Hello sir,

1. This patch resolves linking errors when trying to run in CPU_ONLY mode with no CUDA toolkit installed.
2. On non-windows build, this patch allows MBody_delayedSyn_project and MBody1_project to run without using GENN_PATH and PATH. Hence these two projects can run from more than one genn directories. 
3. Also fix merge conflicts with 7e1a99ec6459a03f6e9.
4. 0ee1bf3c3d98 solves issue #56.


jamesturner246:
Thank you for your contributions  Sayantan2048.

I've merged them into the dev branch, ready for next release. I made a few modifications for portability and maintenance reasons, though.

For example, the GENN_PATH auto-detection script unfortunately does not work on the MAC and Win platforms (option '-f' is only available on GNU's readlink). I've left one of these checks in the buildmodel.sh script, but, for consistency, we will leave it to the user do define their environment variables (there's only 3 to define!). I moved the classol_sim_cpu_only.cc copy code into the pojects' makefiles, rather than the main shared makefile. 

All the best,
James.

Sayantan2048:
@jamesturner246 
Hello sir, 

The idea behind autodetecting GENN_PATH was to enable working from multiple genn directories. Are there any other non-invasive solution to the problem so that I don't have to change GENN_PATH and PATH when I work from more than one genn directories ?

Regards,
Sayantan

jamesturner246:
You should only need one GeNN directory for all work really... A user's own project can be in any directory, say $HOME/my_project, and they all include the same includes from the same GeNN tree. The only use-case I can think of for having multiple GeNN trees is if you're an active developer. 

In my post in your new thread ( https://github.com/genn-team/genn/pull/67 ), I suggest the use of environment 'modules', where switching GENN_PATH and PATH etc. can be achieved using, for example:

module switch [genn_module_1] [genn_module_2]

Would that be suitable for you?

All the best,
James.

PR #66, opened 2016-03-17
Error 53 fixed:Small DT and single precision floating point

shailesh210:
Tried to fix Issue#53: Small DT and single precision floating point
Thank you


jamesturner246:
Thanks shailesh210.

Changes are merged, except .gitignore. Reason being its nice to be able to clean the whole GeNN tree by running 'git clean -fd"!

All the best,
James.

PR #67, opened 2016-03-20
cpu_only mode autocopy for all projects.

Sayantan2048:


jamesturner246:
Thanks again Sayantan2048.

Re. your idea of auto-detecting genn environment variables, I was thinking we could have a Linux module file (see link below) which can be loaded and unloaded by the user at will, which does the auto-detect logic automatically on load, rather than having lots of \#ifdef GENN_PATH code blocks.

https://sourceforge.net/p/modules/wiki/FAQ/

What does everybody else think?

Cheers,
James.

jamesturner246:
Merged.

PR #108, opened 2016-10-20
Parallel code with OpenMP

PabloGN:
I have done some changes for the parallelisation of the code, mainly when using a Raspberry Pi 3, of course without CUDA.
A (not so) strange behaviour when benchmarking in a i7 CPU_ONLY is that the caches perform a lot better when in monothread.

Nevertheless, on RPi3 the system is faster by far, which is what I wanted :-)
To disable OpenMP remove the `-fopenmp` flag from `lib/src/generateRunner.cc` and `userproject/Izh_sparse_project/model/GNUmakefile`

Thank you!!!!!
### Benchmarks

**Raspberry Pi 3:**

NO OpenMP:
    1600 neurons in total, 100s simulation, Elapsed time is 114.0927 
With OpenMP:
    1600 neurons in total, one hundred seconds, Elapsed time is 92.1728

**PC i7:**
    MonoP:
        14k neurons in total, 10k sinapses Elapsed time is 175.2692     (100 seconds)  
    MP (4 cores):
        14k neurons in total, Elapsed time is 342.4771
    MP (8 cores):
        14k neurons in total, Elapsed time is 458.3932

### DIFF
--- a/lib/src/generateCPU.cc
+++ b/lib/src/generateCPU.cc
@@ -77,7 +77,6 @@ void genNeuronFunction(NNmodel &model, //!< Model description
 	string queueOffsetTrueSpk = (model.neuronNeedTrueSpk[i] ? queueOffset : "");
 
 	os << "// neuron group " << model.neuronName[i] << ENDL;
-	os << OB(55);
 
 	// increment spike queue pointer and reset spike count
 	if (model.neuronDelaySlots[i] > 1) { // with delay
--- a/lib/src/generateCPU.cc
+++ b/lib/src/generateCPU.cc
@@ -615,12 +615,12 @@ void genSynapseFunction(NNmodel &model, //!< Model description
 	}
     }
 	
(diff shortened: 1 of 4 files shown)


jamesturner246:
Hi Pablo. Thanks for contributing.

It's interesting that single-thread performs so much better on the i7. Could it be that there is more chance of a cache hit if just a single thread works on a few consecutive cached pages, rather than many threads working on many non-consecutive cache pages, which are fighting for space?

We should not merge it with the changes enabled by default if it has such an adverse performance effect on a common desktop CPU. I think there should be an option to enable it, if needed, using the `GENN_PREFERENCES` mechanism or similar. I will look closer later.

PabloGN:
Yes, the bottleneck might be the RAM memory access if one thread tries to get data updated by another one.
Yes, after the PR I saw the preferences mechanism. Sorry for that.
Thank you...

PR #118, opened 2017-01-31
Convert soft tabs

neworderofjamie:
First and least exciting change - to convert all soft-tabs to spaces.


tnowotny:
We may need to consider to merge this also into the development branch.
Traditionally, we have used the development branch for updates and have left master fairly stable between releases...

PR #119, opened 2017-01-31
use getconf LONG_BIT rather than uname in makefiles

neworderofjamie:
@tnowotny - can you see if  ``getconf LONG_BIT`` works on mac? It should return 64 or 32


tnowotny:
getconf LONG_BIT seems to work in the shell - so probably alright.

tnowotny:
Merge into development?

PR #121, opened 2017-02-06
Code tidy

neworderofjamie:
- General pass on const-correctness i.e. strings that aren't modified are passed around as const references and NModel is passed as a const reference to generators - IMO this is a good way of preventing future bugs in C++
- The ``tS``/``toString`` conversions were entirely unnecessary in a lot of places - std::string has free function + operators to concatenate ``const char *`` and ``std::string`` in both 'directions' (I have left these functions in as some userProjects use them)
- ``std::string`` has a ``empty()`` method which is much more efficient than comparing with ``""`` (let alone ``tS("")``
- Replacement of ``uname -m`` with ``getconf LONG_BIT`` in Makefile so whether *kernel* is 32-bits or not rather than CPU is tested (previously merged into master)
- Replacement of soft-tabs with 4 spaces (previously merged into master)

These changes have been tested on Windows (Visual Studio 2013) and Linux (ARM and x86_64) using the W.I.P. google tests and various examples

### DIFF
--- a/SpineML_to_GeNN/extra_postsynapses.h
+++ b/SpineML_to_GeNN/extra_postsynapses.h
@@ -1,9 +1,9 @@
 
 
-	postSynModel ps;
+  postSynModel ps;
 
 
-			
+
   // Add new postsynapse type - passthroughPS: 
   ps.varNames.clear();
   ps.varTypes.clear();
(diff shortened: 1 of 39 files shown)


tnowotny:
Looking good ... happy to see things getting streamlined.

PR #122, opened 2017-02-08
New model API

neworderofjamie:
So first of all this pull request is not really ready to be merged - it's more as the basis for discussion.

When I was talking to Thomas about stuff I could do, I mentioned that I thought the API for defining models and instantiating them could be improved. In order to demonstrate what I was thinking of, I've had a go at a new API and updated the CPU generator so they can be used. As a top-level example the following is how you might define a LIF neuron in the current system:

```c++
class ClosedFormLIF : public dpclass
{
public:
    virtual double calculateDerivedParameter(int index, vector<double> pars, double dt = 1.0)
    {
        switch (index)
        {
        case 0: // ExpTC
            return std::exp(-dt / pars[1]);
        case 1: // Rmembrane
            return pars[1] / pars[0];

        }
        return -1;
    }
};

const unsigned int MY_LIF = nModels.size();
nModels.push_back(neuronModel());

nModels.back().varNames.reserve(2);
nModels.back().varNames = {"V", "RefracTime"};

nModels.back().varTypes.reserve(2);
nModels.back().varTypes = {"scalar", "scalar"};

nModels.back().pNames.reserve(7);
nModels.back().pNames =
{
  "C",          // Membrane capacitance
  "TauM",       // Membrane time constant [ms]
  "Vrest",      // Resting membrane potential [mV]
  "Vreset",     // Reset voltage [mV]
  "Vthresh",    // Spiking threshold [mV]
  "Ioffset",    // Offset current
  "TauRefrac",  // Refractory time [ms]
};

nModels.back().dpNames.reserve(2);
nModels.back().dpNames = {"ExpTC", "Rmembrane"};
nModels.back().dps = new ClosedFormLIF();

nModels.back().simCode =
  "if ($(RefracTime) <= 0.0)\n"
  "{\n"
  "  if ($(V) >= $(Vthresh))\n"
  "  {\n"
  "    $(V) = $(Vreset);\n"
  "    $(RefracTime) = $(TauRefrac);\n"
  "  }\n"
  "  scalar alpha = (($(Isyn) + $(Ioffset)) * $(Rmembrane)) + $(Vrest);\n"
  "  $(V) = alpha - ($(ExpTC) * (alpha - $(V)));\n"
  "}\n"
  "else\n"
  "{\n"
  "  $(RefracTime) -= DT;\n"
  "}\n";
 nModels.back().thresholdConditionCode = "$(RefracTime) <= 0.0 && $(V) >= $(Vthresh)";
```
In the new approach this definition would look like:

```c++
class ClosedFormLIF : public NeuronModels::Base
{
public:
    DECLARE_MODEL(ClosedFormLIF, 7, 2);

    SET_SIM_CODE(
        "if ($(RefracTime) <= 0.0)\n"
        "{\n"
        "  if ($(V) >= $(Vthresh))\n"
        "  {\n"
        "    $(V) = $(Vreset);\n"
        "    $(RefracTime) = $(TauRefrac);\n"
        "  }\n"
        "  scalar alpha = (($(Isyn) + $(Ioffset)) * $(Rmembrane)) + $(Vrest);\n"
        "  $(V) = alpha - ($(ExpTC) * (alpha - $(V)));\n"
        "}\n"
        "else\n"
        "{\n"
        "  $(RefracTime) -= DT;\n"
        "}\n"
    );

    SET_THRESHOLD_CONDITION_CODE("$(RefracTime) <= 0.0 && $(V) >= $(Vthresh)");

    SET_PARAM_NAMES({
        "C",          // Membrane capacitance
        "TauM",       // Membrane time constant [ms]
        "Vrest",      // Resting membrane potential [mV]
        "Vreset",     // Reset voltage [mV]
        "Vthresh",    // Spiking threshold [mV]
        "Ioffset",    // Offset current
        "TauRefrac"});

    SET_DERIVED_PARAMS({
        {"ExpTC", [](const vector<double> &pars, double dt){ return std::exp(-dt / pars[1]); }},
        {"Rmembrane", [](const vector<double> &pars, double dt){ return  pars[1] / pars[0]; }}});

    SET_INIT_VALS({{"V", "scalar"}, {"RefracTime", "scalar"}});
};

IMPLEMENT_MODEL(ClosedFormLIF);
```
I think:
- This is clearer (although that is a matter of personal taste/opinion)
- It's terser 
- If you want to base a new model on an existing one you can just inherit off it
- If you are defining your own models  you don't need to mess with GeNN globals to do so

Also when it comes to instantiating your model, previously you might do:

```c++
// LIF model parameters
double lifParams[7] =
{
  1.0,    // 0 - C
  20.0,   // 1 - TauM
  -70.0,  // 2 - Vrest
  -70.0,  // 3 - Vreset
  -51.0,  // 4 - Vthresh
  0.0,   // 5 - Ioffset
  2.0,    // 6 - TauRefrac
};

// LIF initial conditions
double lifInit[2] =
{
  -70.0,  // 0 - V
  0.0,    // 1 - RefracTime
};

model.addNeuronPopulation("Excitatory", 1, MY_LIF, lifParams, lifInit);
```

Now you would do:

```c++
// LIF model parameters
auto lifParamVals = ClosedFormLIF::ParamValues(1.0, 20.0, -70.0, -70.0, -51.0, 0.0, 2.0);
auto lifInitVals = ClosedFormLIF::InitValues(-70.0, 0.0);

model.addNeuronPopulation<ClosedFormLIF>("Excitatory", 1, lifParamVals, lifInitVals);
```
Which allows the compiler to check you're passing in the correct number of arguments (I've also made this change backward compatible, allowing existing code to be used unchanged). Before I invest much more time in updating the cuda code generator; and the post-synaptic and weight update models to support this, I figure it's probably worth seeing what you think!


tnowotny:
Hi Jamie,
looks like an impressive amount of work in short time but we should get together and discuss what problem exactly you are trying to solve and how we continue ... so far I looked only at the examples above, which look interesting but I also would be interested in how it looks inside and in what sense it is better...

neworderofjamie:
That would be great - let me know when you have some time.

neworderofjamie:
What do you think about merging this @tnowotny (I'll do it manually as I accidentally merged the fix for #129 into development rather than master)? that way I can start working on stuff like the zero-copy as feature branches off development rather than a mess of branches of branches...

tnowotny:
Ok, I am happy for this to be merged into development. The fix to #129 should probably be merged into master?

neworderofjamie:
Great - will do that. Already merged #129 into master as well

PR #123, opened 2017-02-13
Fix variable scopes

tnowotny:
Here is the fix for the incorrect local variable scopes in kernel functions. It runs fine for all existing tests and examples but none of them is using this particular piece of code. Jamie, can you add a few tests that exercise "support code" and test this?
@mstimberg, this should solve the problems for brian2genn and we should be able to split the threshold conditions as discussed in brian-team/brian2genn#41.

### DIFF
--- a/lib/bin/genn-buildmodel.sh
+++ b/lib/bin/genn-buildmodel.sh
@@ -64,12 +64,6 @@ fi
 
 # generate model code
 make -f "$GENN_PATH/lib/GNUmakefile" $MACROS
-# The following became necessary because from Mac OSX El Capitan, SIP (System Integrity Protection) squashes DYLD_LIBRARY_PATH in sub-shells and libcudart.8.dylib has a @rpath based install path which necessitates to add the cuda path to DYLD_LIBRARY_PATH.
-unamestr=`uname`
-# if [[ "$unamestr" == "Darwin" ]]; then
-#     export DYLD_LIBRARY_PATH=$DYLD_LIBRARY_PATH:$CUDA_PATH/lib
-# fi
-# end of workaround
 if [[ -n "$DEBUG" ]]; then
     gdb -tui --args "$GENERATEALL" "$OUT_PATH"
 else
--- a/lib/bin/genn-buildmodel.sh
+++ b/lib/bin/genn-buildmodel.sh
@@ -64,6 +64,12 @@ fi
 
 # generate model code
 make -f "$GENN_PATH/lib/GNUmakefile" $MACROS
+# The following became necessary because from Mac OSX El Capitan, SIP (System Integrity Protection) squashes DYLD_LIBRARY_PATH in sub-shells and libcudart.8.dylib has a @rpath based install path which necessitates to add the cuda path to DYLD_LIBRARY_PATH.
+unamestr=`uname`
+# if [[ "$unamestr" == "Darwin" ]]; then
+#     export DYLD_LIBRARY_PATH=$DYLD_LIBRARY_PATH:$CUDA_PATH/lib
+# fi
+# end of workaround
 if [[ -n "$DEBUG" ]]; then
     gdb -tui --args "$GENERATEALL" "$OUT_PATH"
 else
--- a/lib/src/generateKernels.cc
+++ b/lib/src/generateKernels.cc
@@ -306,17 +306,39 @@ void genNeuronKernel(const NNmodel &model, //!< Model description
 
         // look for spike type events first.
         if (model.neuronNeedSpkEvnt[i]) {
-            string eCode = model.neuronSpkEvntCondition[i];
-            // code substitutions ----
-            substitute(eCode, "$(id)", localID);
-            substitute(eCode, "$(t)", "t");
-            extended_name_substitutions(eCode, "l", nModels[model.neuronType[i]].varNames, "_pre", "");
-            name_substitutions(eCode, "", nModels[model.neuronType[i]].extraGlobalNeuronKernelParameters, model.neuronName[i]);
-            eCode= ensureFtype(eCode, model.ftype);
-            checkUnreplacedVariables(eCode, "neuronSpkEvntCondition");
-            // end code substitutions ----
-            os << "// test for and register a spike-like event" << ENDL;
-            os << "if (" + eCode + ")" << OB(30);
+            // Create local variable
+            os << "bool spikeLikeEvent = false;" << ENDL;
+
+            // Loop through outgoing synapse populations that will contribute to event condition code
+            for(const auto &spkEventCond : model.neuronSpkEvntCondition[i]) {
+                // Replace of parameters, derived parameters and extraglobalsynapse parameters
+                string eCode = spkEventCond.first;
+
+                // code substitutions ----
+                substitute(eCode, "$(id)", "n");
+                substitute(eCode, "$(t)", "t");
+                extended_name_substitutions(eCode, "l", nModels[model.neuronType[i]].varNames, "_pre", "");
+                name_substitutions(eCode, "", nModels[model.neuronType[i]].extraGlobalNeuronKernelParameters, model.neuronName[i]);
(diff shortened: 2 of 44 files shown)


mstimberg:
Hi Thomas. This looks good to me but it does not quite solve our brian2genn problem yet: I think you applied the changes only to the CUDA kernel, not to the CPU function. We always compile both, don't we? Either way, we'd of course want to make it work in CPU-only mode as well.

lib/bin/genn-buildmodel.sh
```diff
@@ -64,6 +64,12 @@ fi
 
 # generate model code
 make -f "$GENN_PATH/lib/GNUmakefile" $MACROS
+# The following became necessary because from Mac OSX El Capitan, SIP (System Integrity Protection) squashes DYLD_LIBRARY_PATH in sub-shells and libcudart.8.dylib has a @rpath based install path which necessitates to add the cuda path to DYLD_LIBRARY_PATH.
+unamestr=`uname`
```
neworderofjamie: Did you figure out a better fix for this or is it commented out for another reason?

lib/src/generateKernels.cc
```diff
@@ -261,12 +257,18 @@ void genNeuronKernel(const NNmodel &model, //!< Model description
...
+            }
+            os << "Isyn += " << psCode << ";" << ENDL;
             if (!psm.supportCode.empty()) {
                 os << CB(29) << " // namespace bracket closed" << ENDL;
             }
         }
 
+	if (!nModels[nt].supportCode.empty()) {
+	    os << " using namespace " << model.neuronName[i] << "_neuron;" << ENDL;
+	}
```
neworderofjamie: I could be missing something, but are you ever closing this namespace?

neworderofjamie: Am I missing something or is this namespace not getting closed (I guess it should be closed around line 365)?

tnowotny:
Yes, I figures out a better way ... I should fully remove it (I thought I did).
The better way is a -rpath options when linking.

tnowotny:
wait with merging the pull request until I have also attempted a fix for the scoping problem in the CPU side of things ... forgot to do this yesterday.

tnowotny:
it is automatically restricted to the scope of the enclosing if statement. But I am happy to be proven wrong.

neworderofjamie:
Yeah - you're totally right - I just forgot how using worked...

tnowotny:
ok ... I took it as far as I could in the time I have - but still not tests covering this. @mstimberg if you want to try it - please do ... it is entirely possible that I have overlooked something and/or removed a bracket too many. @neworderofjamie - are you able to put in some tests? Should we advance the google style testing so you can do it in that framework rather than adding to the older feature tests?

neworderofjamie:
I was thinking the same - I'll merge this branch into the googletests one
and have a think about tests that cover variables across the various
combinations of scopes.

mstimberg:
At least from the brian2genn side I can say that it works: with these changes in GeNN, the branch `fix_threshold_#41` in brian2genn now correctly deals with thresholds of all sorts (referring to subexpressions or not), both on the CPU and the GPU.

tnowotny:
Hi Jamie – the important bit is to use some support code for the different code strings. Can be as simple as wrapping “exp” into “myexp” as support code. This will then exercise the whole code generation path around support codes.
Thanks,
Thomas

--
Prof Thomas Nowotny
Director of Research and Knowledge Exchange
School of Engineering and Informatics
University of Sussex
Falmer, Brighton BN1 9QJ, UK
Phone: +441273678593
FAX: +441273877873

tnowotny:
Nice one, Marcel! We will merge this into (at least) development (if not master) soon. Hope it's ok to work with this branch for the moment?

neworderofjamie:
Ahh ok - that is simple then! I thought we had to do that and use local
variables across different code blocks e.g between the sim code and the
reset code.

mstimberg:
Sure, no problem. My `fix_threshold_#41` branch is also on top of the branch to add support for scalar variables, so there is a bit of branch-juggling involved either way ;)

tnowotny:
@neworderofjamie , the main thing is just to use support code everywhere. Thinking of it, might be worth just injecting some support code into  COPY OF all the existing google tests, which are already testing some cross-variable use quite nicely.

neworderofjamie:
I have merged this branch into the googletests branch and added tests to test access to support code from all appropriate bits of user code. With the fixes made in this branch all combinations work apart from one (unpleasant) one:

If ``weightUpdateModel::evntThreshold`` refers to functions defined in ``weightUpdateModel::simCode_supportCode`` it breaks. This is because, in ``NNmodel::initLearnGrps``, the ``evntThreshold`` code for each outgoing synapse populations is added to ``NModel::neuronSpkEvntCondition`` and then in the neuron function code generator these tests are combined back together into the test for spike-like events. I guess solutions might include:

1.  Also generate a list of namespace names for support code alongside ``NModel::neuronSpkEvntCondition`` and in the neuron function run each spike-like event test within a scope where that namespace is used.
2. Remove``NModel::neuronSpkEvntCondition`` entirely and generate the spike-like event detection code based on ``NModel::outSyn``in the neuron function generator (again performing each spike-like event test within a scope where the namespace is used)

tnowotny:
I suppose the second suggestion is somewhat more logical. A second question would be whether there should be a separate ```weightUpdateModel::evntThreshold_supportCode``` or whether the ```evntThreshold``` code should share the support code with ```simCode```?

neworderofjamie:
Ok, I will have a go at implementing that if you would like?

I was sort of wondering about the converse actually...How come ``weightUpdateModel`` has separate support code slots for each (well not quite) code block at all? ``neuronModel`` and ``postSynModel`` both just have one support code slot and, as it's intended for defining functions, that makes more sense to me...

tnowotny:
Hum ... to be honest with you I cannot recall why I did one thing one way and the other the other way. It is quite unlikely that within the same neuron, weightupdate, or post-synaptic model, two code strings have a conflict of needing support code of the same name with different meaning.
In other words, I think the neuron and postsynapse approach is right and the weightupdatemodel should only have one support code namespace as well....
if you have the time to clean this up, that'd be great!

neworderofjamie:
Ok will do - unless you have strong opinions otherwise, I'll probably do that as part of genn#122 to avoid making this change break the (admittedly somewhat broken) API!

tnowotny:
that makes sense ... so that would mean I would merge this pull request at its current state and we leave the small problem you identified untouched in this branch.

Later, if we merge #122, we need to make sure that brian2genn does not get left behind .... this is potentially a problem as we are just about to release brian2genn but it would be based on the current API ... makes me wonder whether we should wait for the new one ... but then, that would necessitate some (not sure how serious) recoding of brian2genn.

neworderofjamie:
Well we would need to implement a fix for the spike-like event detection code either way so I think it makes sense to do **that** here and then incorporate that with the merged support code in #122. Regarding brian2genn - As I think we agreed that we'd still provide the old API it **shouldn't** cause any issues but I can give it a test if you'd like.

tnowotny:
ok - we will test it when you move towards something more final on #122 ...

tnowotny:
One more comment I would have ... at some point we also need to look at 
a) including brian2genn in automatic testing
b) checking the status of the SpineML-> GeNN pipeline (with Alex Cope). 
The latter is supposed to be our main toolchain for BoB going forward ...

neworderofjamie:
@tnowotny  - this essentially already got merged with #126 as I was using the google tests to test the support code. Hope that's ok - I wasn't expecting you to merge #126 quite so immediately :)

tnowotny:
oh well .. too excited to get this on the road I guess. No harm done.

neworderofjamie:
Haha support code is indeed an exciting business

tnowotny:
I was more thinking of the testing ;-) ... even though many people don't think that's exciting either.

PR #124, opened 2017-02-16
Fix spike like event support code

neworderofjamie:
Turns out some combination of the two fixes discussed in #123 seems the best idea. However I have encountered a couple of problems/questions:

1. How the 'retest' mechanism implemented in modelSpec.cc worked. I think the **intention** was to set the retest flag if there was more than one non-identical condition but it was a bit unclear. I've reimplemented neuronSpkEvntCondition as a set of code strings and namespace names which should mean identical tests don't get generated at all and the retest flag is only set if this set contains more than 1 element.
2. More seriously, how was the whole support code mechanism supposed to work in CUDA? If I don't decorate the functions I define in support code with ``__global__`` or whatever it complains about calling from GPU to CPU code mid-kernel which makes sense. However if I add the decoration it complains about kernels not being registered (and presumably also won't compile in CPU mode)

### DIFF
--- a/lib/include/modelSpec.h
+++ b/lib/include/modelSpec.h
@@ -29,6 +29,7 @@ Part of the code generation and generated code sections.
 #include "synapseModels.h"
 #include "postSynapseModels.h"
 
+#include <set>
 #include <string>
 #include <vector>
 
@@ -119,7 +120,7 @@ public:
   vector<bool> neuronNeedTrueSpk; //!< Whether spike-like events from a group are required
   vector<bool> neuronNeedSpkEvnt; //!< Whether spike-like events from a group are required
   vector<vector<bool> > neuronVarNeedQueue; //!< Whether a neuron variable needs queueing for syn code
-  vector<string> neuronSpkEvntCondition; //!< Will contain the spike event condition code when spike events are used
+  vector<set<pair<string, string>>> neuronSpkEvntCondition; //!< Will contain the spike event condition code when spike events are used
   vector<unsigned int> neuronDelaySlots; //!< The number of slots needed in the synapse delay queues of a neuron group
   vector<int> neuronHostID; //!< The ID of the cluster node which the neuron groups are computed on
   vector<int> neuronDeviceID; //!< The ID of the CUDA device which the neuron groups are comnputed on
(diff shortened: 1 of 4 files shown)


kernfel:
On 2 - decorate with `__device__ __host__` rather than `__global__` (which is only for standalone kernels).

neworderofjamie:
Awesome - thank you @kernfel !! Also through some magic I have yet to locate it runs fine on CPU

jamesturner246:
Just in case you were still wondering, the magic comes from the `__host__` tag, which compiles it for host. Having both `__host__` and `__device__` tags makes it available in both host and device code.

neworderofjamie:
But does that get stripped somewhere if you're compiling with CPU_ONLY=1 as I guess standard GCC knows nothing of this new-fangled business?

jamesturner246:
Good point. Currently not, it seems..

neworderofjamie:
Ok, I feel this is a separate issue than this pull-request set out to address. @tnowotny if you're happy with this fix, merge it into YOUR pull request (#123) and I can have a think about what do about making support code work better across GPU and CPU. There are also problems with maths function and constant prefixes so perhaps ``ensureMathFunctionFtype`` and ``ensureFtype``should be run over it.

tnowotny:
Ah ... yes... CPU_ONLY mode .... that needs to not have these modifiers ... need to fix this later. I am pulling this into the fixVariableScopes branch now as @neworderofjamie suggested.

tnowotny:
argh ... hold on ... I should have looked a bit more carefully. I am not sure what the retest was about was clear. We may have to have a quick chat offline.

kernfel:
You can `#include <cuda_runtime.h>` in CPU-only code, too; I don't think it adds any code, but it defines the decorations (as empty macros, iirc).

tnowotny:
However, we try to avoid this in CPU_ONLY mode as it is meant to also be available to people that o not have CUDA installed and/or have no NVIDIA hardware.

neworderofjamie:
I think we just need to remove __device__ etc from the support code in CPU
only mode

kernfel:
Or `#define` them as empty, probably less work. :)

neworderofjamie:
Excellent idea

PR #125, opened 2017-02-17
Fixup support code

neworderofjamie:
This should solve the issues with CUDA constructs in support code
- In CPU_ONLY mode all the CUDA qualifiers are #defined to nothing (as @kernfel suggested)
- Support code is run through ensureFType


neworderofjamie:
@tnowotny - could give this a quick check over so we can merge fixVariableScopes? I worry I'm creating something of a nest of branches here :-/

PR #126, opened 2017-02-20
Google tests

neworderofjamie:
So I think these are ready to replace the old feature tests - I have reimplemented all the old tests and added some new ones related to #123. 

The only downside is that these tests do add the requirement of having [google test](https://github.com/google/googletest/archive/release-1.8.0.tar.gz) installed. Once this is done you just need to:
### Linux/Mac
- ``cd tests``
- ``./run_tests.sh`` if you have a GPU  or ``./run_tests.sh -c`` if you do not.
- Code generator coverage can also be calculated and an html report generated (using lcov) using ``./calc_coverage.sh -r`` if you have a GPU or ``./calc_coverage.sh -c -r`` if you do not.

### Windows
- ``cd tests``
- ``run_tests.bat`` if you have a GPU or ``run_tests.bat -c`` if you do not.

Due to machines I have to play with I have tested them on:
- Ubuntu Linux 32-bit ARM on CPU and GPU
- Ubuntu Linux 64-bit X86 on CPU and GPU
- Windows 10 with Visual Studio 2013 on CPU

It would be great if someone could test on GPU with Windows and on Mac just to check I haven't done anything dumb. Also if anyone is some sort of batch file ninja, run_tests.bat is less functional than the shell script version.


tnowotny:
A couple of things: I tested on Mac and first thing, for unknown reason &>> does not work in Mac bash ... mysterious but I found some traces of similar reports on the web.
Second, once that was fixed (replaced with
```
make clean 1>> ../../msg 2>> ../../msg
```
etc), it still fails with the following problem
```
clang++ -std=c++11  -c -o test.o test.cc -I  -isystem /include -I"/home/t/tn/tn41/localdisk_projects/develop/genn//lib/include" -I"/home/t/tn/tn41/localdisk_projects/develop/genn//userproject/include" -I"/usr/local/cuda/include"
clang: error: no such file or directory: '/include'
make: *** [test.o] Error 1
```
I am not sure where the ```-I  -isystem /include``` comes from but I can see that it would be causing trouble. Any thoughts?

neworderofjamie:
So when you install Google tests you need to do something like
```bash
# Add Google Test to path
GTEST_DIR=/home/j/jk/jk421/googletest-release-1.8.0/googletest
export GTEST_DIR
```

tnowotny:
hum ... installed gtest through macports. Do you need the path to the include dir or to the lib? Or both?
i.e. these would be 
```
/opt/local/include/gtest/
/opt/local/lib/
```

neworderofjamie:
I didn't realise you could package manager install it and **seemingly** https://trac.macports.org/ticket/48800 google don't want that to happen (probably because it's effectively a single c++ file produced by one of those concatenation tools)

tnowotny:
Ok ... merging now ... I was able to verify it's working on Mac. We need to look at windows at some stage but Linux + Mac is a good indication that in the core it's ok.

neworderofjamie:
As I say it works on Windows in CPU_ONLY mode (not sure how wide a range of Visual C++s we aim to support)

tnowotny:
Last thing ... we need to add instructions about the new dependency to READMEs and manual eventually ... it is ever so slightly annoying that it needs some manual install steps for google tests.

PR #127, opened 2017-02-22
Development into master

tnowotny:
Merging latest fixes into master with a view of making a micro-release with these soon. That release could be underlying the to-be-released brian2genn.


neworderofjamie:
I have had a scan through this and I can't see anything unintended that has crept it!

PR #128, opened 2017-02-27
Fix syntax for nested template use

mstimberg:
The commit a1d23010 changed line 123 of `modelSpec.h` from
```C++
vector<string> neuronSpkEvntCondition;
```
to
```C++
vector<set<pair<string, string>>> neuronSpkEvntCondition;
```
My compiler (g++ 5.4) does not seem to like this:
```
error: ‘>>’ should be ‘> >’ within a nested template argument list
```
Adding spaces to the closing angular brackets does indeed solve the problem for me.

### DIFF
--- a/lib/include/modelSpec.h
+++ b/lib/include/modelSpec.h
@@ -120,7 +120,7 @@ public:
   vector<bool> neuronNeedTrueSpk; //!< Whether spike-like events from a group are required
   vector<bool> neuronNeedSpkEvnt; //!< Whether spike-like events from a group are required
   vector<vector<bool> > neuronVarNeedQueue; //!< Whether a neuron variable needs queueing for syn code
-  vector<set<pair<string, string>>> neuronSpkEvntCondition; //!< Will contain the spike event condition code when spike events are used
+  vector<set<pair<string, string> > > neuronSpkEvntCondition; //!< Will contain the spike event condition code when spike events are used
   vector<unsigned int> neuronDelaySlots; //!< The number of slots needed in the synapse delay queues of a neuron group
   vector<int> neuronHostID; //!< The ID of the cluster node which the neuron groups are computed on
   vector<int> neuronDeviceID; //!< The ID of the CUDA device which the neuron groups are comnputed on


kernfel:
That's a standards issue. From c++11 onwards, the space between closing > is no longer necessary. Given how much fun @neworderofjamie is having with templates, I'd expect c++11 to become a strong requirement by the time #122 is merged.

neworderofjamie:
Yeah, C++ 11 finally fixed that awful hole in the C++ grammar. @mstimberg - Are you definitely compiling with ``-std=c++11``?

mstimberg:
Ah, thanks , my mistake. I was actually just calling `make` for one of the example projects in `userproject` -- none of them seems to set `-std=c++1`.
As a nitpicking note: the line right above the line that I changed *does* use a space, I guess it would be good to have a consistent style.

neworderofjamie:
Breaking the examples is definitely my bad! Which one are you trying? I had noticed that some of the pre-code generator executables for the examples weren't built using C++ 11 (hence my moving of a lot of stringUtils.h into codeGenUtils.h) but I didn't realise any depended on modelspec

mstimberg:
I tried `MBody1_project`:
```
$ make
g++ -Wall -Winline -O3 -o generate_run generate_run.cc -I"/home/marcel/programming/genn/lib/include" -I"/home/marcel/programming/genn/userproject/include"
In file included from /home/marcel/programming/genn/lib/include/stringUtils.h:5:0,
                 from generate_run.cc:40:
/home/marcel/programming/genn/lib/include/modelSpec.h:123:33: error: ‘>>’ should be ‘> >’ within a nested template argument list
   vector<set<pair<string, string>>> neuronSpkEvntCondition; //!< Will contain the spike event condition code when spike events are used
                                 ^
GNUmakefile:20: recipe for target 'generate_run' failed
make: *** [generate_run] Error 1
```

neworderofjamie:
Ok - I see what's going on - definitely my fault. I'll make a P.R. that fixes this properly.

mstimberg:
This is no longer an issue thanks to the fix merged in #129 .

PR #129, opened 2017-02-27
Forward declare NNModel in stringUtils.h rather than including modelSpec.h

neworderofjamie:
@mstimberg - this should fix the issue encountered in #128! If you could confirm that it does, I'll merge.

### DIFF
--- a/lib/include/stringUtils.h
+++ b/lib/include/stringUtils.h
@@ -2,11 +2,13 @@
 #ifndef STRINGUTILS_H
 #define STRINGUTILS_H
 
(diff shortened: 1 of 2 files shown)


mstimberg:
Yes, works for me, with this change I can build/run `MBody1_project` without problems.

neworderofjamie:
Ok great - if you're happy I will delete the other P.R.

neworderofjamie:
I totally meant to make this a pull request off master - apologies - just merged it into there as well

PR #130, opened 2017-03-01
Zero copy

neworderofjamie:
Providing your GPU supports the extension, this change allows you to flag state variables to be implemented as a single zero-copy array rather than separate host and device arrays:

```c++
// Store spikes from neuron population 'E' in zero-copied memory
model.setNeuronSpikeZeroCopy("E");

// Store membrane voltage of neurons in population 'E' in zero-copied memory
model.setNeuronVarZeroCopy("E", "V");

// Store synaptic weights for synapse population 'IE' in zero-copied memory
model.setSynapseWeightUpdateVarZeroCopy("IE", "g");
```

Under the hood this changes the way the state variables are allocated and freed and removes the calls to ``cudaMemcpy`` from the ``pullXXXXFromDevice`` and ``pushXXXToDevice`` helpers. Any thoughts on the API/implementation would be much appreciated.

Specifically I am still a little unsure about exactly how the ``__device__`` symbols (the ones with the dd_ prefix) work so I'm a little unsure about how sane the implementation of ``deviceZeroCopy`` here https://github.com/genn-team/genn/blob/zero_copy/lib/src/generateRunner.cc#L450-L461 is.

### DIFF
--- a/lib/include/generateRunner.h
+++ b/lib/include/generateRunner.h
@@ -29,20 +29,6 @@ using namespace std;
 
 
 //--------------------------------------------------------------------------
-//! \brief This fucntion generates host and device variable definitions, of the given type and name.
-//--------------------------------------------------------------------------
-
-void variable_def(ofstream &os, const string &type, const string &name);
-
-
-//--------------------------------------------------------------------------
-//! \brief This fucntion generates host extern variable definitions, of the given type and name.
-//--------------------------------------------------------------------------
-
-void extern_variable_def(ofstream &os, const string &type, const string &name);
-
-
-//--------------------------------------------------------------------------
 /*!
   \brief A function that generates predominantly host-side code.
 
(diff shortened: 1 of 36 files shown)


jamesturner246:
The `d_` variables are pointers to device memory which are accessible only in host code. The `dd_` variables are also pointers to device memory, but are only accessible in device code. It's a workaround for devices which don't support unified virtual addressing (UVA) - i.e. devices lower than CC 3.0.

About the zero-copy API: I'm wondering what the `"I"` and `"IE"` arguments represent?

neworderofjamie:
Ahh OK - makes sense. They were just the names for my example neuron and
synapse populations.

tnowotny:
Why do the function names contain "initVal"? Isn't it setting a property of the variable rather than its initial value? In other words should they be called
```
model.setNeuronVarZeroCopy("E", "V");
```
etc?

neworderofjamie:
Excellent point - I wasn't really sure what 'official' GeNN terminology was. I'll rename them (and probably also in other places in the new model API)

neworderofjamie:
@tnowotny - renaming is done - what do you think? I'm not 100% sure it's useful in all situations but definitely speeds up recording simulation state on Jetson

tnowotny:
The only slight doubt I had was on ```VarValues```versus ```InitValues``` ... in a sense these are initial values for the variables, so both names make sense.

neworderofjamie:
Yeah - My thinking was:

1. less typing and 
2. the class doesn't, by definition, hold initial values - it just holds a set of model variable values.

PR #131, opened 2017-03-02
Use the correct synapse name for "using namespace" (synapse dynamics)

mstimberg:
@moritzaugustin reported an error with brian2genn (brian-team/brian2genn#46) which I think is due to an incorrect use of `model.synapseName[i]`, which should be `model.synapseName[k]` instead, or equivalently `synapseName` (which has earlier been introduced already). I'm not 100% sure about this, but it fixes our issue :) I think this got not detected by the tests because it needs several synapse models to occur, but again, I don't claim to really understand this code ;)


tnowotny:
Well spotted. Amazing how such a typo could survive - but then we don't have test coverage for support code. Merged.

PR #132, opened 2017-03-02
Do not use `std::to_string` to convert double values to strings

mstimberg:
Triggered by some "division by zero" warnings, I realized that GeNN's use of `std::to_string` to convert the parameter values into strings to directly insert them into the code is rather dangerous. This function uses the standard `%f` formatting, i.e. only 6 digits after the decimal point. I think in most GeNN models this is not a big problem, since parameters are defined in ms, mV, etc. However, in `brian2genn` we use Brian's unit system which internally represents all numbers in the base unit, e.g. `-70mV` is represented as `-0.070`. We therefore not only lose precision, but GeNN's string formatting will replace every value that is smaller than `1e-7` by `0`, so that means that all those conductances in `nS`, or capacitances in `pF` will completely disappear...

The change proposed with this PR takes a very conservative approach and uses a string representation that is [supposed to give the bit-by-bit equivalent](http://www.boost.org/doc/libs/1_47_0/libs/math/doc/sf_and_dist/html/math_toolkit/utils/fp_facets/examples.html#math_toolkit.utils.fp_facets.examples.simple_example_with_std__stringstreams)  -- my C++ string skills are very rudimentary, though, maybe there's a better solution. This is a bit ugly since the values in the generated code will be quite long (e.g. `-70` will be represented as `-7.00000000000000000e+0`...), but well. I'd be happy with anything that does not replace our constants by zero :wink:

### DIFF
--- a/lib/include/codeGenUtils.h
+++ b/lib/include/codeGenUtils.h
@@ -1,6 +1,7 @@
 #pragma once
 
 // Standard includes
+#include <limits>
 #include <string>
 #include <sstream>
 #include <vector>
@@ -85,7 +86,10 @@ inline void value_substitutions(string &code, NameIter namesBegin, NameIter name
     NameIter n = namesBegin;
     auto v = values.cbegin();
     for (;n != namesEnd && v != values.cend(); n++, v++) {
-        substitute(code, "$(" + *n + ")", "(" + to_string(*v) + ")");
+        stringstream stream;
+        stream.precision(std::numeric_limits<double>::max_digits10);
+        stream << std::scientific << *v;
+        substitute(code, "$(" + *n + ")", "(" + stream.str() + ")");
     }
 }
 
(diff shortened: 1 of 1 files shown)


tnowotny:
Funny this ... you have reinstated something quite similar to what I used to do earlier in my own ```toString()``` function. @neworderofjamie I suspsect you made this change ... would you want to review Marcel's code? I suspect it is not strictly necessary to do the 
```stream.precision(std::numeric_limits<double>::max_digits10);```
bit to avoid the rounding to 0, but potentially if one wanted the full precision one does ...

neworderofjamie:
I am definitely to blame for this - apologies! My purging of ``toString`` was clearly too aggressive. I think we can probably tweak this a little to take the floating point type of the model into account. By the way @mstimberg, are you using development for your Brian2GeNN work?

mstimberg:
> By the way @mstimberg, are you using development for your Brian2GeNN work?

I use whatever works ;) I'd prefer something more stable like master, but I'd like to benefit from fixes like this without having to wait for a long time. For now I am fine using the development branch, but ideally when we do the brian2genn release (which is not *too* far in the future), it would be great to have a GeNN release that we state as dependency.

neworderofjamie:
That is fair enough - I only ask because I recently merged #122 into development which adds a (hopefully) more user-friendly way of defining models, but was a fairly far-reaching change.

tnowotny:
To be honest I am thinking more and more we can just merge this. Yes, having these very long constants in the code isn't pretty, especially when the extra digits are not adding any information but on the other hand it is also not the end of the world and at least we get a guarantee not to incur errors from this ...

tnowotny:
I would then probably also cherry-pick it across to master

neworderofjamie:
Agreed! I'll cherry pick it across right away.

neworderofjamie:
(and add some unit tests of the ``value_substitution`` mechanism)

PR #133, opened 2017-03-03
CodeGenUtil unit tests

neworderofjamie:
As much to get the finer-grained test creation ball rolling that anything I added some tests to cover the bug @mstimberg  found. In the process (and to save writing duplicate tests) I removed the ``extended_name_substitutions`` and ``extended_value_substitutions`` and added the extended options they provide as additional arguments (with defaults) to the standard versions. Once this is merged, I'll cherry pick along with #132 to master.

I'm pretty bad at writing Makefiles so the one that builds the unit tests could almost certainly do with some improvements!


tnowotny:
seems reasonable. Removing the ```extended``` versions a good idea as long as there will never be a confusion as to the meaning of each of the long list of parameters which are in the majority of type ```string``` ...

Issue #92, opened 2016-05-31
Ambiguity in connectivity/conductance schemes


esinyavuz:
I was about to add some stuff in the documentation explaining the different combinations of `GLOBALG` / `INDIVIDUALG` / `INDIVIDUALID` conductance schemes with `ALLTOALL` / `DENSE` / `SPARSE` connectivity schemes, but I am confused myself with the way we have implemented some of these so I write it down here as it can be useful for other users and we can consider simplifying it in future releases.

It looks like we perform some variable substitutions in the `simCode` only if `INDIVIDUALG` is used. We substitute the values of the initial values of the variables without the 'dd_' prefix and the neuron index otherwise. This is because of some historical reasons from the times when we only had the conductance as a variable. On the other hand, it may also make more sense to make the conductance a parameter instead of a variable when `GLOBALG` is used. And it is actually possible to have synapses which have both fixed and variable members, by defining those as parameters or variables.

I will try to explain the current situation and how I thing it can be done:

|  | ALLTOALL | DENSE | SPARSE |
| --- | :-: | :-: | :-: |
| GLOBALG | Fixed values for all synapse members | Fixed values for all synapse members | Fixed values for some synapse members (using sparse indexing) |
| INDIVIDUALG | Variable values for all synapse members | Variable values for all synapse members | Variable values for some synapse members |
| INDIVIDUALID | Fixed variables for some synapse members (using bit shifting) | Fixed variables for some synapse members (using bit shifting) | N/A |

`INDIVIDUALID` was considered to be useful for `DENSE` connectivities, but in practice there is no difference between `ALLTOALL` and `DENSE`. At the end in reality, all we need is to decide whether we have to keep values for all synapse members or for some only, an decide how we do it. Fixed values can be expressed as parameters, index and time dependent ones can be variables. In case of sparse connectivity, we can include a `INDIVIDUALID`-like behaviour for fixed values. Actually this makes code generation much easier, and we can write wrapper functions for existing users.

Also there are still some explicit references to the obsolete hard-coded variable `g` (e.g. [here](https://github.com/genn-team/genn/blob/development/lib/src/generateKernels.cc#L666)). We will get rid of these small mistakes if we simplify the connectivity schemes.

Feel free to edit the table above if I have mistaken anything.

jamesturner246:
I think this is done. Closing.

PR #137, opened 2017-03-15
Synapse type simplification

neworderofjamie:
As discussed previously, the previous systems of combining a ``SynapseConnType`` and a``SynapseGType`` was somewhat counter-intuitive, some combinations made no sense and ``INDIVIDUALID`` was, oddly, a ``SynapseGType``.

I believe the following combinations actually make sense (and this PR fixes the code generator so they all work):

- ``SynapseConnType=ALLTOALL/DENSE`` and ``SynapseGType=INDIVIDUALG`` - now called ``DENSE_INDIVIDUALG`` 
- ``SynapseConnType=SPARSE`` and ``SynapseGType=INDIVIDUALG`` - now called ``SPARSE_INDIVIDUALG``
- ``SynapseConnType=SPARSE`` and ``SynapseGType=GLOBALG`` - now called ``SPARSE_GLOBALG``
- ``SynapseGType=INDIVIDUALID`` - now called ``BITMASK_GLOBALG``

The user now just passes ones of these to ``addSynapsePopulation`` (and because they're defined using a scoped C++11 enum, they can't pass anything else).

Internally these enums are defined in terms of a number of flags which the code generator checks. The plan is that by, for example adding a ``FP16`` flag, I'll be able to add special cases for this without having to update all bits of the generator to be aware of it.

Also I've added more feature tests which test the various valid combinations


tnowotny:
How about ```DENSE/ALLTOALL``` and ```GLOBALG```? Not the most exciting case but could happen?

neworderofjamie:
I wasn't sure if that was a useful case, but if it is, I'm happy to restore it!

By the way, as far as I could tell, the only test the code ever made on ``SynapseConnType`` was whether it was ``SPARSE`` or not so there was no distinction between the behaviour of ``DENSE`` and ``ALLTOALL``.

tnowotny:
Yes, you are right ... ```ALLTOALL``` was never actually necessary. In a sense ```ALLTOALL=DENSE```

esinyavuz:
Also see #92 about different connectivity schemes that used to be there before. It all makes more sense with the latest version of combinations I think.

neworderofjamie:
Thanks for looking at this @esinyavuz - wish I'd seen that earlier - that is exactly the diagram me and Thomas ended up drawing :)

neworderofjamie:
@tnowotny - did you think of a use-case for ``DENSE`` and ``GLOBALG`` or is this ok to merge?

tnowotny:
I think we should not just exclude it even if not that common. I thought it was previously a possibility, or was it not?

neworderofjamie:
Ok - will re-enable that combination - I only left it out because I
couldn't really see a use for it!

neworderofjamie:
Ok it's re-enabled and I have added a test

tnowotny:
ok ... great. I will merge it then.

PR #140, opened 2017-03-21
Model spec refactor

neworderofjamie:
This is the refactor I was talking about on Saturday - it still needs a few bits tidying up, but what do you think?

Basically

- Synapse and neuron groups are now encompassed in ``SynapseGroup`` and ``NeuronGroup`` rather than just being indices into lots of vectors in ``NNmodel``.
- The synapse and neuron groups are now stored in a ``std::map`` in ``NNmodel`` meaning that they are always accessed by name rather than at times being accessed by name and at other by index 
- Tried to move a lot of common code for performing substitutions etc out of the CPU and CUDA generator into ``StandardSubstitutions`` namespace. **Hopefully** this will make maintaining CUDA, CPU and OpenCL backends somewhat easier
- Started trying to encapsulate some of the higher level bits of code generation functionality in ``StandardGeneratedSections`` namespace. Again, **hopefully** this will aid the maintenance of multiple backends.
- Also made some of my older changes consistent with the GeNN naming convention so method names are in lowerCamelCase.

In general these changes should be fully backward compatible apart from where any old code is doing anything really dodgy and directly accessing the vectors within ``NNmodel`` which I think is fair.


jamesturner246:
I remember Thomas not being very keen on replacing the vector index mechanism last time I asked, but I still very strongly agree with this change. Everything is much easier to follow, and we shouldn't have any `vector[i][j]` mixups again.

tnowotny:
This seems reasonable. I wasn't keen earlier as at times refactoring takes over peoples lives and nothing of the more pressing problems are dealt with...
I had a quick look and have two comments:
1 - the "setXXXRequired" functions in neuronGroup etc seem unnecessarily restrictive in just setting these true. Even if we don't have a use case right now where things are switched off again, wouldn't it make sense to have them 
```void setXXXRequired(bool req) { m_XXXRequired= req; }```
2 - I think @neworderofjamie you are right and NeuronGroups may well be allowed to know their own name ... why not?

Unfortunately, I don't think I can thoroughly review the entirety of the changes; maybe @jamesturner246, if you agree with it, can you merge?

neworderofjamie:
Indeed - I'm hopeful that this will be the last refactor of this magnitude for a bit!

Yeah  - good points @tnowotny - it's not entirely finished yet though so hold the merging horses!

neworderofjamie:
I am now pretty happy with this - @tnowotny 's feedback is taken into account and if anyone would like to review that'd be awesome. Otherwise, once #141 is merged, I'll use this as a first candidate for some semblance of continuous integration...

jamesturner246:
Could you summarise the major changes that were made? Don't have to go into too much detail.
Cheers!

tnowotny:
Which reminds me, we will have to rework the documentation to reflect all these changes when we get there. Do you all know how the documentation works?

neworderofjamie:
@jamesturner246 I have updated the description at the top of the pull request to hopefully give a bit more detail. Let me know if you have any questions!

neworderofjamie:
@tnowotny - that thought has been at the back of mind too. Once this is merged I probably should make another pull request where I check there's doxygen tags everywhere and then update the documentation to refer to the new API.

tnowotny:
The one element of this that I have not figured out fully is how to expose different versions of the documentation to users. Of course, it's all versioned in the documentation branch but a typical user would not pull an older version from there and build it ... They would look at the online documentation on the .io website which at any given time can be only one particular version.
Or is this a non-problem?

neworderofjamie:
I think the .io website should probably have the documentation for master.

neworderofjamie:
@tnowotny I think this is probably now ready for merging into development. I get the impression no one is going to have time for a full review, but all the tests pass, I've updated all the examples and done a big update on the documentation so as to be consistent (Jim has been a lucky guinea pig in 'testing' this by going through it). Unless you have strong objections, I'd like the resultant development branch to be what we get people using at CapoCaccia with the aim to make a 3.0.0 release shortly afterwards.

tnowotny:
I suspect this is the right thing to do now. We should also have Jim use this version I think.

PR #141, opened 2017-03-21
Jenkins test

neworderofjamie:
Don't actually merge -testing Jenkins automated testing stuff!


neworderofjamie:
This is somewhat limited because:
- I am waiting on SSH access to configure remote nodes so Jenkins can dole out tests to idle workstations with GPUs so currently this only tests `CPU_ONLY`` mode
- The server doesn't have lcov installed so the code coverage isn't getting updating (need SSH access to install this)

However I think it is probably worth merging into development so at least SOME automated testing will happen on new Pull Requests.

I would be happy to talk you through the details of how it works in person if you'd be interested.

neworderofjamie:
[Build Status](https://gen-ci.inf.sussex.ac.uk/job/GeNN/job/genn/job/jenkins_test/)

PR #142, opened 2017-03-22
Fix for nasty indexing-related bug when padded sizes were being calculated

neworderofjamie:
I was trying to figure out why #140 was calculating a different grid size to development and found this nasty bug. Because ``synDynGroups <= synapseGrpN`` and ``lrnGroups <= synapseGrpN`` was never crashing but wasn't looking at correct synapse groups. Thought it would make sense to separately fix it in master.


tnowotny:
Oh dear ... it's kind of glaringly obvious that this was wrong ... I wonder how many I/we have introduced in there.

jamesturner246:
I take the blame for that one. That was when I was refactoring all of the neuron / synapse / learn padSum calculations into one function. Well spotted Jamie.

neworderofjamie:
Optimistically, this is a whole class of bug that less indexing will avoid!

Issue #147, opened 2017-04-18
Bugs about extraGlobal*KernelParameters


whenov:
I was using `extraGlobalNeuronKernelParameters` and `extraGlobalSynpaseKernelParameters` and met three problems.

1. 
It seems that the device variable generated by `extraGlobalNeuronKernelParameters`/`extraGlobalSynpaseKernelParameters` are not used.

Assume that I have two neuron populations Pop1 and Pop2, and have `param` in their `extraGlobalNeuronKernelParameters`, then the generated `d_paramPop1` and `d_paramPop2` are not used at all, but `paramPop1` and `paramPop2` are used instead, this is inconsistent with the name convention of GeNN.

2.
Besides, if I try to use the `param` variable in Pop2 in synapse model by `param_post`, the compilation won't pass when the model is being built, the variable `dd_paramPop2` is reported as undefined.

3.
Another error is, when I used `param_post` in synapse model, the generated runnerGPU.cc only passed `paramPop1` to `calcSynapses`.

neworderofjamie:
Thank you for raising this issue. I will raise a seperate minor issue for 1, but I have fixed the more serious issues 2 and 3 in pull request #148 if you could give that a test.

Jamie

whenov:
@neworderofjamie It works, thanks for your response!

tnowotny:
As for the d_xxx variables, they should probably not be created at all as the extraGlobal* parameters are passed as kernel arguments directly. This is a leftover from when this was refactored by me.
The point about the naming convention is interesting. In principle we could name them dd_xxx but I think it would be confusing as they are not pointers to device memory as the other dd_* variables.

whenov:
@tnowotny 

This is my understanding: The variable `var` is a host variable pointing to host memory, `d_var` is a host variable pointing to device memory, `dd_var` is a device variable pointing to device memory. So in this case, the created variable should be named after `d_var` instead of `var`.

tnowotny:
Hi @whenov, almost, except that the variable when used in host code, is simply a host variable in host memory (hence ```d_xxx``` would not be appropriate). When used in a kernel, it is a kernel parameter passed during the kernel call and resides temporarily in device memory (likely a register, but control is with the nvcc compiler). We could rename it during the call to have a ```dd_xxx``` name within the kernel but as I said above I believe that this is even more confusing.

PR #148, opened 2017-04-18
Backport of fixes for serious parts of #147

neworderofjamie:
- device prefix shouldn't be used on extra global parameters as they're passed directly to kernel rather than accessed as device globals
- building of list of list of synapse kernel parameters was broken


neworderofjamie:
The raiser of the issue is happy that this has fixed his problems

tnowotny:
Ok, that looks like the right fix - thanks @neworderofjamie

Issue #86, opened 2016-05-02
Automatic code formatting


jamesturner246:
We may have a solution to our messy code output; The `astyle` command. It's already present on my Fedora 23 install, but not sure about Ubuntu / Debian / Mac. Can you test this on Mac, Thomas, and on Ubuntu, Esin?

It takes a C or C++ input and makes it nice and pretty. There's also a MS Visual Studio extension, for Windows users: 

https://visualstudiogallery.msdn.microsoft.com/2f3f04cd-2866-4e47-a671-d1cc9cc3fb02

We could have a tiny post-processing stage where this is run on all the output files.

esinyavuz:
I could install it using apt on ubuntu. But how does it work? Will people need to install it to run GeNN? If it will not impose any dependencies I'm happy to try.

jamesturner246:
It would be an optional thing then, I guess. There's also `clang-format` for clang users - i.e. Mac users. It can be run if either exist.

neworderofjamie:
As I was attempting to read some auto-generated code I was thinking - we have most of what we need for improving this (I am especially thinking of the indentation) already in``CodeHelper``. Rather than writing directly to the ``ostream`` we could change ``CodeHelper`` to own the ``ostream`` of the file being generated and add a ``writeLine`` method to it which strips off any existing leading whitespace and adds back in the correct amount.

tnowotny:
I would be happy with such a solution. At times we probably need to tidy up some of the generator code to no write lines incrementally but collect the content in a string incrementally and then write it using this ...?

neworderofjamie:
I think one of the good things about ``std::ofstream`` is that it does buffering for you so there shouldn't be any particular problem

tnowotny:
I am not sure we are talking about the same thing: I meant that you cannot pass half a code line into your new ```<<```, have it formatted and then pass another bit of the same line in the same way, as it would be formatted as if it was a new line (i.e. extra space) ... or am I missing something? I guess you could detect ```endl``` in the overloaded ```<<``` to safeguard against it and leave the original code generation code alone ... is that what you had in mind?

neworderofjamie:
I was thinking you would probably detect ``endl`` in the code as that would also mean this approach would be able to format user code strings.

tnowotny:
that would work ... shall we call it a "codestream" ;-)

neworderofjamie:
I think you mean CodeStream :tm: :smile:

neworderofjamie:
[...]

PR #150, opened 2017-06-08
Codestreams

neworderofjamie:
My annoyance at the auto-generated code formatting finally got the better of me so I addressed #86! As with many things C++ there were 3 possible ways of implementing this:

1. Wrapping a ``std::ostream`` in my own class (my original plan)
2. Deriving off a ``std::ostream``
3. Descending beneath the waves and deriving off a ``std::streambuf``

1 and 2 both turn out to terrible ideas as it becomes a nightmare to get ``std::endl``, ``std::setw``, ``std::setprecision`` etc to work, but 3 ended up being surprisingly simple if low-level. 

Most of this chance is regex - I didn't see a lot of point keeping the old macros just for the sake of it.


tnowotny:
ready to merge imho

PR #151, opened 2017-07-05
Changed the makefiles so that INCLUDE_FLAGS, CXX_FLAGS and LINK_FLAGS…

tnowotny:
… are only added to, not overwritten. This allows more fine tuning in contexts such as brian2genn

### DIFF
--- a/lib/GNUmakefile
+++ b/lib/GNUmakefile
@@ -55,9 +55,9 @@ ifeq ($(DARWIN),DARWIN)
     CXX                  :=clang++
 endif
 ifndef CPU_ONLY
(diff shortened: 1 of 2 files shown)


neworderofjamie:
I only had the misfortune of learning about NMake the other day, but is ```INCLUDE_FLAGS            =$(INCLUDE_FLAGS) XXXXXXXX``` ok? I THOUGHT variables were immutable (but I'd be happy to be wrong)

tnowotny:
I think it is ok. We have used it before I believe.
Argh ... checks failing - not good, that.

neworderofjamie:
I don't know why it's saying that - it seems to be building quite happily

PR #152, opened 2017-07-10
Allow more efficient implementation of trace-based STDP rules

neworderofjamie:
In the process of investigation spike-based learning in SpineML and finding it lacking, I realised that GeNN is missing some useful functionality for spike-based learning too.

If you want to implement trace-based STDP rules you often want to be able to have state that's only stored and updated once per pre or postsynaptic neuron - typically low-pass-filtered spike trains. Apart from by hacking this into neuron models, GeNN had no method to support this - until now....

Basically this pull request adds pre and postsynaptic state variables to weight update models (in a backward-compatible manner). These are substituted into existing bits of weight update user code in the normal way, but additionally there are two new piece of weight update user code which get inserted into the appropriate neuron kernel when populations spike. The last aspect is the one thing I'm a little concerned about as it seems a little ugly - any thought @tnowotny?

For an example - the following weight update model implements Pfister and Gerstner's triplet rule:

```c++
// PfisterTriplet
class PfisterTriplet : public WeightUpdateModels::Base
{
public:
    DECLARE_WEIGHT_UPDATE_MODEL(PfisterTriplet, 10, 1, 2, 2);

    SET_PARAM_NAMES({
      "tauPlus",  // 0 - Fast presynaptic time constant (ms)
      "tauMinus", // 1 - Fast postsynaptic time constant (ms)
      "tauX",     // 2 - Slow presynaptic time constant (ms)
      "tauY",     // 3 - Slow postsynaptic time constant (ms)
      "A2Plus",    // 4 - Pair-based potentiation strength (dimensionless)
      "A2Minus",    // 5 - Pair-based depression strength (dimensionless)
      "A3Plus",    // 6 - Triplet-based potentiation strength (dimensionless)
      "A3Minus",    // 7 - Triplet-based depression strength (dimensionless)
      "Wmin",     // 8 - Minimum weight
      "Wmax",     // 9 - Maximum weight
    });

    SET_VARS({{"g", "scalar"}});
    SET_PRE_VARS({{"r1", "scalar"}, {"r2", "scalar"}});
    SET_POST_VARS({{"o1", "scalar"}, {"o2", "scalar"}});

    SET_PRE_SPIKE_CODE(
        "scalar dt = $(t) - $(sT_pre);\n"
        "$(r1) = ($(r1) * exp(-dt / $(tauPlus))) + 1.0;\n"
        "$(r2) =  ($(sT_pre) == 0.0) ? 0.0 : ($(r2) + 1.0) * exp(-dt / $(tauX));\n");

    SET_POST_SPIKE_CODE(
        "scalar dt = $(t) - $(sT_post);\n"
        "$(o1) = ($(o1) * exp(-dt / $(tauPlus))) + 1.0;\n"
        "$(o2) = ($(sT_post) == 0.0) ? 0.0 : ($(o2) + 1.0) * exp(-dt / $(tauY));\n");

    SET_SIM_CODE(
        "$(addtoinSyn) = $(g);\n"
        "$(updatelinsyn);\n"
        "scalar dt = $(t) - $(sT_post); \n"
        "if (dt > 0)\n"
        "{\n"
        "    scalar o1 = $(o1) * exp(-dt / $(tauMinus));\n"
        "    scalar newWeight = $(g) - o1 * ($(A2Minus) + ($(A3Minus) * $(r2)));\n"
        "    $(g) = (newWeight < $(Wmin)) ? $(Wmin) : newWeight;\n"
        "}\n");

    SET_LEARN_POST_CODE(
        "scalar dt = $(t) - $(sT_pre);\n"
        "if (dt > 0)\n"
        "{\n"
        "    scalar r1 = $(r1) * exp(-dt / $(tauPlus));\n"
        "    scalar newWeight = $(g) + r1 * ($(A2Plus) + ($(A3Plus) * $(o2)));\n"
        "    $(g) = (newWeight > $(Wmax)) ? $(Wmax) : newWeight;\n"
        "}\n");

    SET_NEEDS_PRE_SPIKE_TIME(true);
    SET_NEEDS_POST_SPIKE_TIME(true);
};

IMPLEMENT_MODEL(PfisterTriplet);
```
Continuing the theme of recreating bits of my thesis using GeNN I implemented the Sjostromet al. (2001) pairing frequency experiment [here](https://github.com/neworderofjamie/genn_examples/tree/master/sjostrom_triplet) which shows that the GeNN version behaves much like the Pfister and Gerstner paper suggests:

sjostrom


tnowotny:
I think this is a good idea, even though I also worry what this means for synaptic delays. In essence, if this is used for a traced based STDP and there is axonic delay, presumably the pre-synaptic trace should be in a delay buffer. Similarly, if there is dendritic delay, on could think of delays for spike backpropagation and postulate a delay buffer for the post-synaptic trace.
What do you think?
For a no-delay situation, this should be fine. I think we have already crossed the line of keeping synapses and neurons apart when we allowed direct access to neuron variables from synapse code ... 
But you are right that adding tests, as usual, is a good idea!

neworderofjamie:
Yeah, having thought a bit more about this in the context of the dendritic
delay, I think the correct thing to do is, as you say, delay the
presynaptic trace in the same way as neuron variables using the variable
queues. The whole question of back propagation delay is trickier - perhaps
one for another day.

neworderofjamie:
@tnowotny this is finally ready and, aside from documentation updates etc is the last feature for the 3.2.0 release! Are you ok to merge? Previous concerns about delays are fixed and I've added new tests.

tnowotny:
I believe the design is right. The code for updating the spike traces needs to be in the neuron kernel. The only other solution would be a separate kernel which seems silly unless the code was very heavy computationally and would lead to a lot of divergence of spiking and silent neurons in the neuron kernel.
I agree that it seems odd if a synapse aspect influences the neuron kernel code but it only makes sense in this circumstance. The average user would never know anyway.

tnowotny:
I did not inspect the code in detail as often; but I think we should go ahead and merge - there are some very busy times ahead.

PR #153, opened 2017-07-12
Multiple postsynaptic input

neworderofjamie:
As discussed post synaptic models can only currently update the ``Isyn`` local variable using this syntax:
```c++
SET_CURRENT_CONVERTER_CODE("$(inSyn) * ($(E) - $(V))");
```
This macro still exists but is now shorthand for:
```c++
SET_APPLY_INPUT_CODE("$(Isyn) += $(inSyn) * ($(E) - $(V));");
```
This new, more flexible code string can now be used to implement more complex code such as:
```c++
SET_APPLY_INPUT_CODE(
    "$(Isyn) += $(inSyn) * ($(E) - $(V));\n"
    "Isyn2 *= $(inSyn);\n");
```
This means that different postsynaptic update rules can inject input in different ways. 
So neuron kernel can correctly initialise these 'additional input variables' neuron models can list additional input variable, their types and initial values e.g.
```c++
SET_ADDITIONAL_INPUT_VARS({{"Isyn2", {"scalar", 1.0}}});
```

### DIFF
--- a/lib/include/newModels.h
+++ b/lib/include/newModels.h
@@ -122,6 +122,7 @@ public:
     typedef std::function<double(const std::vector<double> &, double)> DerivedParamFunc;
     typedef std::vector<std::string> StringVec;
     typedef std::vector<std::pair<std::string, std::string>> StringPairVec;
+    typedef std::vector<std::pair<std::string, std::pair<std::string, double>>> NameTypeValVec;
     typedef std::vector<std::pair<std::string, DerivedParamFunc>> DerivedParamVec;
 
     //----------------------------------------------------------------------------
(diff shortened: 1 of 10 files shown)


neworderofjamie:
Definitely not intention but, as far as I can tell, they weren't added by **this** pull-request - I think Capocaccia could be to blame!

tnowotny:
Which pull request? I could believe that I might have added some stuff that should not have been there in the heat of the moment ...

neworderofjamie:
Haha I think these are just in the development branch to be honest - so yes
almost certainly a 'heat of the moment problem'

PR #154, opened 2017-07-17
Synapse dynamics code which provides input

neworderofjamie:
Currently synapse dynamics code could only update state variables in the neuron or weight update rule. It couldn't provide synaptic input through the postsynaptic update. 

This prevents it being used for updating continuous synapses (which I think is a nicer way to implement these than sending a spike-like-event every time step). This change should let ``$(updatelinsyn)``, ``$(inSyn)`` and ``$(addtoinSyn)`` be used within the synapse dynamics in the same way they are within true-spike handling code. 

I'm really not that familiar with the synapse dynamics code though so a review would be much appreciated!

### DIFF
--- a/lib/include/standardSubstitutions.h
+++ b/lib/include/standardSubstitutions.h
@@ -91,9 +91,9 @@ void neuronReset(
 void weightUpdateThresholdCondition(
     std::string &eCode,
     const SynapseGroup &sg,
-    const DerivedParamNameIterCtx &nmDerivedParams,
-    const ExtraGlobalParamNameIterCtx &nmExtraGlobalParams,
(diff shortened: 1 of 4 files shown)


jamesturner246:
@neworderofjamie, do you know anything about this failing Jenkins test?

neworderofjamie:
I am all for pedantry especially when it makes an already confusing list of parameters more confusing -  that's a good spot!

tnowotny:
If @jamesturner246 has no objections please merge ... sorry I was just too busy over here.

neworderofjamie:
I thought that might be the case @tnowotny so I handed some of these off to @jamesturner246 !

tnowotny:
should probably make a gap-junction like test with it ...

PR #155, opened 2017-07-19
Determining which weight update model extra global parameters to pass to kernels was totally broken

neworderofjamie:
1. For some reason I was merging all three types (synapse dynamics, synapse and post learning) of parameter together
2. Even if I had been building maps of post learning parameters correctly, NNmodel::getSimLearnPostKernelParameters() was returning the wrong variable


jamesturner246:
Everything looks corrected as you describe. Looks good to me.

PR #161, opened 2017-10-06
cuRAND integration

neworderofjamie:
This is one I've been intending to do for a while, but turns out that SpineML required it (and it will make GeNN look somewhat better in the paper I'm writing). Anyway in short this lets you sample from standard random distributions from your GeNN user code. For example the sim code for an Izhikevich neuron with an additional, uniformly-distributed noise current might look like this:

```cpp
const scalar noise = ($(gennrand_uniform) * $(n) * 2.0) - $(n);
$(V)+=0.5*(0.04*$(V)*$(V)+5.0*$(V)+140.0-$(U)+$(Isyn)+noise)*DT; //at two times for numerical stability
$(V)+=0.5*(0.04*$(V)*$(V)+5.0*$(V)+140.0-$(U)+$(Isyn)+noise)*DT;
"$(U)+=$(a)*($(b)*$(V)-$(U))*DT;
```

This gets mapped to C++11 <random> library calls in CPU mode (n=6.5):

```cpp
scalar noise = (std::uniform_real_distribution<float>(0.0f, 1.0f)(rng) * (6.50000000000000000e+00f) * 2.0f) - (6.50000000000000000e+00f);
```

Or cuRAND calls on GPU:

```cpp
scalar noise = (curand_uniform(&dd_rngE[id]) * (6.50000000000000000e+00f) * 2.0f) - (6.50000000000000000e+00f);
```

This mechanism is pretty general so can be used to expose any functions we want to provide different CPU and GPU implementations to user code. The GeNN helper kernels @esinyavuz wrote make the same curand calls but are a pain if you want your code to work on CPU as well and, my experiments show, this is significantly faster.

This change also implements a more correct Poisson spike source which samples from the exponential distribution to determine which timestep to next spike in. I'm using Von Neumann's exponential generator which, because of it's loopyness, is a pain to implement in a user code block so I have added CUDA implementations to runnerGPU.

In order to create seed arrays I have also added a new``init`` kernel which gets launched at the end of the ``initialize`` function. I intend to extend this in future so state variables can be initialised on device to either constant values or from random distributions.

Todo
- [ ] Allow random functions to be used in _synapse_ group user code


tnowotny:
I think this is a very good one - but tests still failing?
how did the "E" get into dd_rngE[...] ?

neworderofjamie:
Thanks! I was quite pleased how it turned out too. The E is the population
name. Regarding the tests, something seems to have gone wrong with the
jenkins server - I'll look into it next week

jamesturner246:
I don't suppose we could use <stdint.h> for a while, rather than <cstdint>? Just until gcc starts using c++11 by default?

neworderofjamie:
Yeah - that was what my change did. A C++11 compiler is required to use GeNN though so it seems a bit weird to me that user code is, by default, built as C++98.

jamesturner246:
That is odd. Why not just have >= c++11 default everywhere?

neworderofjamie:
Precisely!

tnowotny:
Might just be historic. Not too long ago it wasn't necessarily true that we could a C++ supporting compiler on all platforms. By now, I suppose that is safe to assume?

jamesturner246:
I think, given that we use so many C++11 features in code generation now, that we have no choice!

neworderofjamie:
@tnowotny  - if you agree, and once this PR is merged, I'll make another PR to compile generated code as C++11

tnowotny:
I think it is about time we do that - agreed.

neworderofjamie:
Yes, I was thinking of making it more general to address that. Maybe let
you set a C++ standard in the user preferences seems a sensible idea. C++11
seems a sensible default though - that is about what a slightly out of date
gcc or visual studio (with some irritating lacking features) will provide.

neworderofjamie:
I have added some tests of the RNG stuff which simulate 1000 neurons for 1000 time steps (arbitrary) and then test the samples against the desired distribution using a Kolmogorov-Smirnov test (thanks Numerical Recipes :)) so I think this is ready to merge - while, especially in light of Tim Vogel's talk yesterday, per-synapse RNGs would be cool I can't currently figure out the right way to implement them. @jamesturner246  if you have time to review the code that would be great, but if not, no worries.

jamesturner246:
Sure thing, I will have a look through today.

neworderofjamie:
Thank you! Good plan - I'd love to talk you through the current structure/my grand plans some time soon!

PR #162, opened 2017-10-25
C++11 improvements

neworderofjamie:
This came about from an attempt to tidy up the growing number of ``NNmodel::addSynapsePopulation`` methods by moving some of the common code into the ``SynapseGroup`` constructor. However this started crashing which turns out to be because this code was causing temporary objects to be created:
```c++
auto result = m_SynapseGroups.insert(
            pair<string, SynapseGroup>(
                name, SynapseGroup(name, mtype, delaySteps,
                                   WeightUpdateModel::getInstance(), weightParamValues.getValues(), weightVarValues.getValues(),
                                   PostsynapticModel::getInstance(), postsynapticParamValues.getValues(), postsynapticVarValues.getValues(),
                                   srcNeuronGrp, trgNeuronGrp)));
```
Therefore the pointers I was adding to the in and outgoing synapse population lists in the pre and postsynaptic neuron groups were invalid as they pointed to the temporary object. Luckily C++11 adds 'emplace' methods to containers to add objects without temporary object creation however the syntax is a little unpleasant:

```c++
auto result = m_SynapseGroups.emplace(
            std::piecewise_construct,
            std::forward_as_tuple(name),
            std::forward_as_tuple(name, mtype, delaySteps,
                                  WeightUpdateModel::getInstance(), weightParamValues.getValues(), weightVarValues.getValues(),
                                  PostsynapticModel::getInstance(), postsynapticParamValues.getValues(), postsynapticVarValues.getValues(),
                                  srcNeuronGrp, trgNeuronGrp));
```
It works though! To prevent these issues happening again I also deleted the copy constructors in ``SynapseGroup`` (and ``NeuronGroup``). 

I was slightly worried about the compiler compatibility with this as it feels a little bleeding edge but I tested manually on Visual C++ 2013 and the automatic tests cover GCC 4.8, GCC 5 and whatever Clang Thomas's Mac has.

### DIFF
--- a/lib/include/modelSpec.h
+++ b/lib/include/modelSpec.h
@@ -87,6 +87,12 @@ enum FloatType
 class NNmodel
 {
 public:
+    // Typedefines
+    //=======================
+    typedef map<string, NeuronGroup>::value_type NeuronGroupValueType;
+    typedef map<string, SynapseGroup>::value_type SynapseGroupValueType;
+
+
     NNmodel();
     ~NNmodel();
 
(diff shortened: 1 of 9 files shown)


neworderofjamie:
Are you going to get a chance to look at this @jamesturner246? If not - no worries

jamesturner246:
Not sure I understand some of these constructs. I will drop by for a chat now, if that's OK?

neworderofjamie:
Sure! Any time

PR #163, opened 2017-10-30
More flexible variable initialisation

neworderofjamie:
This is something of a 'stepping stone' towards a generic way of initialising state variables (and eventually connectivity) on the GPU. 

### Code 'snippets'
These are essentially simpler versions of the classes  used to define neuron model etc  and are used to generate (stateless) initialisation code. For example to initialise a state variable with a uniform distribution you might define a snippet like this:

```c++
class Uniform : public InitVarSnippet::Base
{
public:
    DECLARE_SNIPPET(Uniform, 2);

    SET_CODE(
        "const scalar scale = $(max) - $(min);\n"
        "$(value)= $(min) + ($(gennrand_uniform) * scale);");

    SET_PARAM_NAMES({"min", "max"});
};
```

### Initialising state variables
In order to use a snippet to initialise a state variable you first define the parameters for the snippet's parameters:
```c++
InitVarSnippet::Uniform::ParamValues vDist(
    -60.0,  // 0 - min
    -50.0); // 1 - max
```

Snippets using these parameters can now be included in  the ``LIF::VarValues`` object passed to ``NNmodel::addNeuronModel`` mixed in with the constant values previously supported:

```c++
LIF::VarValues lifInit(
    initVar<InitVarSnippet::Uniform>(vDist),    // 0 - V
    0.0);                                       // 1 - RefracTime
```

Beyond saving users from reinventing the wheel every time they want to initialise variables in standard ways, this change already potentially improves a few things:
1. If you want to do it yourself you can indicate that you don't want GeNN to perform any initialisation by using ``InitVarSnippet::Uninitialised`` - previously the constant value would be written whatever.
2. Any state variables associated with ``SynapseMatrixType::SPARSE_INDIVIDUALG`` synapse populations will now be initialised in the ``initMODEL_NAME`` function. This could break assumptions in existing models so this behaviour is toggled by the ``GENN_PREFERENCES::autoInitSparseVars`` option.

I understand if you don't have the time or inclination to look through the changes but your thoughts on the syntax etc would be much appreciated as this is basically the framework I plan on implementing on-GPU connectivity calculation and stuff within.


tnowotny:
Hi Jamie,
I had a look at your example and some of the code. Two questions:
1. What was your thinking to introduce this syntax:
```"$(set_value, value);"```
rather than ```"$(set_value)= value;"``` like we do for ```addtoinsyn``` 

2. I wasn't quite sure why you needed a template class ```varInit``` over a template that is a ```snippet```, initialised with parameters, rather than an instance of the ```snippet``` initialised with parameter values. For the naive eye it looks like a spurious step adding to complexity ... but I fear that I have lost touch with the overall code philosophy a little.

neworderofjamie:
1. It may not be entirely necessary in this context but, the idea with the ``$(set_value, value);`` syntax is that it allows a bit more flexibility in the code generation so, for example, if you need to do something under the hood other than simply assign a variable you wouldn't need to introduce a 2nd stage like the ``$(updatelinsyn);`` bit that's required in weight update models to do the atomic add.
2. There isn't a particular reason other than to make the way it operates as similar as possible to ``NNmodel::addNeuronPopulation`` and friends where you pass parameters and the model type rather than creating instances of models bound to parameters. I agree that the approach you suggest would also be totally valid there as well.

tnowotny:
I can't recall any more why we actually put the ```$(updatelinsyn)``` thing in there. For some  reason Esin thought ```$(addtoinsyn)= something;``` could not immediately translated into both, the explicit action to set ```$(addtoinsyn)``` to ```something``` and doing the neccesary update operation to make it globally available ... memory is failing me.

But I se your point regarding felxibility ... it just doesn't really look like anything familiar.
Coupld of thoughts, maybe it should be ```init_value``` rather than ```set_value```. And why could we not replace ```$(init_value)= something;``` with whatever is needed based on detecting ```$(init_value)``` and ```=``` and ```;``` at the end?

I see your point about the parallel to ```addNeuronPopulation```. It still seems odd to have the  ```VarInitSnippet``` and ```initVar``` ... I guess a minimal improvement I could suggest would be more consistent naming; how about ```initVarSnippet``` and ```initVar```?

Just some thoughts ... overall I am very happy with the direction of travel.

neworderofjamie:
I would imagine it's because substituting ``atomicAdd(&dd_inSynPop[ipost], something);`` in place of ``$(addtoinsyn) = something;`` for all possible forms of ``something`` is a little tricky (you need to count brackets in ``something`` etc). The ``$(functionName, param1, param2)`` syntax was my attempt to solve that  in a general, easy to parse way that extends to functions with multiple parameters (initially used for the random distribution calls). However, in this context, I can't honestly see any reason not to use ``$(value) = something`` which is definitely clearer and more familiar-looking. I'll switch it over.

Agreed! Those are definitely better names - will change those. Glad you're pleased with the direction this is going - I was pleased with flexibility and GeNN-likeness of this.

tnowotny:
Are you happy with it? From my side I am happy to merge I think ...

neworderofjamie:
Great! If you're happy; and both me and Jenkins are happy - I'll merge :)

PR #164, opened 2017-11-03
Glibc bug warning

neworderofjamie:
Based on discussions with Marcel the best solution and most general solution I can think of for GeNN in general is a runtime check. As GeNN isn't set up to static link Glibc the only solution is to set the LD_BIND_NOW environment variable at runtime so, if we're building with G++ (seemingly clang sets this flag too so we have to explicitly test NOT clang) on x86_64 and linking against one of the effected Glibc versions we perform a runtime check on the LD_BIND_NOW environment variable in the initialize function.

Machines tested:

- [x] x86_64 Ubuntu 16 desktop - warning should be shown if LD_BIND_NOW isn't set
- [x] Jetson TX1 Ubuntu 16 - warning shouldn't be shown because architecture isn't correct (aarch64)
- [x] x86_64 Windows machine - warning shouldn't be shown 
- [x] x86_64 Ubuntu 14 laptop - warning shouldn't be shown because Glibc is older (2.19)
- [x] Mac OSX machine using Clang - warning shouldn't be shown because alternative libc is used

### DIFF
--- a/lib/src/generateInit.cc
+++ b/lib/src/generateInit.cc
@@ -115,7 +115,7 @@ void genInit(const NNmodel &model,          //!< Model description
     os << "#if defined(__GNUG__) && !defined(__clang__) && defined(__x86_64__) && __GLIBC__ == 2 && (__GLIBC_MINOR__ == 23 || __GLIBC_MINOR__ == 24)" << std::endl;
     os << "    const char *ldBindNow = std::getenv(\"LD_BIND_NOW\");" << std::endl;
     os << "    if(ldBindNow == NULL || strcmp(ldBindNow, \"1\") != 0) {" << std::endl;
-    os << "        fprintf(stderr, \"Warning: a bug has been found in glibc 2.23 or glibc 2.24 (https://bugs.launchpad.net/ubuntu/+source/glibc/+bug/1663280)\"" << std::endl;
+    os << "        fprintf(stderr, \"Warning: a bug has been found in glibc 2.23 or glibc 2.24 (https://bugs.launchpad.net/ubuntu/+source/glibc/+bug/1663280) \"" << std::endl;
     os << "                        \"which results in poor CPU maths performance. We recommend setting the environment variable LD_BIND_NOW=1 to work around this issue.\\n\");" << std::endl;
     os << "    }" << std::endl;
     os << "#endif" << std::endl;
--- a/lib/src/generateInit.cc
+++ b/lib/src/generateInit.cc
@@ -7,8 +7,6 @@
 #include <cmath>
 #include <cstdlib>
 
-
-
 // GeNN includes
 #include "codeStream.h"
 #include "global.h"
@@ -113,8 +111,7 @@ void genInit(const NNmodel &model,          //!< Model description
     // **NOTE** if we are using GCC on x86_64, bugs in some version of glibc can cause bad performance issues.
     // Best solution involves setting LD_BIND_NOW=1 so check whether this has been applied
     os << "#if defined(__GNUG__) && !defined(__clang__) && defined(__x86_64__) && __GLIBC__ == 2 && (__GLIBC_MINOR__ == 23 || __GLIBC_MINOR__ == 24)" << std::endl;
-    os << "    const char *ldBindNow = std::getenv(\"LD_BIND_NOW\");" << std::endl;
-    os << "    if(ldBindNow == NULL || strcmp(ldBindNow, \"1\") != 0) {" << std::endl;
+    os << "    if(std::getenv(\"LD_BIND_NOW\") == NULL) {" << std::endl;
     os << "        fprintf(stderr, \"Warning: a bug has been found in glibc 2.23 or glibc 2.24 (https://bugs.launchpad.net/ubuntu/+source/glibc/+bug/1663280) \"" << std::endl;
     os << "                        \"which results in poor CPU maths performance. We recommend setting the environment variable LD_BIND_NOW=1 to work around this issue.\\n\");" << std::endl;
     os << "    }" << std::endl;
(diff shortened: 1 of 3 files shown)


neworderofjamie:
@mstimberg - your thoughts would also be appreciated on this solution!

tnowotny:
Generally agree but worry: will the warning be visible enough to any user to be picked up? We write a lot of stuff on cout/cerr where it could get lost - or not?

neworderofjamie:
That is also my concern but I'm not totally sure what to do about it - it being a Linux issue we could make it red :)

tnowotny:
What would the downside be of just setting the environment variable to 1 if we see it's appropriate? We could warn that that is what we are doing?

tnowotny:
I see your point - people could just build it and take it away ... in the original Brian thread wasn't there an alternative solution where an argument was used in the makefile instead? I can't remember the details ...

mstimberg:
Hi, yes I see the problem. In Brian2GeNN, we call `./main` directly from Python, so that's where we set the environment variable. If someone would use the generated code directly without passing through Python, they'd have the same problem you are describing.
The only alternative I was mentioning was to compile things statically, but I think that's not an option for you.
Another alternative I could think of (not sure that's very elegant): the file that is generated for the user to run (`generate_run` I think?) is actually a shell script that sets `LD_BIND_NOW` and then calls the actual executable.

neworderofjamie:
@mstimberg - ``generate_run`` is just something Thomas tends to use in his models. Personally I tend to run them directly with ``./simulator`` or whatever which doesn't give a lot of scope for injecting environment variables

neworderofjamie:
We could set it in ``genn-buildmodel`` as Thomas suggests  **and** add the run-time warning I've added in this change. That should help the bulk of people never see the warning.

tnowotny:
I think that is a good idea.

lib/src/generateInit.cc
```diff
@@ -107,6 +110,16 @@ void genInit(const NNmodel &model,          //!< Model description
     std::string oB = "", cB = "";
 #endif // _WIN32
 
+    // **NOTE** if we are using GCC on x86_64, bugs in some version of glibc can cause bad performance issues.
+    // Best solution involves setting LD_BIND_NOW=1 so check whether this has been applied
+    os << "#if defined(__GNUG__) && !defined(__clang__) && defined(__x86_64__) && __GLIBC__ == 2 && (__GLIBC_MINOR__ == 23 || __GLIBC_MINOR__ == 24)" << std::endl;
+    os << "    const char *ldBindNow = std::getenv(\"LD_BIND_NOW\");" << std::endl;
+    os << "    if(ldBindNow == NULL || strcmp(ldBindNow, \"1\") != 0) {" << std::endl;
```
mstimberg: A minor point: while I think it's customary to set `LD_BIND_NOW` to 1, from `ld`'s point of view it only has to be present, i.e. any value is fine.

neworderofjamie: Excellent point! Do you have any thoughts on how safe ``LD_BIND_NOW=1`` is to set on a user's environment without their explicit say so?

mstimberg: You mean permanently instead of just for the run of GeNN? I think it could lead to a noticeable slowdown of program starts (never really tried this myself) because all libraries are loaded at the beginning instead of when needed. Here's a quote from the [Gentoo wiki](https://wiki.gentoo.org/wiki/Hardened/Position_Independent_Code_internals#Using_prelink_and_LD_BIND_NOW) (not sure whether this still applies on modern "fast" machines):
> But you should be careful and know that using LD_BIND_NOW is not recommended on machines where responsiveness is an issue, clicking on an icon in KDE or GNOME and waiting 20 seconds for evolution to start is sometimes inacceptable by users. In doubt, use prelink! 

Using `prelink` could actually be another option (never used it, though), but I don't think it is installed by default, so not ideal either...

neworderofjamie: That was my fear - it wouldn't be great if genn-buildmodel set the environment variable and then the user wonders why everything launched from that terminal runs really slow

tnowotny:
ok - this doesn't leave us much room. To have discussed all options - how about static linking in the case where it's necessary? I don't think it is completely impossible, or is it?

neworderofjamie:
Frankly no :/ I briefly tried prelinking and it looks like CUDA will make that, as well as statically linking, fail.

mstimberg:
Still wondering about the wrapper script. @neworderofjamie: When you say you run your simulations directly, does that mean that your projects to not make use of `userproject/includes/makefile_common_gnu.ml`? Because if they did, you could change the `$(EXECUTABLE)` target so that it creates a shell script called `$(EXECUTABLE)` (so from the user side, nothing would change, `./simulator` would still run it), alongside the real executable named, say, `real_$(EXECUTABLE)`, called by the shell script after setting `LD_BIND_NOW`. Or am I missing something?

neworderofjamie:
Hahah that is hacky but we could indeed do that! I will attempt it on Monday

mstimberg:
It's not even *that* hacky - my `/usr/bin` is full of wrapper scripts around executable binaries :)

neworderofjamie:
@tnowotny it would be great if you could give this a go on your Linux machine and your Mac (to check it does nothing)

tnowotny:
runs fine on Mac and does nothing as expected. Was not able to test on the linux box yet as it is in the middle of a large benchmark run

neworderofjamie:
Ok great, unless you or @mstimberg  have any further thoughts on this, feel free to merge

mstimberg:
All looking good to me (didn't try it out, though), the only minor nitpick is that the warning still seems to explicitly test for `LD_BIND_NOW=1` instead of just `LD_BIND_NOW`.

neworderofjamie:
Oh yeah - totally forgot about that - will change the code accordingly this afternoon and merge.

PR #165, opened 2017-11-30
New build system for Windows

neworderofjamie:
Back in the (pre Visual Studio 2010) day - Visual Studio used different build systems for the IDE (whatever black magic happened inside the project files) and the commandline (Nmake). Now these have been unified using a build system called MSBuild which is a bit like Ant and all those Java-centric tools - XML-based and declarative. ANYWAY - using NMake on Windows (and commandline-centric development in general) is a bit painful and it would be *really* nice to be able to write and debug GeNN simulation code within the IDE. This change implements this by:

1. Building ``genn.lib`` using an MSBuild script (lib/lib_genn.vcxproj) (this can b
2. Building ``generateALL.exe`` using an MSBuild script (lib/generate.vcxproj) 
3. Update ``genn-buildmodel.bat`` to do 1 and 2 (this seemed as good a time as any to remove the deprecated ``buildmodel.bat``)
4. Added an equivalent of userproject\include\makefile_common_win.mk for MSBuild (userproject\include\genn.props and userproject\include\genn_cpu_only.props)
5. Updated code generator to generate suitable MSBuild script for building usercode

All of these build steps can be done from the command line e.g.:
```msbuild generate.vcxproj /p:Configuration=Release```
Or from the IDE by double clicking on the vcxproj. ``Release``, ``Debug``, ``Release_CPU_ONLY`` and ``Debug_CPU_ONLY`` configurations are defined for all projects and all the intermediates are kept separate making it less messy to switch build configuration. This is way nicer than using CMake or anything similar as you can e.g. add files to libgenn using the IDE. 

In terms of backward compatibility:
* I don't *THINK* anyone is doing anything directly with the NMake based building of libgenn so I have removed it. @tnowotny - Brian2GeNN doesn't do anything cunning here does it?
* I have left the code to generate NMakefiles and userproject\include\makefile_common_win.mk intact so existing projects will build fine (e.g. the userprojects)

Todo:
- [ ] Switch SpineML over - this is definitely something that would benefit from the presence of a debugger!


jamesturner246:
Nice. I had MSbuild in the back of my mind too, but all I knew at the time was make. Would like to see it in action!

neworderofjamie:
Thanks! Come by my office and see the wonder of the Visual Studio debugger!

tnowotny:
Looks like the right thing to do. brian2genn is using the normal userproject\include\makefile_common_win.mk  route as far as I can remember, i.e. as long as the usual userprojects continue to work, brian2genn should as well.
I suppose at some point we will have to upgrade brian2genn to a more recent GeNN release and at that point we also might consider switching this over?

neworderofjamie:
OK cool - I will merge this then!

Issue #149, opened 2017-04-30
Support for -std=c++14 and later


whenov:

jamesturner246:
are you referring to C++14 features?

whenov:
Yes. I was trying to use `make_unique`. It would be nice if user can use the full power of their compiler if they use a newer version of gcc.

jamesturner246:
I know it's been a while, but, just in case you're still wondering, you can manually hack GeNN in the meantime. All you need to do is replace all occurrences of `-std=c++11` to `-std=c++14` to get what you want.

neworderofjamie:
@whenov - not sure if you're still interested but #192 means that you can use whatever C++ standard you would like to build your simulator by setting ``CPP_STANDARD=c++14`` before you include makefile_common_gnu.mk - it's still not possible to use newer C++ features in models so I'm leaving this open

whenov:
@neworderofjamie Thanks for your information!

neworderofjamie:
Plan for final part of this:
1. Don't ``#include`` model (that was gross anyway), build it separately and link it (add an extern declaration somewhere)
2. Add a std parameter to genn-buildmodel.sh which gets passed through to make which uses it as C++ standard for building model

**NOTE** this doesn't effect Windows as VC++ just builds using whatever C++ standard it supports

neworderofjamie:
Sadly 1. is not quite that simple - if you build model separately there's no easy way of accessing GENN_PREFERENCES global. For GeNN 4 will allow the standard which ``genn_generator`` (which is only one, small file to be built with a user-specified standard)

PR #167, opened 2017-12-18
Runner now always gets compiled as C++11

neworderofjamie:
Using the new RNG support previously toggled C++11 compilation (as it used C++11 RNGs for CPU_ONLY mode).  As discussed in #161, this seemed a bit weird - we should be using C++11 to compile generated code as the generator itself requires it *and* the user project makefile also compiles everything else in the user project as C++11.


tnowotny:
Is there an issue where users would want something even newer than C++11?

neworderofjamie:
There is the potential for that (issue #149 infact) but I think that's one for another day - it would essentially involve passing a standard through to genn-buildmodel and the userproject makefile.

tnowotny:
Ok I will merge the pull request as I agree with the direction of travel - haven't had the time to review the code but as tests have passed I think it is safe to assume it's fine.

PR #168, opened 2017-12-18
Support for initialising variables on GPU

neworderofjamie:
This is the second part of #163 - allowing variable initialisation code to be run on the GPU. A side effect of this is that you don't need to allocate memory on both host and device for variables you have no interest in reading back - this is potentially a big memory saving for Jetson. This is controlled through an extended version of the API for controlling whether variables are zero-copied or not:
```cpp
auto *e = model.addNeuronPopulation<LIF>("E", Parameters::numExcitatory, lifParams, lifInit);

e->setVarMode("V", VarMode::LOC_DEVICE_INIT_DEVICE);
```
``VarMode`` is an enum that defines all the supported variable modes:
```cpp
enum class VarMode : uint8_t
{
    LOC_DEVICE_INIT_DEVICE,
    LOC_HOST_DEVICE_INIT_HOST,
    LOC_HOST_DEVICE_INIT_DEVICE,
    LOC_ZERO_COPY_INIT_HOST,
    LOC_ZERO_COPY_INIT_DEVICE,
};
```
These are bitfields with fields specifying where a variable should be located i.e. on the host, device or both and where initialisation code should be run. The default value (this can be set with ``GENN_PREFERENCES::defaultVarMode``) for this is  ``LOC_HOST_DEVICE_INIT_HOST`` which matches current behaviour.

One caveat with this change is that, models with variables set as ``LOC_DEVICE_INIT_DEVICE``, cannot be simulated on the CPU. 

## Sub-tasks
- [x] Variable mode correctly handled in existing runner generation code
- [x] Host initialisation not performed on device-initialised variables
- [x] GPU initialisation of GPU-initialised variables
- [x] Add variable mode to synapse group's ``inSyn`` variable
- [x] Check ``copyStateToDevice`` is not copying state already initialised on device
- [x] Extend timing system to time initialisation kernels
- [x] Correctly setup Phillox RNG for GPU-initialising variables with probabilistic distributions
- [x] Measure performance on a few device with 1 thread per synapse and 1 thread per post neurons


neworderofjamie:
Some preliminary results:

# Quadro K2000
desktop_gpu

# Jetson TX1
jetson_gpu

tnowotny:
These benchmarks look extremely promising. But why do the checks fail?

neworderofjamie:
@tnowotny - assuming the tests don't fail this is ready to go! Having updated various models to use it and typical usage is illustrated in https://github.com/neworderofjamie/genn_examples/blob/master/va_benchmark/model.cc - basically you set:
 ``GENN_PREFERENCES::defaultVarMode = VarMode::LOC_DEVICE_INIT_DEVICE;`` 
meaning you save memory for internal state by not allocating it on host at all and then just set variables you wish to record so that memory is allocated on host so you can pull them e.g. ``e->setSpikeVarMode(VarMode::LOC_HOST_DEVICE_INIT_DEVICE);``

PR #169, opened 2018-01-03
Option to pass include flags from genn-buildmodels - allow models to include headers from other directories

neworderofjamie:
Changes in master to use += to append to ``CXXFLAGS`` etc in GNUmakefile didn't SEEM to work from command line without ``override`` directive - am I being stupid @tnowotny or was this intended for another purpose?

### DIFF
--- a/lib/GNUmakefile
+++ b/lib/GNUmakefile
@@ -55,39 +55,39 @@ ifeq ($(DARWIN),DARWIN)
     CXX                  :=clang++
 endif
 ifndef CPU_ONLY
-    CXXFLAGS             :=-std=c++11 -Wall -Wextra -DNVCC=\"$(NVCC)\"
+    override CXXFLAGS             +=-std=c++11 -Wall -Wextra -DNVCC=\"$(NVCC)\"
 else
-    CXXFLAGS             :=-std=c++11 -Wall -Wextra -DCPU_ONLY
+    override CXXFLAGS             +=-std=c++11 -Wall -Wextra -DCPU_ONLY
 endif
 ifdef DEBUG
-    CXXFLAGS             +=-g -O0 -DDEBUG
+    override CXXFLAGS             +=-g -O0 -DDEBUG
 endif
 
 ARFLAGS                  :=-rcs
 
 # Global include and link flags
 ifndef CPU_ONLY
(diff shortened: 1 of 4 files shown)


tnowotny:
Oh dear - I am no Makefile guru ... never seen the ```override``` directive before. I think @jamesturner246 should comment.

neworderofjamie:
So what was the intended purpose of 7d77583f543d51defc79f436596431186688784b? How does Brian2GeNN interact with genn-buildmodel?

jamesturner246:
My understanding is that, if you set any makefile macros at the command line, such as `make MYMACRO=myval`, then any assignments to `MYMACRO` in the actual makefile are ignored - they are overridden by the `make` argument. 

The only way to overwrite or add to a macro inside a makefile, that was overriden by the `make` argument, is to use the override directive.

jamesturner246:
I see you are now overriding `INCLUDE_FLAGS` at the command line. So adding `override` before all the `INCLUDE_FLAGS` assignments in the makefile is correct.

neworderofjamie:
Thanks @jamesturner246  - my make knowledge comes largely from copy-and-paste + stack overflow so that's good to know. @tnowotny my only concern is - how were you overriding these variables in Brian2GeNN?

tnowotny:
I think we never needed to override things ... we added some values up front, the common makefile added to this; the makefile template for brian2genn looks like this:
```
ROOTDIR         := {{ROOTDIR}}
EXECUTABLE      := main
{% if not use_GPU %}
CPU_ONLY            = 1
{% endif %}
SOURCES         := main.cpp {% for source in source_files %} {{source}} {% endfor %} brianlib/randomkit/randomkit.cc

INCLUDE_FLAGS   := -I. -Ibrianlib/randomkit -Lbrianlib/randomkit
OPTIMIZATIONFLAGS := {{compiler_flags}}
NVCC_OPTIMIZATIONFLAGS := {{nvcc_compiler_flags}}
SIM_CODE := {{model_name}}_CODE

include $(GENN_PATH)/userproject/include/makefile_common_gnu.mk```
which seems to have done the trick.

tnowotny:
Note: this is not using command line arguments but merges the two makefiles through the ```include``` directive. That could be the difference.

neworderofjamie:
Ahhh - thanks @tnowotny! Using ``include`` does indeed mean you don't need to use override.

PR #170, opened 2018-01-26
Update userprojects

neworderofjamie:
The plan with this PR is to tidy up the user projects a bit and, in the process, update them so they use any new GeNN features which are applicable:
### HHVclampGA_project
- [x] Didn't work with recent changes so tidied up - my knowledge of this model is pretty minimal though so could you check @tnowotny 
- [x] New windows build system
### Izh_sparse_project
- [x] Tidy
- [x] Use new variable initialisation system for randomising neuron params
- [x] New windows build system
### MBody1_XXXX
- I've updated the main version, but I don't feel there's a lot of purpose in keeping more than one of these here as 1) some of them are variants we added at Capocacia which I don't feel belong here
and b) some like the 'userdef' one seem to exist to demonstrate stuff that's no longer relevant
### Model_Schmuker_2014_classifier_project
- Probably going to leave this one alone!
### OneComp_project
- [x] Tidy
- [x] New windows build system
### PoissonIzh_project
- [x] Tidy
- [x] Use new simple Poisson source
- [x] New windows build system
### SynDelay_project
- [x] Tidy
- [x] New windows build system


neworderofjamie:
Ok the userprojects are now a little tidier, work with the current version of GeNN and build using the new build system on Windows. The only question is regarding the mushroom bodies - ``MBody_map_project`` and ``MBody_map_robot1_project`` should definitely be deleted as they're relics of last Capocaccia, but I'd like to hear your thoughts on  ``MBody_delayedSyn_project``, ``MBody_individualID_project`` and ``MBody_userdef_project``!

tnowotny:
```MBody_map_project``` and ```MBody_map_robot1_project``` should definitely be deleted ... agreed, though should I keep a copy in a personal repo I wonder?
```MBody_delayedSyn_project``` - was our first larger network model with synapses having delay; but maybe not particularly informative beyond what the ```SynDelay_project``` does ... remove?
```MBody_individualID_project``` - this is the only example where we use the (now somewhat ancient) individualID synapses, i.e. dense synaptic connections where all connections have the same weight and who is connected to whom is stored in a matrix of single bits. I don't even know whether it still works, it had quite a few if-then specific branches in the code generation to do this one.
```MBody_userdef_project``` - Esin created this one to demonstrate how to define everything from scratch as opposed to using predefined models. This is now pretty much how we always do things ... is it still informative to see ```MBody1_project``` versus ```MBody_userdef_project``` in comparison?

neworderofjamie:
It is hard to actually ever delete anything from git but yes, putting a copy of the Capocaccia models somewhere is probably a good idea. 

I'm not really super-convinced that the userdef model demonstrates anything useful any more **but** that is a good point about ``MBody_individualID_project``, I think there's some room for optimisations in how it's implemented, BUT ``SynapseMatrixType::BITMASK_GLOBALG`` as it's now catchingly called is actually pretty optimal in terms of memory:
bitfield

How about I just make synaptic delays and bitmask connectivity options for one ultimate MBody1 model? I've looked at the diffs of the current models and that would trivial (and would definitely help future maintenance of these)

tnowotny:
ok ... good plan. I am thrilled to see that my efforts with the bitmask weren't complete rubbish back then (nice name by the way).

tnowotny:
I have made copies of the Capocaccia projects (in a private repo within BoB organisation for now)

neworderofjamie:
Ok - I'm pretty happy with the MBody1 to end all MBody1s! If you have time to **actually** check it's working (rather than my check that all variants result in approximately the same number of spikes) that would be awesome but, if not, just let me know your thoughts.

tnowotny:
I will try to run a real test on whether it does something reasonable neuroscience-wise ... hold on for a moment.

PR #174, opened 2018-02-12
Improved numerical method

neworderofjamie:
In order to optimally set max connections for SparseProjections, we need to calculate the maximum row length for fixed probability connections. Using the inverse CDF of the binomial distribution works nicely but, I found when simulating the Potjans Diesmann microcircuit, that calculating the CDF by summing binomial coefficients failed for large n. Re-reading numerical recipes told me that using the incomplete beta function is the correct approach. 

As Alex has found another issue with SpineML I'd like to fix before I release I thought I'd fix this in SpineML too!

### DIFF
--- a/spineml/generator/connectors.cc
+++ b/spineml/generator/connectors.cc
@@ -23,81 +23,108 @@
 // Anonymous namespace
 namespace
 {
-// Calculate log factorial using lookup table and log gamma function from standard library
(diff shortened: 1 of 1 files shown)


tnowotny:
If based on numerical recipes and all tests pass it must be right (you have tested it on the large Potjans -Diesmann model I am sure)

PR #179, opened 2018-03-02
Sanity fix

neworderofjamie:
I saw this when I was trying to do some stuff with bitmask connectivity and, while the origins of this code are lost in the mists of time beyond the reach of even git blame, you copied it around so you get this PR :)
```cpp
int UIntSz = sizeof(unsigned int) * 8;
int logUIntSz = (int) (logf((float) UIntSz) / logf(2.0f) + 1e-5f);
```
HOWEVER:
1. These values are used exclusively for indexing connectivity bitmasks which don't use ``unsigned int`` at all, but ``uint32_t``. Therefore ``UIntSz`` **should** be 32 by definition whereas ``unsigned int``doesn't actually have a guaranteed size (although I think we'd be very unlucky to locate a device that can run GeNN but has non-32-bit integers...)
2. Any compiler will replace ``gid / 32`` with the appropriate shift operation and it saves the floating point log-horror (C++ does also provide ``log2``!)

### DIFF
--- a/lib/src/generateCPU.cc
+++ b/lib/src/generateCPU.cc
@@ -49,8 +49,6 @@ void generate_process_presynaptic_events_code_CPU(
     const string &ftype)
 {
     bool evnt = postfix == "Evnt";
-    int UIntSz = sizeof(unsigned int) * 8;
-    int logUIntSz = (int) (logf((float) UIntSz) / logf(2.0f) + 1e-5f);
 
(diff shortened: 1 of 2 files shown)


tnowotny:
you still have a bug in it ... I think a " is misplaced somewhere.

neworderofjamie:
Oops :blush:

PR #180, opened 2018-03-13
Ragged matrix data structure

neworderofjamie:
Going forwards, there are a number of issues with ``SparseProjection`` as a means of representing sparse matrices. 

1. It is actually very tricky to build in parallel (i.e. on the GPU) - you would have to calculate row lengths using one kernel, then take an exclusive sum (doing this efficiently in parallel is, in itself, an open research problem), then allocate ``ind`` and finally generate synapses
2. It doesn't support different index data types e.g. 16-bit indices could improve performance and save memory (especially on Jetson) if the postsynaptic population has less than 2^16 neurons.
3. You can't 'align' rows in memory so operations are hard to vectorise i.e. working on pairs of 16-bit indices and weights is likely to be the route to maximum throughput on Jetson.
4. Postsynaptic learning requires not just a ``remap`` entry per synapse but also a ``revInd`` - tripling the size of the data structure.
5. Synapse dynamics require an additional ``preInd`` entry per synapse - further increasing the size of the data structure.

This pull request hopefully addresses all of these issues with a new _ragged_ matrix data structure. This basically consists of a ``numPre`` x ``maxConnections`` **row-major** array of postsynaptic indices and a ``numPre`` array of row lengths: For example if ``numPre=2``, ``numPost=10`` and ``maxConnections=4``:

Row length | 0 | 1 | 2 | 3 | 
------------- | --- | --- | --- | --- |
4                 | 2 | 4 | 6 | 9| 
2                 | 6 | 8| X | X | 

If postsynaptic learning is required this can be augmented with a ``maxSourceConnections`` x ``numPost`` **column-major** array of indices into the original array:

Col height | 0 | 0 | 1 | 0 | 1 | 0 | 2 | 0 | 1 | 1 |
------------| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
0               | X | X | 0 | X | 1 | X | 2 | X | 5 | 3 |
1               | X | X | X | X | X | X | 4 | X | X | X |

This PR addresses the points above:
1. Because the data structure is created during ``allocMem`` based on the value of ``maxConnections`` specified in the model, it can be trivially build in parallel using a thread per row.
2. For forward compatibility, the ``RaggedProjection`` structure is templated on the index type to allow future optimisations using smaller types.
3. Padding to support alignment etc could simply be applied to ``maxConnectiosn``. 
4. Because a presynaptic index can be determined by dividing the index stored in the column-wise array by ``maxConnections``,  **two** extra arrays are no longer required. Hopefully this will reduce memory bandwidth and thus improves performance.
5. The need for ``preInd`` **could** be alleviated by using the same postsynaptic parallelism strategy used for handing spikes **but** performance is poor for very sparse matrices (see below) - instead we add a ``synRemap`` array that remaps a single array of synapses into indices into the ``ind`` array


neworderofjamie:
chart 1

Note poor performance of postsynaptically parallelised continuous connections when GPU is not being utilised i.e. with very sparse connectivity where there are few columns

neworderofjamie:
Hey Thomas,

This P.R. forms the basis for a lot of stuff I've been working on for GeNN e.g. on-device connectivity generation. It would be great if you could at least read through the P.R. and give any feedback on the data structure / my reasoning for why it's necessary (although, aside from numerous new tests, there isn't **that** many actual code changes)

Cheers

Jamie

tnowotny:
I noticed that your example matrices in the explanatory text above have indexing mistakes ;-)
When looking at performance, this hasn't bought you any speed-up, but I suppose that was not the main goal here yet. Am I guessing right that the benchmark is not considering building connectivity on the GPU? 
I can't look at the detailed code at the moment but the overall reasoning seems sound as an alternative matrix format. 
As you have added tests and they pass I assume you got the indexing all sorted out correctly, so good to merge from my side.

neworderofjamie:
Good spot - fixed the indexing (I think). The benchmark I included doesn't include initialization on GPU (that's for a subsequent P.R.!) - it's just the simulation time.

PR #182, opened 2018-03-21
Fix sparse synapse pre parallelism and shared memory usage

neworderofjamie:
Basically I was adding some tests to test the new ragged matrix connectivity in presynaptic span mode and realised it was broken for small populations (i.e. when it tries to used shared memory to accumulate postsynaptic input).

Many of the isses that caused this are likely to be the fault of past me but:
- If span type was presynaptic, we were toggling the shared memory based on the source population size which is definitely wrong
- We were never initialising the shared memory array if the span type was presynaptic.
- Previously  when postsynaptic parallelism was used, what happens was:
  ```
  linsyn (register) = dd_inSyn (global memory)
  for each spike block {
    shLg (shared mem) = 0
    **accumulate spikes into shLg**
    **synch**
    linsyn += shLg
    **synch**
  }
  dd_inSyn = linsyn
  ```
  Unless I'm missing something, it's unnecessary to synch all the threads and copy back to linsyn after   every block of spikes and, furthermore, it's unnecessary to involve linsyn at all in the sparse case meaning we can just do:
    ```
  shLg (shared mem) = dd_inSyn (global memory)
  for each spike block {
    **accumulate  spikes into shLg**
  }
  dd_inSyn = shLg
  ```

My only remaining concern is that, with presynaptic parallelism, we should perhaps use shared memory atomics to update the shared memory array as it's definitely possible that two threads could try and add input to the same postsynaptic neuron at once.

### DIFF
--- a/lib/include/synapseGroup.h
+++ b/lib/include/synapseGroup.h
@@ -139,10 +139,6 @@ public:
     //! Get variable mode used by postsynaptic model state variable
     VarMode getPSVarMode(size_t index) const{ return m_PSVarMode[index]; }
 
-    //! Is this synapse group too large to use shared memory for combining postsynaptic output
-    // **THINK** this is very cuda-specific
-    bool isPSAtomicAddRequired(unsigned int blockSize) const;
-
     void addExtraGlobalNeuronParams(std::map<string, string> &kernelParameters) const;
     void addExtraGlobalSynapseParams(std::map<string, string> &kernelParameters) const;
     void addExtraGlobalPostLearnParams(std::map<string, string> &kernelParameters) const;
--- a/lib/src/generateKernels.cc
+++ b/lib/src/generateKernels.cc
@@ -128,11 +128,14 @@ void generatePreParallelisedSparseCode(
             string wCode = evnt ? wu->getEventCode() : wu->getSimCode();
             substitute(wCode, "$(t)", "t");
 
-            if (sg.isPSAtomicAddRequired(synapseBlkSz)) { // SPARSE using atomicAdd
+            // If postsynaptic input should be accumulated directly in global memory
+            if (sg.isPSAtomicAddRequired(synapseBlkSz)) {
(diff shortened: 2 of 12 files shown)


neworderofjamie:
@tnowotny - I've merged the other two more trivial PRs to save your time but this one does need your input!

tnowotny:
Jamie, I think you are right on both accounts:
- making the use of shared memory more efficient by cutting out linsyn looks correct.
- I also think atomics will be required in the presynaptic parallelization case. Do you think using shared memory is even worth it on this occasion?

neworderofjamie:
Great! I think, potentially, presynaptic parallelism into shared memory **could** be good for situations with high fan-in i.e. MBONs so I will add the shared memory atomics and benchmark

neworderofjamie:
@tnowotny - I've done some research and, according to https://devblogs.nvidia.com/gpu-pro-tip-fast-histograms-using-shared-atomics-maxwell/, shared memory atomics are slower than global memory atomics on pre-maxwell devices so I've updated the presynaptically parallelised code to use atomics but disabled this unless it's running on Maxwell or newer

tnowotny:
uhm ... on older cards it will do atomics on global memory, correct?

neworderofjamie:
Yes indeed! That's what the ``shouldAccumulateInSharedMemory`` test aims to achieve

tnowotny:
The tests seem to pass and I think the logic seems sound. What did you get on speed benchmarks?

neworderofjamie:
On the Jetson TX1 (Maxwell with poor memory bandwidth so probably the most optimistic choice) - simulating 20000x 100Hz Poisson spike sources, connected to 10 postsynaptic neurons with 10% sparse connectivity - if we use shared memory you get about a 15% speedup.

That's a somewhat contrived example, but this code path was clearly broken before and now can result in **some** speedup.

PR #191, opened 2018-07-02
Dendritic delay

### LINKED ISSUE #9: Heterogeneous delays
"delay granularity" and individual delays in a group

neworderofjamie:
This turned out reasonably cleanly, in the weight update model you use a new syntax to apply input with dendritic delay (this can come from any source, this model just makes it entirely heterogenous by copying it from a variable):

```cpp
class StaticPulseDendriticDelay : public Base
{
public:
    DECLARE_MODEL(StaticPulseDendriticDelay, 0, 2);

    SET_VARS({{"g", "scalar"},{"d", "uint8_t"}});

    SET_SIM_CODE("$(addToDenDelay, $(g), $(d));\n");
};
```

This makes it totally orthogonal to all the previous functionality I've added for initialising variables etc so it's easy to initialise delay from a (quantised) normal distribution or whatever. Once you've created a synapse population you then set the number of dendritic delay slots:

```cpp
synPop->setNumDendriticDelaySlots(maxDelay)
```

And everything else is opaque to the user! The only thing to note is that this will affect the timings of learning rules with any kind of postsynaptic term, but this is hopefully still useful for static networks - might even make @esinyavuz happy and mean we can close #9!

Regarding the performance issue we talked about last week, switching the reset logic to a separate kernel definitely helped (@tnowotny what do you think about doing this for other reset logic as it's a lot tidier than having it potentially inserted in 3 separate places?). **HOWEVER** it looks like the actual performance problem is due to the summing and decaying of inputs from a large number of incoming synapse populations in each neuron kernel which is dwarfing the time taken to update the neurons. I think the solution is to _merge_ postsynaptic dynamics which have:

1. Same postsynaptic model
2. No postsynaptic model variables (this is conservative - in some cases this could be relaxed)
2. Matching postsynaptic parameter values

So that, in this case, all the excitatory synapse populations would ``atomicAdd`` into one linSyn/dendritic delay ringbuffer and the neuron model would only have an excitatory and inhibitory input to apply and decay. That would definitely be a change for another P.R. though!


lib/include/newWeightUpdateModels.h
```diff
@@ -158,6 +158,32 @@ class StaticPulse : public Base
...
+    " $(addToDenDelay, $(g), $(d));\n\
+    \endcode*/
+class StaticPulseDendriticDelay : public Base
+{
+public:
+    DECLARE_MODEL(StaticPulseDendriticDelay, 0, 2);
+
+    SET_VARS({{"g", "scalar"},{"d", "uint8_t"}});
+
+    SET_SIM_CODE("$(addToDenDelay, $(g), $(d));\n");
```
tnowotny: Hum ... this is the main naming/syntax question I suppose.
It resonates with the ```$(addtoinSyn) = ...``` and ```$(updatelinsyn)```, but
a) should it be called ```$(addtoinSynDelay, x, y)``` ?
b) should we finally also provide the syntax ```$(addtoinSyn, x)```?

neworderofjamie:
It is currently orthogonal to the homogeneous (axonal) delays which I think is no bad thing as I feel axonal delay **may** be a bit more efficient and regarding timing in learning rules, having both may be useful (PyNN at least used to have a parameter which set what fraction of the synaptic delay was axonal and hence how much was dendritic).

I think we should definitely add ``$(addtoinSyn, x)`` and,for the sake of consistency, probably rename the new version to ``$(addtoinSynDelay, x, y)`` - will do both tomorrow

neworderofjamie:
In an attempt to combine consistency with camelCase naming, we now have the following two bits of usercode syntax: ``$(addToInSynDelay, x, y)`` and ``$(addToInSyn, x)``. Once the tests pass I'll merge.

PR #192, opened 2018-07-03
Improvements to *nix userproject Makefile

neworderofjamie:
GeNN is definitely helping my Make skills but this is still definitely the bleeding edge so I'd appreciate your thoughts!

* Allow C++ standard to be specified (necessary when building with bob_robotics)
* Added dependency building to userproject and generated makefile so shouldn't need to make clean all the time
* Recursive sim code make rule should be phony so it always gets called (it's up to the sim code make file whether or not to do anything)

It works on GCC 5.4 and GCC 7.3, but a test on Mac would also be much appreciated! Good test using a model without runner etc e.g. SynDelay_project is:
1. ``genn-buildmodel``
2.  ``make``- check simulator works
3. touch simulator cc file
5. ``make`` - check only simulator.cc gets rebuilt, not generated code
6. touch something in generated code e.g. synapseKrnl.cc
7. ``make`` - check runner.cc gets rebuilt

### DIFF
--- a/lib/src/generateRunner.cc
+++ b/lib/src/generateRunner.cc
@@ -2767,7 +2767,7 @@ void genMakefile(const NNmodel &model, //!< Model description
     // Start with correct NVCC flags to build shared library or object file as appropriate
     // **NOTE** -c = compile and assemble, don't link
     string cxxFlags = GENN_PREFERENCES::buildSharedLibrary ? "-shared -fPIC" : "-c";
-    cxxFlags += " -DCPU_ONLY -std=c++11";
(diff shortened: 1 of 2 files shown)


tnowotny:
Tested this on SynDelay example on MAC (but CPU_ONLY because no NVIDIA GPU). All seems to work as described.
However, I am not sure what SynDelaySim is supposed to do - I looked at the output and it contains NaNs. This is likely a completely different issue.

neworderofjamie:
Good spot! I think this goes back to a question I asked @jamesturner246 about this model when I updated it to the new API - I'll fix it in another PR

tnowotny:
I suspect this one is good to go.

PR #193, opened 2018-07-05
Neuron Vm was driven to NaN in the SynDelay example

tnowotny:
Changed Izhikevich neuron model to not output unrealistically high Voltage values when driven to spike. Also changed the SynDelay project so that neurons are not fried through too strong synapses.


tnowotny:
@neworderofjamie ... feel free to instead fix to shot synapses!
Also, how can code coverage go down so much for a trivial change like this?

neworderofjamie:
We will see when this build finishes but I **think** it was due to your mac still working a previous build and thus not taking part in that one

PR #195, opened 2018-07-09
More precise writing of very large and small floating point values

### LINKED ISSUE #194: Unusual behaviour of std::to_string
we replaced my hand-hacked ```to_string``` function with the C++11 ```std::to_string``` function; however, I am not sure I am happy with its policies on conversion of double/float into strings. 
E.g. it converts ```DBL_MAX``` into
```179769313486231570814527423731704356798070567525844996598917476803157260780028538760589558632766878171540458953514382464234321326889464182768467546703537516986049910576551282076245490090389328944075868508455133942304583236903222948165808559332123348274797826204144723168738177180919299881250404026184124858368.000000```
Maybe it's ok, if the compiler understands it (does it?).
But apparently (according to its description here: https://en.cppreference.com/w/cpp/string/basic_string/to_string)
```1e-40``` is returned as the string ```0.000000```
That is surely not what we want. Do we need to revert to the stringstream- based hack or even sprintf that gives the expected outputs? I didn't see any options to control the output of ```std::to_string```.
Any thoughts, @neworderofjamie ?
Thanks for pointing out the true origin of the problem @mstimberg

### PR DESCRIPTION
**Pull request by neworderofjamie**
Fixes #194
* New ``writePreciseString`` function which sets scientific format and large enough width to print full precision (the same formatting we were previously using in ``value_substitutions``)
* Used this in ``value_substitutions``
* Used this for writing better-formatted ``SCALAR_MIN`` and ``SCALAR_MAX`` values

For double precision this results in 
```c++
#ifndef SCALAR_MIN
#define SCALAR_MIN 2.22507385850720138e-308
#endif
#ifndef SCALAR_MAX
#define SCALAR_MAX 1.79769313486231571e+308
#endif
```

And for single precision:
```c++
#ifndef SCALAR_MIN
#define SCALAR_MIN 1.175494351e-38f
#endif
#ifndef SCALAR_MAX
#define SCALAR_MAX 3.402823466e+38f
#endif
```

### DIFF
--- a/lib/include/codeGenUtils.h
+++ b/lib/include/codeGenUtils.h
@@ -194,7 +194,9 @@ void writePreciseString(std::ostream &os, T value)
     os << value;
 
     // Reset to default formatting
-    os << std::defaultfloat;
+    // **YUCK** GCC 4.8.X doesn't seem to include std::defaultfloat
+    os.unsetf(std::ios_base::floatfield)
+    //os << std::defaultfloat;
 
     // Restore previous precision
     os << std::setprecision(previousPrecision);
--- a/lib/include/codeGenUtils.h
+++ b/lib/include/codeGenUtils.h
@@ -1,6 +1,7 @@
 #pragma once
 
 // Standard includes
+#include <iomanip>
 #include <limits>
 #include <string>
 #include <sstream>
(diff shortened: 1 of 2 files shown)


tnowotny:
That looks like a solid fix - thanks!

PR #196, opened 2018-07-24
add current sources (cherry-picked from python_wrapper branch)

braintimeException:
Formalize current sources as a model.

### DIFF
--- a/lib/include/currentSource.h
+++ b/lib/include/currentSource.h
@@ -1,0 +1,110 @@
+#pragma once
+
+// Standard includes
+#include <map>
+#include <set>
+#include <string>
+#include <vector>
+
+// GeNN includes
+#include "global.h"
+#include "currentSourceModels.h"
+#include "variableMode.h"
+
+//------------------------------------------------------------------------
+// CurrentSource
+//------------------------------------------------------------------------
+class CurrentSource
+{
+public:
+    CurrentSource(const std::string &name, const CurrentSourceModels::Base *currentSourceModel,
+                const std::vector<double> &params, const std::vector<NewModels::VarInit> &varInitialisers, int hostID, int deviceID) :
+        m_Name(name),
+        m_CurrentSourceModel(currentSourceModel), m_Params(params), m_VarInitialisers(varInitialisers),
+        m_VarMode(varInitialisers.size(), GENN_PREFERENCES::defaultVarMode),
+        m_HostID(hostID), m_DeviceID(deviceID)
+    {
+    }
+    CurrentSource(const CurrentSource&) = delete;
+    CurrentSource() = delete;
+
+    //------------------------------------------------------------------------
+    // Public methods
+    //------------------------------------------------------------------------
+    //! Function to enable the use zero-copied memory for a particular state variable (deprecated use NeuronGroup::setVarMode):
+     /*! May improve IO performance at the expense of kernel performance */
+    void setVarZeroCopyEnabled(const std::string &varName, bool enabled)
+    {
+        setVarMode(varName, enabled ? VarMode::LOC_ZERO_COPY_INIT_HOST : VarMode::LOC_HOST_DEVICE_INIT_HOST);
+    }
+
+    //! Set variable mode of neuron model state variable
+    /*! This is ignored for CPU simulations */
+    void setVarMode(const std::string &varName, VarMode mode);
+
+    void initDerivedParams(double dt);
+
+    //------------------------------------------------------------------------
+    // Public const methods
+    //------------------------------------------------------------------------
+    const std::string &getName() const{ return m_Name; }
+
+    //! Gets the current source model used by this group
+    const CurrentSourceModels::Base *getCurrentSourceModel() const{ return m_CurrentSourceModel; }
+
+    const std::vector<double> &getParams() const{ return m_Params; }
+    const std::vector<double> &getDerivedParams() const{ return m_DerivedParams; }
+    const std::vector<NewModels::VarInit> &getVarInitialisers() const{ return m_VarInitialisers; }
+
+    int getClusterHostID() const{ return m_HostID; }
+
+    int getClusterDeviceID() const{ return m_DeviceID; }
+
+    bool isZeroCopyEnabled() const;
+    bool isVarZeroCopyEnabled(const std::string &var) const{ return (getVarMode(var) & VarLocation::ZERO_COPY); }
+
+    //! Get variable mode used by current source model state variable
+    VarMode getVarMode(const std::string &varName) const;
+
+    //! Get variable mode used by current source model state variable
+    VarMode getVarMode(size_t index) const{ return m_VarMode[index]; }
+
+    void addExtraGlobalParams(std::map<std::string, std::string> &kernelParameters) const;
+
+    //! Does this current source require any initialisation code to be run
+    bool isInitCodeRequired() const;
+
+    //! Does this current source require an RNG to simulate
+    bool isSimRNGRequired() const;
+
+    //! Does this current source group require an RNG for it's init code
+    bool isInitRNGRequired(VarInit varInitMode) const;
+
+    //! Is device var init code required for any variables in this current source
+    bool isDeviceVarInitRequired() const;
+
+    //! Can this current source run on the CPU?
+    /*! If we are running in CPU_ONLY mode this is always true,
+        but some GPU functionality will prevent models being run on both CPU and GPU. */
+    bool canRunOnCPU() const;
+
+private:
+    //------------------------------------------------------------------------
+    // Members
+    //------------------------------------------------------------------------
+    std::string m_Name;
+
+    const CurrentSourceModels::Base *m_CurrentSourceModel;
+    std::vector<double> m_Params;
+    std::vector<double> m_DerivedParams;
+    std::vector<NewModels::VarInit> m_VarInitialisers;
+
+    //!< Whether indidividual state variables of a neuron group should use zero-copied memory
+    std::vector<VarMode> m_VarMode;
+
+    //!< The ID of the cluster node which the neuron groups are computed on
+    int m_HostID;
+
+    //!< The ID of the CUDA device which the neuron groups are comnputed on
+    int m_DeviceID;
+};
--- a/lib/include/currentSource.h
+++ b/lib/include/currentSource.h
@@ -18,11 +18,10 @@ class CurrentSource
 {
 public:
     CurrentSource(const std::string &name, const CurrentSourceModels::Base *currentSourceModel,
-                const std::vector<double> &params, const std::vector<NewModels::VarInit> &varInitialisers, int hostID, int deviceID) :
+                const std::vector<double> &params, const std::vector<NewModels::VarInit> &varInitialisers) :
         m_Name(name),
         m_CurrentSourceModel(currentSourceModel), m_Params(params), m_VarInitialisers(varInitialisers),
-        m_VarMode(varInitialisers.size(), GENN_PREFERENCES::defaultVarMode),
-        m_HostID(hostID), m_DeviceID(deviceID)
+        m_VarMode(varInitialisers.size(), GENN_PREFERENCES::defaultVarMode)
     {
     }
     CurrentSource(const CurrentSource&) = delete;
@@ -31,12 +30,6 @@ public:
     //------------------------------------------------------------------------
     // Public methods
     //------------------------------------------------------------------------
-    //! Function to enable the use zero-copied memory for a particular state variable (deprecated use NeuronGroup::setVarMode):
-     /*! May improve IO performance at the expense of kernel performance */
-    void setVarZeroCopyEnabled(const std::string &varName, bool enabled)
-    {
-        setVarMode(varName, enabled ? VarMode::LOC_ZERO_COPY_INIT_HOST : VarMode::LOC_HOST_DEVICE_INIT_HOST);
-    }
 
     //! Set variable mode of neuron model state variable
     /*! This is ignored for CPU simulations */
@@ -55,13 +48,6 @@ public:
     const std::vector<double> &getParams() const{ return m_Params; }
     const std::vector<double> &getDerivedParams() const{ return m_DerivedParams; }
     const std::vector<NewModels::VarInit> &getVarInitialisers() const{ return m_VarInitialisers; }
-
-    int getClusterHostID() const{ return m_HostID; }
-
-    int getClusterDeviceID() const{ return m_DeviceID; }
-
-    bool isZeroCopyEnabled() const;
-    bool isVarZeroCopyEnabled(const std::string &var) const{ return (getVarMode(var) & VarLocation::ZERO_COPY); }
 
     //! Get variable mode used by current source model state variable
     VarMode getVarMode(const std::string &varName) const;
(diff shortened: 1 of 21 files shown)


lib/include/currentSource.h
```diff
@@ -0,0 +1,110 @@
...
+    }
+    CurrentSource(const CurrentSource&) = delete;
+    CurrentSource() = delete;
+
+    //------------------------------------------------------------------------
+    // Public methods
+    //------------------------------------------------------------------------
+    //! Function to enable the use zero-copied memory for a particular state variable (deprecated use NeuronGroup::setVarMode):
+     /*! May improve IO performance at the expense of kernel performance */
+    void setVarZeroCopyEnabled(const std::string &varName, bool enabled)
```
neworderofjamie: There's no need to add these deprecated functions for new models - I plan on removing **a lot** of deprecated stuff in GeNN 4.0.0

lib/include/currentSourceModels.h
```diff
@@ -0,0 +1,114 @@
...
+public:
+    //----------------------------------------------------------------------------
+    // Declared virtuals
+    //----------------------------------------------------------------------------
+    //! Gets the code that defines current injected each timestep 
+    virtual std::string getInjectionCode() const{ return ""; }
+
+    //! Gets code which defines the condition when the should be applied described current source model.
+    /*! This evaluates to a bool (e.g. "t > 20 && t < 40"). */
+    virtual std::string getTimeConditionCode() const{ return ""; }
```
neworderofjamie: I don't think the time condition needs to be separate - I don't think it gains anything over including ``if($(t) > $(startTime) && $(t) <= $(endTime))`` in the injection code.

neworderofjamie: This would require that the injection code wasn't a single statement, but I think that's nicer anyway i.e.:
```c++
if($(t) > $(tStart) && t < $(tStop)) {
    $(injectCurrent, $(amp) * sin($(omega)*$(t) + $(phase)) + $(offset));
}
```
You can then substitute the newer 'function' syntax with (you can also directly inject into ``Isyn``):
```c++
functionSubstitute(code, "injectCurrent", 1, "Isyn += $(0)");
```

lib/include/currentSourceModels.h
```diff
@@ -0,0 +1,114 @@
...
+/*! It has 6 parameters:
+
+    - \c tStart - time at which the current starts [ms]
+    - \c tStop  - time at which the current stops [ms]
+    - \c amp    - amplitude of the sine wave [nA]
+    - \c freq   - frequency of the sine wave [Hz]
+    - \c phase  - phase of the sine wave [rad]
+    - \c offset - offset of the sine wave [nA]
+*/
+class ACSource : public Base
```
neworderofjamie: I think this model is too PyNN-specific to be included in GeNN - I've never really seen it used

lib/include/currentSource.h
```diff
@@ -0,0 +1,110 @@
...
+    const CurrentSourceModels::Base *m_CurrentSourceModel;
+    std::vector<double> m_Params;
+    std::vector<double> m_DerivedParams;
+    std::vector<NewModels::VarInit> m_VarInitialisers;
+
+    //!< Whether indidividual state variables of a neuron group should use zero-copied memory
+    std::vector<VarMode> m_VarMode;
+
+    //!< The ID of the cluster node which the neuron groups are computed on
+    int m_HostID;
```
neworderofjamie: This MPI stuff is a bit half-finished at best (it was the result of a previous GSoC project) but I'm pretty sure current sources shouldn't have a host and device ID - they should **always** be simulated on the same device and host the neuron they're being injecting into is located on

lib/src/generateCPU.cc
```diff
@@ -307,6 +307,35 @@ void genNeuronFunction(const NNmodel &model, //!< Model description
                         }
                     }
 
+                    if (n.second.isInjected()) {
```
neworderofjamie: I think this block is exactly the same (aside from dd_ prefixes and stuff) so could you move it into ``StandardGeneratedSections`` to save duplication between the two CPU and CUDA code generators.

Sorry for how painful adding stuff to the code generators is - a refactor is on its way!

lib/src/generateCPU.cc
```diff
@@ -307,6 +307,35 @@ void genNeuronFunction(const NNmodel &model, //!< Model description
...
+                            VarNameIterCtx scVars(scm->getVars());
+                            DerivedParamNameIterCtx scDerivedParams(scm->getDerivedParams());
+                            ExtraGlobalParamNameIterCtx scExtraGlobalParams(scm->getExtraGlobalParams());
+                            os << "// current source " << sc->getName() << std::endl;
+                            if (!scm->getTimeConditionCode().empty())
+                            {
+                                string tcCode = scm->getTimeConditionCode();
+                                StandardSubstitutions::currentSourceTimeCondition(tcCode, sc,
+                                                    scDerivedParams, scExtraGlobalParams);
+                                os << "if (" << tcCode << ")" << std::endl <<"{" << std::endl;
```
neworderofjamie: As I said earlier, I don't think having a separate time condition is a good idea, but for future reference you don't need to manually insert { and }. You have two options both of which :

Automatic scopes which lets you reflect {}s in the code generator in the generated code:
```c++
{
    CodeStream::Scope b(os);
    os << "magical indented code" << std::endl;
}
```

Slightly less automatic open and closed brackets (for when you have conditionals):
```c++
os << CodeStream::OB(14);
os << "magical indented code" << std::endl
os << CodeStream::CB(14);
```

However, it probably is good to put a ``CodeStream::Scope`` inside the ``for (const auto *sc : scs)`` loop, to prevent variables from different current sources conflicting.

lib/src/standardSubstitutions.cc
```diff
@@ -295,4 +295,39 @@ std::string StandardSubstitutions::initVariable(
...
+    std::string &iCode,
+    const CurrentSource *sc,
+    const VarNameIterCtx &scmVars,
+    const DerivedParamNameIterCtx &scmDerivedParams,
+    const ExtraGlobalParamNameIterCtx &scmExtraGlobalParams,
+    const std::vector<FunctionTemplate> functions,
+    const std::string &ftype,
+    const std::string &rng)
+{
+    substitute(iCode, "$(t)", "t");
```
neworderofjamie: It's always good to substitute in ``$(id)`` as the neuron ID too - if nothing else it lets you hack about indexing extra global variables :smile:

lib/include/modelSpec.h
```diff
@@ -436,6 +437,114 @@ class NNmodel
...
+        auto targetGroup = findNeuronGroup(targetNeuronGroupName);
+        auto source = findCurrentSource(currentSourceName);
+        targetGroup->injectCurrent(source);
+    }
+    
+
+    // PUBLIC STATIC FUNCTIONS
+    //========================
+    
+    static void allocateExtraGlobalParam( void** egp, int size )
```
neworderofjamie: What do these do? There shouldn't really be any CUDA-specific stuff in ``NNmodel``

lib/include/currentSourceModels.h
```diff
@@ -0,0 +1,114 @@
...
+#define SET_TIME_CONDITION_CODE(TIME_CONDITION_CODE) virtual std::string getTimeConditionCode() const{ return TIME_CONDITION_CODE; }
+#define SET_EXTRA_GLOBAL_PARAMS(...) virtual StringPairVec getExtraGlobalParams() const{ return __VA_ARGS__; }
+
+//----------------------------------------------------------------------------
+// CurrentSourceModels::Base
+//----------------------------------------------------------------------------
+namespace CurrentSourceModels
+{
+//! Base class for all current source models
+class Base : public NewModels::Base
```
neworderofjamie: None of the models you've included use them but ``NewModels::Base`` also includes state variable. I think current sources should definitely have these but you aren't declaring, allocating, initialising or substituting these anywhere! The places to look are generateRunner.cc (I'm afraid that's **REALLY** messy) for declaring and allocating; and generateInit.cc for initialising. In both cases I would recommend doing much the same as the code does for postsynaptic model variables

lib/include/modelSpec.h
```diff
@@ -436,6 +437,114 @@ class NNmodel
...
+                                     const typename CurrentSourceModel::VarValues &varInitialisers,
+                                     int hostID = 0, int deviceID = 0)
+    {
+        return addCurrentSource<CurrentSourceModel>(name, CurrentSourceModel::getInstance(), paramValues, varInitialisers, hostID, deviceID);
+    }
+
+    //! Injects an existing current source into an existing neuron group
+    /*! \param currentSourceName string name of the current source
+        \param targetNeuronGroupName string name of the targen neuron group */
+    void injectCurrent(const string &currentSourceName, const string &targetNeuronGroupName)
```
neworderofjamie: I think a better API would be **just** to have an ``addCurrentSource`` method that takes the target neuron group name as a parameter. This API is weird because:

1. You can be left with 'orphan' current sources which don't inject anywhere
2. Especially once you add state variables, it's very ambiguous as to how a current source that injects into multiple, different sized populations works

lib/include/currentSource.h
braintimeException: I found these functions in `NeuronGroup` and I wasn't sure whether leaving them out would break anything.

lib/include/modelSpec.h
braintimeException: Oops they don't really belong to this commit. I added them after the conversation we had about `SpikeSourceArray` where you have mentioned that it would be good to have functions to allocate and free `extraGlobalParam`s

lib/include/currentSourceModels.h
braintimeException: Initially I thought of current sources as code snippets that do not require any variables, which is not always the case...

lib/include/modelSpec.h
braintimeException: the reason I made separate addition and injection functions is the ability to inject a current source into multiple populations. The state vars of current sources could be renamed using both the name of the population that is injected and the name of the current source to disambiguate them. This does not solve the 'orphan' problem though

lib/include/modelSpec.h
neworderofjamie: Ok cool - if you can remove them from this branch,  we can look at this again when it comes to merging the python_wrapper branch

lib/include/currentSource.h
neworderofjamie: No worries - no one could expect you to predict my future GeNN plans or how the undocumented MPI support works :smile:

lib/include/currentSourceModels.h
braintimeException: and the derived parameter `omega` in it also produces warnings due to unused argument `dt` - another reason to drop it :laughing:

lib/include/modelSpec.h
neworderofjamie: Yeah - I think making one function simplifies a whole lot of things and removes a lot of edge cases. If you want to inject into multiple populations I don't see any issue with simply calling ``addCurrentSource`` multiple times.

lib/include/currentSourceModels.h
neworderofjamie: Exactly :smile: 

(if you remove the name of the parameter from the lambda function (or any other function) declaration i.e. ``[](const vector<double> &pars, double){...}`` the warning goes away btw)

lib/include/modelSpec.h
```diff
@@ -460,89 +457,57 @@ class NNmodel
...
+        auto result = m_CurrentSources.emplace(std::piecewise_construct,
+            std::forward_as_tuple(currentSourceName),
+            std::forward_as_tuple(currentSourceName, model,
+                                  paramValues.getValues(), varInitialisers.getInitialisers()));
 
         if(!result.second)
         {
-            gennError("Cannot add a current source with duplicate name:" + name);
-            return NULL;
+            auto source = findCurrentSource(currentSourceName);
```
neworderofjamie: I get what you're trying to do here, but it's a bit inconsistent with the other types of thing you can add to GeNN models. For your use case it should be trivial to make the names unique in PyNN.

lib/src/generateInit.cc
```diff
@@ -358,6 +358,26 @@ unsigned int genInitializeDeviceKernel(CodeStream &os, const NNmodel &model, int
                                 }
                             }
                         }
+                        // Loop through current sources
+                        if (n.second.isInjected()) {
```
neworderofjamie: I don't really think the ``NeuronGroup::isInjected`` method is really necessary at all - if there's nothing in the vector then nothing happens

lib/src/generateCPU.cc
```diff
@@ -307,34 +307,11 @@ void genNeuronFunction(const NNmodel &model, //!< Model description
...
-                                StandardSubstitutions::currentSourceInjection(iCode, sc,
-                                                    scVars, scDerivedParams, scExtraGlobalParams,
-                                                    cpuFunctions, model.getPrecision(), "rng");
-                                os << "    Iinj += " << iCode << ";" << std::endl;
-                            }
-                            if (!scm->getTimeConditionCode().empty()) os << "}" << std::endl;
-                        }
-                    }
+                    // check for current sources and insert code if necessary
+                    StandardGeneratedSections::neuronCurrentInjectionLocalVarDeclare(
```
neworderofjamie: These seem to follow each other in both cases - why not have them as a single function?

lib/src/standardSubstitutions.cc
```diff
@@ -297,37 +297,25 @@ std::string StandardSubstitutions::initVariable(
...
+    const std::string &rng,
+    const std::string &popName)
 {
     substitute(iCode, "$(t)", "t");
-    name_substitutions(iCode, "l", scmVars.nameBegin, scmVars.nameEnd, "");
-    value_substitutions(iCode, sc->getCurrentSourceModel()->getParamNames(), sc->getParams());
-    value_substitutions(iCode, scmDerivedParams.nameBegin, scmDerivedParams.nameEnd, sc->getDerivedParams());
-    name_substitutions(iCode, "", scmExtraGlobalParams.nameBegin, scmExtraGlobalParams.nameEnd, sc->getName());
-    substitute(iCode, "$(Iinj)", "Iinj");
+    name_substitutions(iCode, "l", csmVars.nameBegin, csmVars.nameEnd, cs->getName());
```
neworderofjamie: What's this "1" for?

neworderofjamie:
Ok - all the tests pass, the only issue is that ``StandardSubstitutions::currentSourceInjection`` takes a ``popName`` parameter which isn't used and thus causes a warning

lib/include/modelSpec.h
braintimeException: It's not because of PyNN, but rather to allow injecting one current source in several populations, but I agree, in case of existing current source the API which takes all current source parameters and than ignores them looks weird.

lib/src/generateInit.cc
braintimeException: right

lib/src/standardSubstitutions.cc
braintimeException: I guess it stands for "local"? This "l" is prepended to all neuron vars so I took the same syntax for current sources for the sake of consistency.

lib/include/modelSpec.h
neworderofjamie: Especially if the new one you've added has totally different parameters, just the same name :fearful:

lib/src/standardSubstitutions.cc
neworderofjamie: Duh - of course that's what it's for! Sorry I was being dumb on that one...

neworderofjamie:
@braintimeException I've tidied up some odds and ends with the current sources, fixed some bugs and added a feature test which should boost the code coverage a little. Assuming the test passes on our various build machines I'll add some documentation and merge this branch this afternoon. You should probably merge the changes I've made back into your branch though - as I've renamed the built in GeNN models and removed some might have been relying on (basically you probably should reimplement all the models you're using in Python like the weird AC one)

PR #197, opened 2018-07-25
Regex fix

neworderofjamie:
Aside from hacking at Jenkins to, in turn, hack at Centos to enable a newer version of GCC this fixes the substitution issues found by @mstimberg in https://github.com/brian-team/brian2genn/pull/60. What I have done is:

1. Moved the SpineML regex-based variable replace function into libgenn
2. Used it in ``ensureMathFunctionFtype``
3. Updated regex to treat numbers as legitimate parts of variable names
4. Added tests to test the various cases @tnowotny  and @mstimberg  came up with

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -105,13 +105,7 @@ for(b = 0; b < builderNodes.size; b++) {
                     
                     echo pwd()
                     echo env.PATH;
-                    
-                    // **YUCK** if dev_toolset is in node label - enable it to get newer GCC
-                    if("dev_toolset" in nodeLabel) {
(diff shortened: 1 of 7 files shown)


tnowotny:
I have reviewed. It indeed only needed a small fix. I think this is all good now. I suggest you merge when the last tests have passed.

neworderofjamie:
Great! I was thinking I might rename the function to
``regexVariableSubstitution`` or something as it's not purely a
substitution function.

tnowotny:
Yeah ... that is fine by me ... and no issue with breaking user code as it's purely internal.

tnowotny:
good to go now?

PR #200, opened 2018-08-01
More current source

neworderofjamie:
I started switching some of my models over to using current sources and found a few issues:

1. State variables can be initialised on either the GPU or CPU (largely for CPU_ONLY mode/legacy) - in CPU mode, current source variables weren't initialised
2. Current source variables were being read from global memory into registers, but never being written back if they got updated

I fixed both of these as well as updating one of the tests to cover initialising current source variables.

### DIFF
--- a/lib/src/standardGeneratedSections.cc
+++ b/lib/src/standardGeneratedSections.cc
@@ -95,18 +95,17 @@ void StandardGeneratedSections::neuronSpikeEventTest(
                                                          functions, ftype, rng);
 
         // Open scope for spike-like event test
-        os << CodeStream::OB(31);
+        {
+            CodeStream::Scope b(os);
 
-        // Use synapse population support code namespace if required
-        if (!spkEventCond.second.empty()) {
-            os << " using namespace " << spkEventCond.second << ";" << std::endl;
+            // Use synapse population support code namespace if required
+            if (!spkEventCond.second.empty()) {
+                os << " using namespace " << spkEventCond.second << ";" << std::endl;
+            }
+
+            // Combine this event threshold test with
+            os << "spikeLikeEvent |= (" << eCode << ");" << std::endl;
         }
-
-        // Combine this event threshold test with
-        os << "spikeLikeEvent |= (" << eCode << ");" << std::endl;
-
-        // Close scope for spike-like event test
-        os << CodeStream::CB(31);
     }
 }
 //----------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


lib/src/standardGeneratedSections.cc
```diff
@@ -119,33 +118,32 @@ void StandardGeneratedSections::neuronCurrentInjection(
...
-            string iCode = csm->getInjectionCode();
-            substitute(iCode, "$(id)", localID);
-            StandardSubstitutions::currentSourceInjection(iCode, cs,
-                                csVars, csDerivedParams, csExtraGlobalParams,
-                                functions, ftype, rng);
-            os << iCode << std::endl;
+
+        // Read current source variables into registers
+        for(const auto &v : csVars.container) {
+            os <<  v.second << " l" << v.first << " = " << devPrefix << v.first << cs->getName() << "[" << localID << "];" << std::endl;
```
braintimeException: In theory there can be several current sources per population and if some of their variables have the same name, there will be a collision. For example, if for any reason I want to inject DC into half of the population, and AC into another half, controlling the injection by, say boolean ``applyCurrent`` variable in both cases, then ``applyCurrent`` will be defined twice. Appending name of the current source solves it.

lib/src/standardGeneratedSections.cc
```diff
@@ -119,33 +118,32 @@ void StandardGeneratedSections::neuronCurrentInjection(
...
+        string iCode = csm->getInjectionCode();
+        substitute(iCode, "$(id)", localID);
+        StandardSubstitutions::currentSourceInjection(iCode, cs,
+                            csVars, csDerivedParams, csExtraGlobalParams,
+                            functions, ftype, rng);
+        os << iCode << std::endl;
+
+        // Write updated variables back to global memory
+        for(const auto &v : csVars.container) {
+             os << devPrefix << v.first << cs->getName() << "[" << localID << "] = l" << v.first << ";" << std::endl;
```
braintimeException: Same issue with local variable name as above

lib/src/standardSubstitutions.cc
```diff
@@ -308,7 +308,7 @@ void StandardSubstitutions::currentSourceInjection(
     const std::string &rng)
 {
     substitute(iCode, "$(t)", "t");
-    name_substitutions(iCode, "l", csmVars.nameBegin, csmVars.nameEnd, cs->getName());
+    name_substitutions(iCode, "l", csmVars.nameBegin, csmVars.nameEnd);
```
braintimeException: see above

lib/src/standardGeneratedSections.cc
neworderofjamie: This is why I put all the code for each current source in a scope (that's what the ``CodeStream::Scope b(os);`` bit does) - if you look at the generated code it'll be enclosed in {} so the fact the names are the same doesn't matter.

braintimeException: ahh my bad! Then this PR is good to go.

neworderofjamie: Do you have a fix for the extra global parameters bug we can cherry pick in?

braintimeException: I did not commit it yet, so I think I will simply push it here

braintimeException:
I think, once the tests finish, this PR can be merged

PR #201, opened 2018-08-02
Merging of compatible postsynaptic models

neworderofjamie:
I think I've talked to you about this before but what this change does is, if the GENN_PREFERENCE is set, merge together the postsynaptic models of a neuron population's incoming synapse populations if
1. They use the same model class
2. They have the same maximum number of dendritic delay slots
3. They have no state variables (it's ambiguous as to how these should be initialised)
4. Their parameter and derived-parameter values match

When postsynaptic models are merged:

1. Memory is saved as only one dendritic delay buffer and inSyn is allocated
2. Hopefully the L2 cache is thrashed slightly less
3. The postsynaptic model updating/applying phase of the neuron kernel is much reduced

Overall, on a K40, this change more than halves the time spent by the neuron kernel on the Potjans microcircuit model. It's a little messy but I feel worth it!

### DIFF
--- a/lib/src/neuronGroup.cc
+++ b/lib/src/neuronGroup.cc
@@ -93,7 +93,6 @@ void NeuronGroup::mergeIncomingPSM()
     // Create a copy of this neuron groups incoming synapse populations
     std::vector<SynapseGroup*> inSyn = getInSyn();
 
-    std::cout << "Neuron group: " << getName() << std::endl;
     do
     {
         // Remove last element from vector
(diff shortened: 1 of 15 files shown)


lib/src/neuronGroup.cc
```diff
@@ -88,6 +88,74 @@ void NeuronGroup::calcSizes(unsigned int blockSize,  unsigned int &idStart, unsi
...
+void NeuronGroup::mergeIncomingPSM()
+{
+    // Create a copy of this neuron groups incoming synapse populations
+    std::vector<SynapseGroup*> inSyn = getInSyn();
+
+    // Loop through un-merged incoming synapse populations
+    for(unsigned int i = 0; !inSyn.empty(); i++) {
+        // Remove last element from vector
+        SynapseGroup *a = inSyn.back();
+        inSyn.pop_back();
```
tnowotny: isn't pop_back returning the last element of the vector, i.e. allowing
```SynapseGroup *a = inSyn.pop_back();```
?

neworderofjamie: Sadly it doesn't - it must have for **some** programming language though because in my head it should too

PR #202, opened 2018-08-09
Python wrapper

braintimeException:
The python wrapper or **pygenn** is mostly ready.
### What is supported:

- Define, build, run and record, all without leaving cozy Python shell.
- `GeNNModel` Python helper class is the main proxy for model manipulation and simulation.
- New models (`NeuronModels`, `PostsynapticModels`, `WeightUpdateModels`, `CurrentSourceModels` and `InitVarSnippet`s) can be defined directly in Python. There are also `createCustom*Class` functions to simplify this process.
- Add populations and current sources using `addNeuronPopulation`, `addSynapsePopulation` and `addCurrentSource` functions which closely resemble their GeNN counterparts.
- After the model is defined in can be built and loaded using convenience methods `build` and `load` in `GeNNModel`.
- Input / output works via `numpy` array views which works as a pointer (or IS a pointer as far as I understand) to the underlying C-array. If you want to record, simply copy the contents of such array view elsewhere.
- Can run in both GPU and CPU modes

### How to run:
- install [SWIG](http://www.swig.org/) (tested with version 3.0.12)
- build using `make -f GNUMakefilePyGeNN pygenn`
- add `pygenn` folder to your `PYTHONPATH`
- enjoy

### TODO by @neworderofjamie before merging
- [x] Test on Mac OSX
- [ ] Add build infrastructure for Windows
- [ ] Investigate packaging as prebuild package for pip
- [x] Figure out what to do about testing
- [x] Numpy implementation of connection tuples to sparse matrix.


lib/swig/SharedLibraryModel.h
```diff
@@ -0,0 +1,322 @@
...
+#endif
+            return false;
+        }
+
+    }
+
+    // Retrive symbols to pull/push from/to the device from shared model
+    // and store them in a map for fast lookup
+    void initIO( const std::string &popName,
+                 const std::array<bool, 5> &availableDTypes )
```
neworderofjamie: It is super not-important but you could potentially use a ``std::bitset<5>``  here

lib/swig/SharedLibraryModel.h
```diff
@@ -0,0 +1,322 @@
...
+#endif
+
+    VoidFunction m_AllocateMem;
+    VoidFunction m_Initialize;
+    VoidFunction m_InitializeModel;
+    VoidFunction m_StepTimeGPU;
+    VoidFunction m_StepTimeCPU;
+    VoidFunction m_ExitGeNN;
+    
+    PopIOMap populationsIO;
```
neworderofjamie: This should be ``m_PopulationsIO`` in keeping with the general naming convention.

lib/GNUmakefile
```diff
@@ -22,6 +22,9 @@
 # Include makefile which builds libgenn
 include $(GENN_PATH)/lib/GNUMakefileLibGeNN
 
+# Include makefile which builds python wrapper
+include $(GENN_PATH)/lib/GNUMakefilePyGeNN
```
neworderofjamie: I think it would be nicer if GNUMakefilePyNN **included** GNUMakefileLibGeNN rather than the other way around

lib/GNUMakefilePyGeNN
```diff
@@ -0,0 +1,136 @@
...
+SWIG_SUBMODULES                    :=$(addprefix $(PYGENN_PATH)/_,$(SWIG_SUBMODULES))
+SWIG_SUBMODULES                    :=$(addsuffix .so,$(SWIG_SUBMODULES))
+
+pygenn:
+	$(MAKE) pygenn_pre
+	$(MAKE) pygenn_post
+
+pygenn_pre: prepare_swig_config $(PYGENN_PATH) $(LIBGENN_OBJ_PATH) $(PYGENN_PIC_OBJ)
+
+pygenn_post: $(PYGENN_CUSTOM_OBJ) $(SWIG_SUBMODULES)
```
neworderofjamie: Some comments in this makefile would be much appreciated!

lib/GNUMakefilePyGeNN
```diff
@@ -0,0 +1,136 @@
...
+    endif
+endif
+
+$(info $$PYGENN_PATH = ${PYGENN_PATH})
+
+ifdef MPI_ENABLE
+    PYGENN_LINK_FLAGS           +=$(shell mpiCC -showme:link)
+endif
+
+PYGENN_PIC_OBJ                     :=global modelSpec neuronGroup synapseGroup currentSource initVarSnippet utils codeStream codeGenUtils sparseUtils neuronModels postSynapseModels synapseModels newNeuronModels newPostsynapticModels newWeightUpdateModels currentSourceModels standardSubstitutions standardGeneratedSections generateALL generateCPU generateInit generateKernels generateMPI generateRunner
```
neworderofjamie: If you included GNUMakefileLibGeNN as I suggest below, I'm **fairly** sure this is just a copy of ``LIBGENN_OBJ``. It would be **REALLY** good if we didn't have to maintain that list in more places (having it in the Windows and *nix build system is bad enough)

lib/GNUMakefilePyGeNN
```diff
@@ -0,0 +1,136 @@
...
+
+pygenn_pre: prepare_swig_config $(PYGENN_PATH) $(LIBGENN_OBJ_PATH) $(PYGENN_PIC_OBJ)
+
+pygenn_post: $(PYGENN_CUSTOM_OBJ) $(SWIG_SUBMODULES)
+	$(SWIG) $(SWIG_CPU) -outdir $(PYGENN_PATH) -o $(PYGENN_PATH)/$(SWIG_MODULE_NAME)_wrap.cc $(PYGENN_SWIG_FILE_PATH)
+	$(CXX) $(CXXFLAGS) -shared -fPIC -DSWIG $(PYGENN_PATH)/$(SWIG_MODULE_NAME)_wrap.cc -o $(PYGENN_PATH)/_$(SWIG_MODULE_NAME).so $(PYGENN_PIC_OBJ) $(PYGENN_CUSTOM_OBJ) $(INCLUDE_FLAGS) -I $(PYTHON_INCLUDE) -I $(SWIG_PATH) $(PYGENN_LINK_FLAGS)
+	cp $(SWIG_PATH)/GeNNModel.py $(PYGENN_PATH)/GeNNModel.py
+	cp $(SWIG_PATH)/GeNNGroups.py $(PYGENN_PATH)/GeNNGroups.py
+
+$(PYGENN_PATH)/_StlContainers.so: $(PYGENN_PATH)/stl_containers_wrap.cc
```
neworderofjamie: It would be really good if (after you include GNUMakefileLibGeNN :smile:) you used the ``LIBGENN_PREFIX`` to make a unique object directory for each variant of the modules. Otherwise it's very easy to get into a mess of incompatible bits of CPU_ONLY and normal versions mixed together.

lib/swig/customParamValues.h
```diff
@@ -0,0 +1,37 @@
+#ifndef CUSTOMPARAMVALUES_H
+#define CUSTOMPARAMVALUES_H
+namespace CustomValues
+{
+class ParamValues
+{
+public:
+    template<typename T>
```
neworderofjamie: How come this is templated?

lib/swig/customParamValues.h
```diff
@@ -0,0 +1,37 @@
+#ifndef CUSTOMPARAMVALUES_H
```
neworderofjamie: We're generally **trying** to switch to using ``#pragma once`` rather than old style include guards

lib/swig/generate_swig_config.py
```diff
@@ -0,0 +1,538 @@
+import sys
```
neworderofjamie: What does this file do? It needs a lot of commenting!

lib/include/generateALL.h
```diff
@@ -51,3 +51,9 @@ void chooseDevice(NNmodel &model,       //!< the nn model we are generating code
                   const string &path,   //!< path the generated code will be deposited
                   int localHostID);     //!< ID of local host
 #endif
+
+#ifdef SWIG
```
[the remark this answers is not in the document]
braintimeException: This makes a lot of sense

lib/include/modelSpec.h
```diff
@@ -233,9 +233,11 @@ class NNmodel
                                      const typename NeuronModel::VarValues &varInitialisers,
                                      int hostID = 0, int deviceID = 0)
     {
+#ifndef SWIG
```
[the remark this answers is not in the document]
braintimeException: Initially I had `initGeNN` wrapped, but after careful examination I realized that all it does is initializing legacy models and as you said will most likely be deprecated. I thought it was a good idea not to include it in SWIG, but I can revert it back.

lib/swig/SharedLibraryModel.h
braintimeException: :+1:

lib/GNUMakefilePyGeNN
```diff
@@ -0,0 +1,136 @@
...
+##   initial version: 2018-05-14
+##
+##--------------------------------------------------------------------------
+
+SWIG                               :=swig -c++ -python
+SWIG_MODULE_NAME                   :=pygenn
+SWIG_PATH                          :=$(GENN_PATH)/lib/swig
+PYGENN_PATH                        :=$(GENN_PATH)/lib/pygenn
+PYGENN_SWIG_FILE_PATH              :=$(GENN_PATH)/lib/$(SWIG_MODULE_NAME).i
+PYTHON_INCLUDE                     ?=$(shell python-config --includes)
```
neworderofjamie: For me, this returns: ``-I/usr/include/python2.7 -I/usr/include/x86_64-linux-gnu/python2.7`` so when you do ``-I $(PYTHON_INCLUDE)`` lower down it fails

neworderofjamie: (it builds if I override it to ``/usr/include/python2.7`` from the command line though)

lib/include/modelSpec.h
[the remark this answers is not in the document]
neworderofjamie: I think it's better to include it (or at least just add a hack that sets ``GeNNReady= 1``) than to scatter #ifdefs everywhere

lib/GNUmakefile
braintimeException: GNUMakefilePyNN implicitly includes GNUMakefileLibGeNN by coming after it. However many variables are not very helpful because they are libgenn specific. For example, `LINK_FLAGS` contain `-lgenn` which is not needed and not built if the target is `pygenn`.

lib/GNUMakefilePyGeNN
braintimeException: it is a copy of original `LIBGENN_OBJ` however the `LIBGENN_OBJ` are already fully qualified paths by this point.

lib/GNUMakefilePyGeNN
braintimeException: I will try to adapt both make files to minimize redundancy

lib/GNUmakefile
neworderofjamie: I know, but the point of ``GNUMakefile`` to to build generateAll (which admittedly may not be clear from it's current name...) and generateAll links to libgenn so its makefile includes ``GNUMakefileLibGeNN``. Likewise the point of ``GNUMakefilePyGeNN`` is to build PyGeNN so it should include ``GNUMakefileLibGeNN`` rather than be part of ``GNUMakefile``

neworderofjamie: How come pygenn doesn't link against libgenn?

lib/swig/customParamValues.h
braintimeException: Well, it is just a `Snippet::ValueBase` with a relaxed constructor... And yes, the template here is absolutely unnecessary :sweat_smile:

neworderofjamie: Hahah

lib/GNUMakefilePyGeNN
neworderofjamie: Yeah I'd be more than happy if you made GNUMakefileLibGeNN more flexible i.e. by keeping the unqualified paths separate.

lib/GNUMakefilePyGeNN
braintimeException: Right... Python 2.7 is still the standard on many systems, isn't it?

lib/swig/generate_swig_config.py
braintimeException: it generates interface files for SWIG. Some interfaces require many templates specializations and the interfaces for models look alike. So this file does all the copy-paste work :smiley: Additionally it reads the models files and creates template specializations for them, so when a new model is added to GeNN, it will wrap it without any additional work.

neworderofjamie:
Ahh is that what it is! But yes, ubuntu 16 LTS is gonna stay supported for
a good long time

lib/GNUMakefilePyGeNN
braintimeException: Oh, so the issue is not the version, but the second include? I also have two, but they are the same on my system... I guess, taking only the first one should solve this.

neworderofjamie:
Yeah - python-config returns the -Is as well rather than just a path
, but you then add a - I yourself to the command line which breaks it

lib/GNUMakefilePyGeNN
braintimeException: Hah indeed I do add `-I`... I wonder why it works :thinking:

braintimeException: By the way I cannot guarantee that Python 2.7 will work...

neworderofjamie: I still don't really think restricting stuff to Python 3 is a legitimate option - a **lot** of people still use 2.7

lib/swig/generate_swig_config.py
neworderofjamie: It would also be really nice if this stuff generated pydoc help - especially as someone who doesn't know how it works ``help(pygenn.NNmodel.addNeuronPopulation)`` gives me no clues as to what parameters are

braintimeException: actually everything is supposed to work via GeNNModel class

lib/GNUMakefilePyGeNN
braintimeException: I will test it with 2.7.

lib/swig/generate_swig_config.py
neworderofjamie: that explains why **that** wasn't working then :smile:

lib/swig/GeNNGroups.py
```diff
@@ -0,0 +1,477 @@
...
+
+        Args:
+        values -- iterable or single value or parameter space is initVar is specified
+
+        Keyword args:
+        initVar -- type as string or instance of a class derived from InitVarSnippet.Custom
+        """
+        self.initRequired = False
+        if initVar is not None:
+            setRandomInit( initVar, values )
```
neworderofjamie: Where is this function located and what does it do? Are InitVarSnippets implemented or should that be added to the Todo?

lib/swig/GeNNModel.py
```diff
@@ -0,0 +1,934 @@
...
+        pathToModel -- path where to place the generated model code. Defaults to the local directory.
+        """
+
+        if self._built:
+            raise Exception("GeNN model already built")
+        self._pathToModel = pathToModel
+
+        for popName, popData in self.synapsePopulations.items():
+            if popData.sparse:
+                popData.pop.setMaxConnections( popData.size )
```
neworderofjamie: For sparse projections, max connections means (apologies it is badly named) the largest number of postsynaptic neurons any presynaptic neuron is connected to i.e. the longest length of a sparse synaptic matrix row - on that basis I'm not sure if this is correct.

lib/swig/GeNNGroups.py
braintimeException: `InitVarSnippet`s were the last thing I implemented (actually just yesterday) and are not thoroughly tested, but they hopefully work. `self.initRequired`'s purpose is to indicate that a list of values was provided and if it is true, these values will be pushed to device when the model is loaded.

lib/swig/GeNNModel.py
braintimeException: oh. I clearly misunderstood the name.

lib/swig/GeNNModel.py
```diff
@@ -0,0 +1,934 @@
...
+                        sa createCustomNeuronClass
+        paramSpace  -- dict with param values for the NeuronModels class
+        varSpace    -- dict with initial variable values for the NeuronModels class
+        """
+        if self._built:
+            raise Exception("GeNN model already built")
+        if popName in self.neuronPopulations:
+            raise ValueError( 'neuron population "{0}" already exists'.format( popName ) )
+       
+        nGroup = NeuronGroup( popName )
```
neworderofjamie: This applies to all three types of group but I don't quite understand the rationale here. How come you re-implement all the properties of the C++ ``NeuronGroup`` in Python and then immediately duplicate them into a newly created C++ ``NeuronGroup``? Is this so you can defer the creation of the C++ object in future? If not, can't you just call into the wrapped C++ objects?

lib/swig/GeNNGroups.py
```diff
@@ -0,0 +1,477 @@
...
+        else:
+            try:
+                iter( values )
+                self.initVal = pg.uninitialisedVar()
+                self.values = list( values )
+                self.initRequired = True
+            except TypeError:
+                self.initVal = pg.NewModels.VarInit( values )
+
+class ModelPreprocessor(object):
```
neworderofjamie: Static classes aren't really very Pythonic - it would be much nicer just to move this into a module called model_preprocessor and call into that.

lib/swig/GeNNGroups.py
neworderofjamie: Cool that they're implemented! What I meant though is where is ``setRandomInit``? I can't find it in the module...

lib/swig/GeNNModel.py
```diff
@@ -0,0 +1,934 @@
+from os import path
+from subprocess import check_call
+import json
```
neworderofjamie: Is this used?

lib/swig/GeNNGroups.py
braintimeException: oops that's a typo. It must read `self.setInitVar`.

lib/swig/GeNNModel.py
```diff
@@ -0,0 +1,934 @@
...
+            self._T = self._slm.assignExternalPointerSingle_f('t')
+        if self._scalar == 'double':
+            self.T = self._slm.assignExternalPointerSingle_d('t')
+        if self._scalar == 'long double':
+            self._T = self._slm.assignExternalPointerSingle_ld('t')
+        self._TS = self._slm.assignExternalPointerSingle_ull('iT')
+
+        for popName, popData in self.neuronPopulations.items():
+            self._slm.initNeuronPopIO( popName )
+            self.pullStateFromDevice( popName )
```
neworderofjamie: How come you're **pulling** here? Also for synapse groups etc

lib/swig/GeNNModel.py
braintimeException: import json will be purged.

lib/swig/GeNNModel.py
braintimeException: Ah this is from one of the previous versions. There was an option to provide a mask and rewrite just few of selected variables. In combination with `VarMode::LOC_HOST_DEVICE_INIT_DEVICE` pulling was necessary, but it was complicating the module too much. Now it is unnecessary.

lib/GNUMakefilePyGeNN
neworderofjamie: Another issue with this makefile structure that, when you try and do something else with GeNN and you it can't find your python/numpy source, you can't run genn-build-model.sh!

braintimeException:
Fixed some of the issues, more to come.

braintimeException:
With this fix Python 2.7 seems to work fine, at least on a simple example.

lib/swig/GeNNModel.py
```diff
@@ -45,16 +47,11 @@ def __init__(self, precision=None, modelName='GeNNModel', enableDebug=False,
...
-            self._localhost = 0
-
+        self._localhost = pg.initMPI_pygenn()
         
         pg.setDefaultVarMode( pg.VarMode_LOC_HOST_DEVICE_INIT_DEVICE )
-        pg.cvar.debugCode = enableDebug
-        pg.cvar.autoInitSparseVars = autoInitSparseVars
-        pg.cvar.buildSharedLibrary = True
+        pg.GeNNPreferences.cvar.debugCode = enableDebug
+        pg.GeNNPreferences.cvar.autoInitSparseVars = autoInitSparseVars
```
neworderofjamie: This behaviour is also purely legacy to retain compatibility with old models- just set it to True

lib/swig/GeNNModel.py
braintimeException: After double checking these lines, I believe pulling is still needed, since pull/push affects **all** variables of the population. So, if values need to be pushed only to one variable, without pulling other variables will get whatever garbage was in the memory at the time of allocation.

A check whether **any** variables need to be pushed could avoid unnecessary pulling/pushing.

neworderofjamie: Ahh ok! So firstly you should be doing manual variable initialisation **between** the calls to the auto-generated ``initialize()`` function and the auto-generated ``initializeMODEL_NAME()`` function. That way, assuming that:
1. ``autoInitSparseVars=True`` (as it should be)
2. The variables you are manually initialising are marked as uninitialised (which it looks liek they are)
then everything that needs to be copied will be copied by ``initializeMODEL_NAME()`` function - you shouldn't need to do any manual pushing or pulling

braintimeException: I have checked the `initMODEL_NAME()` function and it only has `initializeSparseArrays()` in it. I tried providing a list of initial voltages and they do not appear in `initializeDevice()` as it should be, so `unintializedVar()` must be working, at least partially...

neworderofjamie: How strange, if you have set ``GENN_PREFERENCES::autoInitSparseVars`` to true there **should** be a call to ``copyStateToDevice(true);`` before ``initializeSparseArrays()`` is called.

braintimeException: OK, it seems I have overlooked that I was setting `autoInitSparseVars=false` elsewhere. Now the `initMODEL_NAME()` has `copyStateToDevice(true)`, `initializeAllSparseArrays()` and `initiazileSparseDevice()` in it. Sorry for misleading you.

However, `copyStateToDevice(true)` has no effect, because when it comes to individual `cudaMemcpy`s they are all wrapped in `if(!hostInitializedOnly) {` and therefore are never executed.

Also it seems that if the call were `copyStateToDevice(false)`, it would overwrite the variables initialized in `initializeDevice()` which in turn is called by `initialize()`.

It feels like I am missing some crucial bit somewhere...

neworderofjamie:
Ahah - that means the var mode for those variables flags is wrong. When
you're initialising any state variables yourself on the host (rather than
passing a VarInit or a constant value through the VarValues you should set
the var mode to ``VarMode::LOC_HOST_DEVICE_INIT_HOST``

lib/swig/GeNNModel.py
braintimeException: yeap, `VarMode::LOC_HOST_DEVICE_INIT_HOST` was the missing bit, thanks.

neworderofjamie:
Fantastic!

lib/swig/GeNNModel.py
```diff
@@ -15,14 +55,14 @@ class GeNNModel( object ):
     """
 
     def __init__(self, precision=None, modelName='GeNNModel', enableDebug=False,
-                 autoInitSparseVars=False, cpuOnly=False):
+                 autoInitSparseVars=True, cpuOnly=False):
```
neworderofjamie: I don't think there's even any point exposing this at all - the behaviour when ``autoInitSpareVars=False`` is going to be removed in the next major version.

braintimeException:
I have also found another bug in current sources related to auto initialization. `pushCURRENT_SOURCE_NAMEToDevice()` does not appear in `copyStateToDevice()`. Shall I fix it here?

neworderofjamie:
Oh shit! That's a good spot - if you could fix it in a separate branch off
master then merge that into your branch that would be excellent

braintimeException:
I get warnings that `addtoinsyn` is never used. Is it because of recent change to the new syntax?

neworderofjamie:
Sadly there are a lot of warnings in the generated code - don't worry about it

tnowotny:
agreed!

PR #203, opened 2018-08-13
Current Sources: fix copyStateToDevice()

braintimeException:
Functions to push states of the current sources in `copyStateToDevice()` were not generated.

### DIFF
--- a/lib/src/generateRunner.cc
+++ b/lib/src/generateRunner.cc
@@ -2463,6 +2463,14 @@ void genRunnerGPU(const NNmodel &model, //!< Model description
             os << "push" << n.first << "SpikesToDevice(hostInitialisedOnly);" << std::endl;
         }
 
+        for(const auto &cs : model.getRemoteCurrentSources()) {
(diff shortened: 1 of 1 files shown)


lib/src/generateRunner.cc
```diff
@@ -2463,6 +2463,14 @@ void genRunnerGPU(const NNmodel &model, //!< Model description
             os << "push" << n.first << "SpikesToDevice(hostInitialisedOnly);" << std::endl;
         }
 
+        for(const auto &cs : model.getRemoteCurrentSources()) {
+            os << "push" << cs.first << "StateToDevice(hostInitialisedOnly);" << std::endl;
+        }
```
neworderofjamie: Remote one shouldn't be pushed as they're sitting on another MPI node (in some theretical world where the MPI support works!)

braintimeException:
Why are remote `NeuronGroup`s are pushed then?

braintimeException:
GitHub has too many buttons :man_facepalming:

neworderofjamie:
It's a little confusing but, if the spikes from remote neuron groups are required on the local machine then the spike variables need to a)be present and b)get pulled

PR #204, opened 2018-08-24
Small memory optimisations

neworderofjamie:
These is a couple of small optimisations I made while benchmarking:

1. As well as the spike IDs you can also cache the row/column lengths when processing pre/postsynaptic events in shared memory to reduce memory bandwidth for synapse kernels
2. Devices with compute capability >= 3.5 have a small, high speed L1 cache for readonly data which can be used to accelerate memory reads that have some locality but aren't coalesced  - here we use this when reading presynaptic variables from the postsynaptic learning kernel - assuming up to 10^4ish neurons being processed by a SM they should fit reasonably well in cache

(Also I tried to remove some of the unused variables)
Also also, in operation future code generator tidy, I think the mechanism for 'wrapping' variables could be swapped in as  more generic alternative to prefixes, suffixes and extensions in ``name_substitution`` etc.


tnowotny:
ok ... sounds pretty good. I am curious about the wrapping functions to decorate variables. how is it generally suppose to be used? Are they just functions you stick a base name into and out come some appropriate prefixes and suffixes with it?

neworderofjamie:
It's not an 100% thought out idea but, for example, there are a lot of places in the code where we have:
```c++
name_substitutions(rCode, "l", nmVars.nameBegin, nmVars.nameEnd, "");
```
to replace variable names with the local copies that have been read into registers. This could instead be expressed as a function into which, as you say,  you stick a base name in and it returns a version with the "l" prefix:
```c++
std::string localNeuronVar(const std::string &varName)
{
  return "l" + varName;
}

name_substitution(rCode, nmVars.nameBegin, nmVars.nameEnd, localNeuronVar);
```

tnowotny:
I see ... I don't feel strongly about it in either direction ...

neworderofjamie:
Writing it down I'm not sure if I think it's an improvement (yey code review) - I'll see what that cache stuff looks like expressed as a prefix and a suffix

tnowotny:
Ok - to argue your case, it attaches a name ```localNeuronVar``` to the process which could make the purpose of the transformation clearer?

PR #205, opened 2018-08-24
Sparse matrix initialisation on GPU

### LINKED ISSUE #31: Provide a helper function to let users use curand libraries to create their own connectivity matrices using CUDA

neworderofjamie:
This is a larger, but hopefully more exciting P.R. that's been W.I.P for quite a while (@esinyavuz - it even fixes #31). Basically it extends the syntax for initialisation variables to initialising sparse synaptic connectivity (bitmask or ragged - as discussed before the yale structure is very difficult to initialise in parallel). Currently it can be used for any connectivity where each row can be generated independently (so, for example, not FixedNumberPre in PyNN-speak) although this could be implemented. You can also augment this with some serial initialisation via the extra global parameters. 

Advantages are:
- Big setup-time performance boost - >200x on Jetson TX2 and >20x on workstation for microcircuit model.
- I think connectivity building is one of the trickiest things for new GeNN users to get their heads around (especially when to get good performance you may need to write multiple versions for different sparse data structures) so, by including a (small) library of connectors we'll be making using GeNN rather a lot simpler.
- I've found that there's a lot of nuance to writing connection-generation code that's performant and doesn't break when numPre*numPost doesn't fit in 32-bits!
- This same syntax is also intended to be used for the future accidentally Izhikevich-inspired resampling connectivity at runtime system

So now all a user needs to do to create a synapse 

```c++
InitSparseConnectivitySnippet::FixedProbability::ParamValues fixedProb(0.1);

model.addSynapsePopulation<WeightUpdateModels::StaticPulse, GeNNModels::ExpCurr>(
        "EE", SynapseMatrixType::RAGGED_GLOBALG, NO_DELAY,
        "E", "E",
        {}, excitatoryStaticSynapseInit,
        excitatoryExpCurrParams, {},
initConnectivity<InitSparseConnectivitySnippet::FixedProbability>(fixedProb));
```

Tests are already added so all that remains is to add some comments to explain how the fixed probability connector works (it's essentially the negative binomial distribution)


tnowotny:
Wow - what can I say - this looks very impressive. Great work @neworderofjamie .
I am particularly chuffed to see how the bitmask clung on to have a revival now ;-)
I don't think I can review the code in more detail here but would hope your tests would have picked up any major breakage. Have you updated/ are you regularly updating the manual?

neworderofjamie:
Thank you! For static synapses with ``GLOBAL_G`` I think it's pretty unbeatable memory-wise.

Yeah I have added a fair number of new tests in this PR (and it's been sitting around for ages being used by me) so I'm **relatively** confident! I think I'll leave the manual until we actually release - that was I can make a much smaller PR with manual updates and maybe switching some examples to using this system

neworderofjamie:
figure_1-2

Looking at it again I'm pretty sure stacked bars don't work with a log scale, but this at least gives an idea of how much this and #163 have improved performance on the microcircuit (300,000,000 synapses with normalised weights)

tnowotny:
stacked bars don't make sense wit log scale indeed ... but you are not the only one to try.

PR #207, opened 2018-08-29
Seperate specification of precision for time

neworderofjamie:
This ended up being pretty simple - basically just adding a different method to ``NNmodel`` to query  the _time_ precision separately. Because of the ``ensureFType`` logic you probably need to be a little careful what you do with your double precision time or it could easily get down-cast but , in STDP rules, I tend to do this:
```c++
```
which I think will work nicely - difference in time is calculated from double precision variables and **then** cast back to single precision (which should easily be able to represent this).  Then subsequent operations on ``dt`` will be done with single-precision, faster maths functions/intrinsics.

### DIFF
--- a/lib/include/modelSpec.h
+++ b/lib/include/modelSpec.h
@@ -71,12 +71,20 @@ enum SynapseGType
 #define CPU 0 //!< Macro attaching the label "CPU" to flag 0
 #define GPU 1 //!< Macro attaching the label "GPU" to flag 1
 
-// Floating point precision to use for models
+//!< Floating point precision to use for models
 enum FloatType
 {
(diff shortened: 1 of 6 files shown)


tnowotny:
Looks pretty good. How did you get all usages of "t" at such short notice?

neworderofjamie:
Searching for " t" was a good start but really there's not that many - Aside from where they're declared/allocated ``t`` and ``sTXXX`` are only referenced a few times and the type is referenced even fewer.

neworderofjamie:
Ancient hardware and people who aren't running long simulations and want to save some memory bandwidth and/or who have GeForces with 1/32 double precision throughput!

PR #211, opened 2018-10-16
Examples manual update

neworderofjamie:
As promised here is the final piece of the GeNN 3.2.0 release


neworderofjamie:
If you're fiddling....it might be better to just exclude the whole ``tests`` directory from Doxygen..

tnowotny:
I don't know ... I just added some file comments to make clear what's part of the test suite.
Do you think people should look into the test suite?

tnowotny:
I will stop mucking about ... the serious business seems all ok as far as I can tell ...

tnowotny:
Once the tests have passed (indicating I didn't totally break all tests by incorrectly adding doxygen comments through a bash script) it should be fine to merge.

neworderofjamie:
Not particularly and, definitely not, via the doxygen docs! there's just a
lot of classes with the same name which are going to be confusing.

PR #213, opened 2018-11-19
Fix for issues with sm_21 devices on windows

neworderofjamie:
The Windows build system requires target architecture to be specified as ``compute_XX,sm_YY``. We were using the value from the device properties for both **but** there is no ``compute_21`` 'virtual architecture' corresponding to SM_21.

Based on the NVCC help this is unique so I have just added a simple hack.
- compute_20, sm_20
- ?, sm_21
- compute_30,sm_30
- compute_32, sm_32
- compute_35, sm_35
- compute_37, sm_37
- compute_50, sm_50
- compute_52, sm_52
- compute_53, sm_53'.

### DIFF
--- a/lib/src/generateRunner.cc
+++ b/lib/src/generateRunner.cc
@@ -2918,11 +2918,15 @@ void genMSBuild(const NNmodel &model,   //!< Model description
 #else
     os << "  <Import Project=\"$(GENN_PATH)\\userproject\\include\\genn.props\"/>" << endl;
     os << endl;
-    const string computeCapability = to_string(deviceProp[theDevice].major) + to_string(deviceProp[theDevice].minor);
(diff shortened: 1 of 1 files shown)


tnowotny:
Sounds reasonable - only worry, what happens when the new versions come online? Have you kept it open-ended for higher versions or do we have to amend each time like for the blocksize optimization code?

tnowotny:
right ... but if no special case it would continue to work "as is", correct?

neworderofjamie:
Yes - that bit definitely works - just waiting on stefan to confirm the
previously broken one

tnowotny:
can this be merged now?

neworderofjamie:
Waiting on Stefan (as the lucky owner of the device in question) to test the fix - just sent him an email reminder.

neworderofjamie:
Just tested on Stefan's laptop and the fix works

PR #214, opened 2018-11-22
First attempt to provide gamma distributed random variables.

neworderofjamie:
@tnowotny - I cherry picked these changes into a new branch as they're not really python related and, this way, I can add tests and merge it seperately


neworderofjamie:
One question is whether we want our gamma to handle scaling for you? Previously I've gone the route of using minimal parameters (i.e. ``$(gennrand_normal)`` takes no parameters) to match curand and scipy style - what do you think?

tnowotny:
I would be fine with removing scaling, i.e. removing ```b```. Even though ```std::gamma_distribution``` seems to include scaling. I am fine either way, I think.

neworderofjamie:
Aside from a bad copy-paste-from-numerical-recipes error - this seems good to go - will just wait to check the new test passes everywhere.

neworderofjamie:
I propose we merge this and the whole Python wrapper into master - it's not doing any harm to 'regular' GeNN users and means we can continue to add features in small P.R.s onto master rather than continuing to add to #202 which has become an un-reviewable nightmare

PR #215, opened 2018-11-23
Exposing CPU_ONLYness to PyGeNN

neworderofjamie:
This means tests etc won't need whether to use CPU or not hardcoded.

- Add a function to the python wrapper which basically returns whether ``CPU_ONLY`` is defined.
-  ``GeNNModel`` now has a ``use_cpu`` property which when set to ``None`` defaults to whether module was built CPU_ONLY or not, otherwise checks validity

### DIFF
--- a/pygenn/genn_model.py
+++ b/pygenn/genn_model.py
@@ -69,15 +69,16 @@ class GeNNModel(object):
     """
 
     def __init__(self, precision=None, model_name="GeNNModel",
-                 enable_debug=False, cpu_only=False):
+                 enable_debug=False, cpu=None):
         """Init GeNNModel
         Keyword args:
         precision       --  string precision as string ("float", "double"
                             or "long double"). defaults to float.
         model_name      --  string name of the model. Defaults to "GeNNModel".
         enable_debug    --  boolean enable debug mode. Disabled by default.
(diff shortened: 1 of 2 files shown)


neworderofjamie:
@tnowotny could you have a quick look at this when you get a minute?

pygenn/genn_model.py
```diff
@@ -69,15 +69,16 @@ class GeNNModel(object):
     """
 
     def __init__(self, precision=None, model_name="GeNNModel",
-                 enable_debug=False, cpu_only=False):
+                 enable_debug=False, cpu=None):
```
tnowotny: Is it not "cpu_only= None"?

tnowotny: Oh - I see ... just also renamed it everywhere. I don't mind. We called it "cpu_only" in the past because one can of course also use the cpu in the GPU/CPU version.

neworderofjamie: This was intentional - ``CPU_ONLY`` is returned by the wrapper to signify that there's no GPU support, here you're saying whether you want to use the CPU or not (if that makes sense)

tnowotny: I see now - very sensible ... you could of course choose to use the CPU even if GPU support is available.

PR #217, opened 2018-12-03
Issues with seeding of device RNG

neworderofjamie:
If the ``NNmodel`` seed was manually set, this was correctly used to initialize the CUDA RNGs. However, if it was set to zero, the seed of the CUDA RNGs weren't randomly initialised. In this change I generate ``sizeof(unsigned long long)`` worth of system randomness to seed it in this case.

Also I spotted what I think is a potential bug in the ``initializeDevice`` kernel - it was entirely possible that, while ``dd_rng`` was initialized at the top of the kernel, the updated state didn't make it to all the states. Am I right in this interpretation of CUDA  @tnowotny? I don't think even a ``__threadfence`` would help here so I just added a pre-initialisation kernel to initialise the RNG.

### DIFF
--- a/lib/src/generateInit.cc
+++ b/lib/src/generateInit.cc
@@ -233,14 +233,10 @@ unsigned int genInitializeDeviceKernel(CodeStream &os, const NNmodel &model, int
 {
     // init kernel header
     os << "extern \"C\" __global__ void initializeDevice(";
-    const auto &params = model.getInitKernelParameters();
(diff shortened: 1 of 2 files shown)


tnowotny:
Oh boy ... that if (id==0) construct would never have worked. No synchronization  between blocks unless they introduced something brand new I am not aware of.

The other issue also seems solved appropriately - pull request approved.

PR #218, opened 2018-12-04
Better memory calculation

neworderofjamie:
Not sure why I never merged this before, but I resurrected this branch which fixes a number of issues with memory calculation:
1. It can calculate the size of various cuRAND structures
2. It calculates ragged matrix sizes
3. Memory in bytes could easily overflow 32-bit so it switches everything to ``size_t``


tnowotny:
sounds perfectly reasonable.

PR #233, opened 2019-04-11
Fix for undefined get_code variable used in create_custom_init_var_snippet_class

ajaysub110:
Hello,

I have replaced `if get_code is not None` which checks if a 'get_code' key is already present in the `body` dictionary with an equivalent `if 'get_code' in body.keys()` since get_code was an undefined variable and hence gave a NameError while running the create_custom_init_var_snippet_class() method.

Thanks,
Ajay


pygenn/genn_model.py
```diff
@@ -969,7 +969,7 @@ def create_custom_init_var_snippet_class(class_name, param_names=None,
 
     body = {}
 
-    if get_code is not None:
```
neworderofjamie: All the problem was is that ``get_code`` here should have been ``var_init_code`` - the parameter that got passed in!

neworderofjamie:
Can you revert this and remove all the automatically generated files (none of the genn_wrapper should be added)

neworderofjamie:
You got it! Nice work

PR #234, opened 2019-04-12
GeNN 4.0.0

### LINKED ISSUE #68: Use PCIBusId instead of CUDA device number
When debugging SM_2.x devices, the device which operates the monitor seems to be no longer present in the cuda device ID list. See below:

http://stackoverflow.com/questions/36160799/do-cuda-device-ids-change-when-debugging/36167025#36167025

If, for instance, the device operating the display is device 0 during code generation, and GeNN decides that device 0 is the optimum device to generate code on, then when the generated code is debugged, device 0 - which cannot be debugged on and operate the display simultaneously - is not available, and device 1 becomes device 0. So executing and debugging the same executable may not run it on the same device architecture.

Suggested solution is to get and set device by PCIe bus ID only:

cudaDeviceGetByPCIBusId
and
cudaDeviceGetPCIBusId

### LINKED ISSUE #149: Support for -std=c++14 and later

### LINKED ISSUE #225: No proper install
The installation instructions do not really describe an 'installation' in the usual sense of the word, merely an unpacking into 'a convenient location', but then goes on to talk about the software being 'installed' into /usr/local/GeNN. I tried reading between the lines and installing it as root under /usr/local, which of course assumes the ability to install as root, but running the installation test as a normal user then fails as it tries to write into the directory it was installed in. It became clear that GeNN is not actually designed to be installed in the conventional sense, but to be used is situ wherever it is unpacked, and that that will need to be done separately by each user on a machine.

At the very least the so-called installation instructions at http://genn-team.github.io/genn/documentation/3/html/Installation.html are misleadingly named, and contain misleading advice.

The better fix would be to create a proper install target in the makefile such that the software can be installed using the traditional `make; sudo make install` steps, together with fixing the test framework so that it does not necessarily try to put its output where the source is.

neworderofjamie:
So this isn't really a pull request as I'll probably end up overwriting master with this rather than attempting to merge as the changes are too many! However this seems a good forum for obtaining feedback. Things that would be particularly good to test are:
- On *nix does ``make install`` work correctly and figure out which backends your machine supports?
- On Windows does it work with your CUDA/MSVC install?
- How annoying is updating your models?
- (Assuming a suitable one is available) do the python wheels linked from the bottom of this post work for you and let you run the example in the documentation? (install using ``pip install NAME_OF_WHEEL``)

In terms of the refactor, I made some [slides](https://docs.google.com/presentation/d/1UYLlMHWG8e6PKznxrmzY7h3V5WwM85nICiIHvckjp1Q/edit?usp=sharing ) about the general ideas for @tnowotny and @alexdewar which may also be helpful. Finally the updated manually is not currently linked to the github.io page but is online [here](https://genn-team.github.io/genn/documentation/4/html/index.html) and the release notes are as follows:

Release Notes for GeNN v4.0.0
====
This release is the result of a second round of fairly major refactoring which we hope will make GeNN easier to use and allow it to be extended more easily in future.
However, especially if you have been using GeNN 2.XX syntax, it breaks backward compatibility.

User Side Changes
----
1. Totally new build system - ``make install`` can be used to install GeNN to a system location on Linux and Mac (fixes #225) and Windows projects work much better in the Visual Studio IDE.
2. Python interface now supports Windows and can be installed using binary 'wheels'.
3. No need to call ``initGeNN()`` at start and ``model.finalize()`` at end of all models.
4. Initialisation system simplified - if you specify a value or initialiser for a variable or sparse connectivity, 
   it will be initialised by your chosen backend. If you mark it as uninitialised, it is up to you to initialize it 
   in user code between the calls to ``initialize()`` and ``initializeSparse()`` (where it will be copied to device).
5. ``genn-create-user-project`` helper scripts to create Makefiles or MSBuild projects for building user code
6. State variables can now be pushed and pulled individually using the `pull<var name><neuron or synapse name>FromDevice()` and `push<var name><neuron or synapse name>ToDevice()` functions.
7. Management of extra global parameter arrays has been somewhat automated.
8. ``GENN_PREFERENCES`` is no longer a namespace - it's a global struct so members need to be accessed with . rather than ::.
9. ``NeuronGroup``, ``SynapseGroup``, ``CurrentSource`` and ``NNmodel`` all previously exposed a lot of methods that the user wasn't <i>supposed</i> to call but could. 
   These have now all been made protected and are exposed to GeNN internals using derived classes (``NeuronGroupInternal``, ``SynapseGroupInternal``, ``CurrentSourceInternal``, ``ModelSpecInternal``) that make them public using ``using`` directives.
10. Auto-refractory behaviour was controlled using ``GENN_PREFERENCES::autoRefractory``, this is now controlled on a per-neuron-model basis using the ``SET_NEEDS_AUTO_REFRACTORY`` macro.
11. The functions used for pushing and pulling have been unified somewhat this means that ``copyStateToDevice`` and ``copyStateFromDevice`` functions no longer copy spikes and ``pus<neuron or synapse name>SpikesToDevice`` and ``pull<neuron or synapse name>SpikesFromDevice`` no longer copy spike times or spike-like events.
12. Standard models of leaky-integrate-and-fire neuron (``NeuronModels::LIF``) and of exponentially shaped postsynaptic current (``PostsynapticModels::ExpCurr``) have been added.
13. When a model is built using the CUDA backend, the device it was built for is stored using it's PCI bus ID so it will always use the same device - fixes #68 
14. -s option for buildmodel.sh allows C++ standard used to build model to be set - fixes #149

Deprecations
----
1. Yale-format sparse matrices are no longer supported.
2. GeNN 2.X syntax for implementing neuron and synapse models is no longer supported.
3. ``$(addtoinSyn) = X; $(updatelinsyn);`` idiom in weight update models has been replaced by function style ``$(addToInSyn, X);``.

Binary wheels
====
- https://gen-ci.inf.sussex.ac.uk/job/GeNN/job/genn/job/genn_4/197/artifact/cuda10-pygenn-0.2-cp27-cp27mu-linux_x86_64.whl
- https://gen-ci.inf.sussex.ac.uk/job/GeNN/job/genn/job/genn_4/197/artifact/cuda10-pygenn-0.2-cp37-cp37m-linux_x86_64.whl
- https://gen-ci.inf.sussex.ac.uk/job/GeNN/job/genn/job/genn_4/197/artifact/cuda10-pygenn-0.2-cp37-cp37m-win_amd64.whl
- https://gen-ci.inf.sussex.ac.uk/job/GeNN/job/genn/job/genn_4/197/artifact/cuda9-pygenn-0.2-cp27-cp27m-macosx_10_13_intel.whl


tnowotny:
This is a major piece of work and I like the design. It's hard to review a large change like this but I had a peek into the CUDA backend code and it seemed like a reasonable division of labour between the different generator functions. The automatic tests are not passing - do you know why?

neworderofjamie:
Indeed - it's always a little hard to know what to do with something that's not usefully broken up into reviewable chunks - glad you like the general idea though.

The automatic tests is an interesting question - there is something about the configuration of 'master' - the CentOS machine that's running Jenkins that causes test failures for exactly 3 tests. I think it's a standard library rather than a compiler issue but, beyond that, I have failed to figure it out.

neworderofjamie:
Finally figured that out - after much inspection of library versions turned out to be just a bug in that group of tests

PR #235, opened 2019-04-13
Host-initialised variable not copied to device in Izh_sparse

jamesturner246:
A very minor edit to Izh_sparse. The inhibitory U variable is an `uninitialisedVar()`, initialised on host, but is not copied to device since it is not set to `varMode::LOC_HOST_DEVICE_INIT_HOST`. This is causing undefined behaviour.

Is this a more general problem with uninitialised vars, where a user may be unaware that their host-initialised vars are not copied over when copy-to-device is called? It certainly tripped me up for a while.


neworderofjamie:
Good spot and totally agree - this whole system is somewhat simplified in GENN 4(#234)

PR #237, opened 2019-04-30
Mention AUR packages in readme file

alexdewar:
I've made a couple of GeNN packages on the Arch User Repo, mostly for me (they're source packages which look like this: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=genn). Thought it would be worth a mention even though the intersection of Arch users and GeNN users is probably small for now (n = 1)!

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,19 @@ restarted to initialise the variables set using the `SETX` command.
 
 This completes the installation.
 
+### ARCH LINUX INSTALL
+For Arch Linux there are GeNN packages available [from the AUR](https://aur.archlinux.org/packages/?O=0&K=genn).
+
+Install with your AUR helper of choice, like so:
+```sh
+yay -S genn
+```
+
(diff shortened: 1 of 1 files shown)


alexdewar:
Sorry, merged by accident! Feel free to unmerge if you don't like it

neworderofjamie:
No that's cool - how difficult is the updating process for an AUR release?

alexdewar:
Really easy as long as the build process is the same. You just update the value of ``pkgver`` and regenerate the file hashes.

alexdewar:
...I mean that's what I have to do as the maintainer when there's a new version. Users can just update it like an ordinary package with ``yay`` or whatever

alexdewar:
Also.... FYI while I was playing with things I found that both ``run_tests.sh`` and trying to build pygenn with ``setup.py`` didn't work for me, which is weird because I'm on my desktop machine and clearly both of these things work through Jenkins! The tests failed saying it couldn't find a ``.gcno`` file and I got a compiler error when trying to build pygenn.

PR #242, opened 2019-05-24
SpineML simulator library

neworderofjamie:
The SpineML simulator was previous a single monolithic executable. This P.R. encapsulates the simulator in a class ``SpineMLSimulator::Simulator`` which lets you integrate SpineML simulations with your own applications e.g. for robotics. The general usage of this class looks something like this (see https://github.com/BrainsOnBoard/bob_robotics/pull/118 for a more complete example):

```c++
// Create SpineML simulator object
SpineMLSimulator::Simulator simulator("experiment1.xml", ".");

// Get logger
const auto *steerLeft = simulator.getExternalLogger("steer left");

// Get input
auto *sensorLeft = simulator.getExternalInput("sensor left");

while(true) {
    float left = XXXXX;
    std::fill(sensorLeft->getBufferBegin(), sensorLeft->getBufferEnd(), left);
   
    // Advance SpineML simulation
    simulator.stepTime();

   // Print first output
   std::cout << *steerLeft->getStateVarBegin() << std::endl
}
```
Any thoughts would be much appreciated! Because there's no new SpineML logger or input types to represent this new sort of external IO, you specify them in SpineML by setting ``hostname`` to ``0.0.0.0``.


src/spineml/simulator/inputValue.cc
```diff
@@ -323,11 +303,39 @@ void SpineMLSimulator::InputValue::External::update(double, unsigned int,
...
+//----------------------------------------------------------------------------
+// SpineMLSimulator::InputValue::ExternalNetwork
+//----------------------------------------------------------------------------
+SpineMLSimulator::InputValue::ExternalNetwork::ExternalNetwork(double dt, unsigned int numNeurons, const pugi::xml_node &node)
+: External(dt, numNeurons, node)
+{
+    // Read connection stats
+    const std::string connectionName = node.attribute("name").value();
+    const std::string hostname = node.attribute("host").value();
+    const int port = node.attribute("tcp_port").as_int();
```
alexdewar: Out of interest, does pugiXML throw an exception if these attributes are not present? Or should we be testing for this?

src/spineml/simulator/inputValue.cc
```diff
@@ -342,7 +350,21 @@ std::unique_ptr<SpineMLSimulator::InputValue::Base> SpineMLSimulator::InputValue
         return std::unique_ptr<Base>(new TimeVaryingArray(dt, numNeurons, node));
```
alexdewar: I'd say use ``make_unique`` instead, but GeNN still supports C++11, right?

src/spineml/simulator/inputValue.cc
```diff
@@ -342,7 +350,21 @@ std::unique_ptr<SpineMLSimulator::InputValue::Base> SpineMLSimulator::InputValue
...
+        if(hostName == "0.0.0.0") {
+            External *inputValue = new External(dt, numNeurons, node);
+
+            // Add to map of external inputs
+            const std::string name = node.attribute("name").value();
+            if(!externalInputs.emplace(name, inputValue).second) {
+                LOGW << "External input with duplicate name '" << name << "' encountered";
+            }
+
+            return std::unique_ptr<Base>(inputValue);
```
alexdewar: Technically one could get a memory leak here if an exception is thrown in this block. If you make ``inputValue`` a ``unique_ptr`` from the outset you can avoid this and use ``get()`` to get the raw pointer out

src/spineml/simulator/logOutput.cc
```diff
@@ -274,16 +233,73 @@ void SpineMLSimulator::LogOutput::AnalogueNetwork::record(double, unsigned int t
...
+// SpineMLSimulator::LogOutput::AnalogueNetwork
+//----------------------------------------------------------------------------
+SpineMLSimulator::LogOutput::AnalogueNetwork::AnalogueNetwork(const pugi::xml_node &node, double dt,
+                                                              const std::string &port, unsigned int popSize,
+                                                              const filesystem::path &logPath,
+                                                              const ModelProperty::Base *modelProperty)
+    : AnalogueExternal(node, dt, port, popSize, logPath, modelProperty)
+{
+    // Check size determined by indices/population size matches attribute
+    const unsigned int size = getIndices().empty() ? popSize : (unsigned int)getIndices().size();
```
alexdewar: Probably better to use a ``size_t`` here. MSVC and gcc use different types for ``size_t``

alexdewar: ...or ``auto``

src/spineml/simulator/logOutput.cc
```diff
@@ -274,16 +233,73 @@ void SpineMLSimulator::LogOutput::AnalogueNetwork::record(double, unsigned int t
...
+    //assert(size == node.attribute("size").as_uint());
+
+    // Allocate output buffer
+    m_OutputBuffer.resize(size);
+
+    // Read connection stats
+    const std::string connectionName = node.attribute("name").value();
+    const std::string hostname = node.attribute("host").value();
+    const int tcpPort = node.attribute("tcp_port").as_int();
```
alexdewar: I believe port numbers are always ``uint16_t``. You might get a warning about a narrowing conversion, depending on how fussy your compiler is

src/spineml/simulator/logOutput.cc
```diff
@@ -274,16 +233,73 @@ void SpineMLSimulator::LogOutput::AnalogueNetwork::record(double, unsigned int t
...
+void SpineMLSimulator::LogOutput::AnalogueNetwork::recordInternal()
+{
+    // If no indices are specified, transform all values in model property into double precision
+    // **TODO** once precision is switchable this could be optimised out
+    if(getIndices().empty()) {
+        std::transform(getModelPropertyHostStateVarBegin(), getModelPropertyHostStateVarEnd(), m_OutputBuffer.begin(),
+                    [](scalar x)
+                    {
+                        return static_cast<double>(x);
+                    });
```
alexdewar: Wouldn't ``std::copy`` do here? I *think* it casts values to the output type

src/spineml/simulator/simulator.cc
```diff
@@ -0,0 +1,820 @@
...
+    // Attempt to load model library
+#ifdef _WIN32
+    auto libraryPath = outputPath / "run" / "runner_Release.dll";
+    LOGI << "Experiment using model library:" << libraryPath;
+    m_ModelLibrary = LoadLibrary(libraryPath.str().c_str());
+#else
+    auto libraryPath = outputPath / "run" / (networkName + "_CODE") / "librunner.so";
+    LOGI << "Experiment using model library:" << libraryPath;
+    m_ModelLibrary = dlopen(libraryPath.str().c_str(), RTLD_NOW);
+#endif
```
alexdewar: I'd personally be inclined to wrap the library loading etc. into its own class and header file to reduce the if-deffery in this file, but it's up to you

src/spineml/simulator/simulator.cc
```diff
@@ -0,0 +1,820 @@
...
+    m_DT = eulerIntegration.attribute("dt").as_double(0.1);
+    LOGI << "DT = " << m_DT << "ms";
+
+    // Read duration from simulation and convert to timesteps
+    m_DurationMs = simulation.attribute("duration").as_double() * 1000.0;
+
+
+    // Create directory for logs (if required)
+    const auto logPath = outputPath / "log";
+    filesystem::create_directory(logPath);
```
alexdewar: I believe this function returns false if it fails. You might want to test for this.

src/spineml/simulator/simulator.cc
```diff
@@ -0,0 +1,820 @@
...
+{
+    const unsigned long long numTimeSteps = calcNumTimesteps();
+    // Get name of target
+    std::string target = node.attribute("target").value();
+
+    // Find size of target component
+    auto targetSize = componentSizes.find(target);
+    if(targetSize == componentSizes.end()) {
+        throw std::runtime_error("Cannot find component '" + target + "'");
+    }
```
alexdewar: Maybe this var should be called ``targetSizeElem`` or something? It doesn't actually represent the target size

src/spineml/simulator/simulator.cc
```diff
@@ -0,0 +1,820 @@
...
+            // If there is a model property object for this port return an analogue log output to read it
+            auto portProperty = targetProperties->second.find(port);
+            if(portProperty != targetProperties->second.end()) {
+                if(shouldLogToFile) {
+                    return std::unique_ptr<LogOutput::Base>(new LogOutput::AnalogueFile(node, getDT(), numTimeSteps, port, targetSize->second,
+                                                                                        logPath, portProperty->second.get()));
+                }
+                else if(hostName == "0.0.0.0") {
+                    // Create logger
+                    LogOutput::AnalogueExternal *log = new LogOutput::AnalogueExternal(node, getDT(), port, targetSize->second,
```
alexdewar: As before, I'd declare the ``unique_ptr`` here to prevent a possible memory leak

src/spineml/standalone_simulator/main.cc
```diff
@@ -0,0 +1,55 @@
...
+int main(int argc, char *argv[])
+{
+    try
+    {
+        if(argc < 2) {
+            throw std::runtime_error("Expected experiment XML file passed as arguments");
+        }
+
+        // Read min log severity from command line
+        const plog::Severity minSeverity = (argc > 3) ? (plog::Severity)std::stoi(argv[3]) : plog::info;
```
alexdewar: Should this be ``argc > 2`` and ``argv[2]``? Or are you using the second command line argument for something else?

src/spineml/standalone_simulator/main.cc
```diff
@@ -0,0 +1,55 @@
...
+        plog::init(minSeverity, &consoleAppender);
+
+#ifdef _WIN32
+        // Startup WinSock 2
+        WSADATA wsaData;
+        if(WSAStartup(MAKEWORD(2, 2), &wsaData) != 0) {
+            throw std::runtime_error("WSAStartup failed");
+        }
+
+#endif  // _WIN32
```
alexdewar: I think you also need to call ``WSACleanUp`` (or whatever it's called) before the program exits, or your socket can end up locked

include/spineml/simulator/input.h
```diff
@@ -40,28 +40,28 @@ class Base
     //----------------------------------------------------------------------------
     // Declared virtuals
     //----------------------------------------------------------------------------
-    virtual void apply(double dt, unsigned int timestep) = 0;
+    virtual void apply(double dt, unsigned long long timestep) = 0;
```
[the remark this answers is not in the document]
neworderofjamie: I **think** it's standard from C++11 onwards. However, the reason for this change was that it's the type that GeNN uses internally for timesteps and previously we were truncating so, if a compiler doesn't support it then GeNN won't work either :fearful:

include/spineml/simulator/inputValue.h
```diff
@@ -166,14 +167,35 @@ class External : public Base
...
-    virtual void update(double dt, unsigned int timestep,
-                        std::function<void(unsigned int, double)> applyValueFunc) override;
+    virtual void update(double dt, unsigned long long timestep,
+                        std::function<void(unsigned int, double)> applyValueFunc) final;
+
+    //------------------------------------------------------------------------
+    // Public API
+    //------------------------------------------------------------------------
+    std::vector<double>::iterator getBufferBegin(){ return m_Buffer.begin(); }
+    std::vector<double>::iterator getBufferEnd(){ return m_Buffer.end(); }
```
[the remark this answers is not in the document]
neworderofjamie: Well this is an input so you can't do much to it with a const pointer. However I didn't want to expose a non-const reference to ``m_Buffer`` as that would let you fuck it up in many ways that would break the internals.

src/spineml/simulator/inputValue.cc
neworderofjamie: It returns empty strings (https://pugixml.org/docs/manual.html#access.attrdata) so the failure mode should be that the network client barfs

src/spineml/simulator/inputValue.cc
neworderofjamie: Yeah, GeNN does indeed use C++11. For whatever reason Visual Studio 2013 is the minimum targeted compiler (it was because it was installed on university managed Windows machines but, now it's more that my windows machine at home has that...) and that barely does C++11.

src/spineml/standalone_simulator/main.cc
neworderofjamie: Yeah - the 2nd command line argument is working directory (the default used for SpineCreator is very irritating)

src/spineml/standalone_simulator/main.cc
neworderofjamie: :+1:

include/spineml/simulator/input.h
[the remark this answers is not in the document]
alexdewar: Ok, fair enough!

include/spineml/simulator/inputValue.h
```diff
@@ -4,6 +4,7 @@
 #include <functional>
```
alexdewar: This file's name should probably be in snake-case because that's what we've used elsewhere

include/spineml/simulator/inputValue.h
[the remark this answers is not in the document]
alexdewar: That's a good point.

src/spineml/simulator/inputValue.cc
alexdewar: Ah the good ol' ``std::barf_exception``!

src/spineml/simulator/inputValue.cc
alexdewar: Ok cool. There is always the option of copying and pasting a ``make_unique`` function from the internet -- it is handy (and safer in certain situations)

include/spineml/simulator/inputValue.h
neworderofjamie: Slightly confusingly GeNN uses lower-camel case for filenames

src/spineml/simulator/inputValue.cc
neworderofjamie: Very good point - fixed!

include/spineml/simulator/inputValue.h
alexdewar: Haha I forgot this was a PR on the GeNN repo :-p

src/spineml/simulator/logOutput.cc
neworderofjamie: I guess, for the sake of fussy compilers, we could add a cast here https://github.com/genn-team/genn/blob/spineml_simulator_library/src/spineml/simulator/networkClient.cc#L60?

src/spineml/simulator/logOutput.cc
neworderofjamie: Yeah - population sizes **should** definitely be ``size_t``. The problem is that a lot of GeNN and hence SpineML  uses ``unsigned int`` for population size and I'm not sure if it's worth fixing this particular bit in this P.R.

src/spineml/simulator/logOutput.cc
alexdewar: Yeah, that would work, although port numbers should always be ``uint16_t``, so you could just use that from the outset.

src/spineml/simulator/simulator.cc
neworderofjamie: it doesn't distinguish between the case where the output directory exists (normal) and where it can't create it due to lack of disk space or something though so it's not particularly meaningful

src/spineml/simulator/simulator.cc
neworderofjamie: Done!

src/spineml/simulator/simulator.cc
neworderofjamie: Yeah - that's a future project. There's very similar logic here, in PyGeNN and in BoB robotics now

src/spineml/simulator/logOutput.cc
neworderofjamie: hmmm that would be cleaner-looking but it's very unclear to me from the documentation whether it's the case. It suggests that for trivial types, it'll do block copies using memmove but nothing about different types...

include/spineml/simulator/simulator.h
```diff
@@ -0,0 +1,191 @@
...
+#include "modelProperty.h"
+
+//----------------------------------------------------------------------------
+// Macros
+//----------------------------------------------------------------------------
+#ifdef _WIN32
+    #define LIBRARY_HANDLE HMODULE
+#else
+    #define LIBRARY_HANDLE void*
+#endif
```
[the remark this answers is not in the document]
neworderofjamie: Done

src/spineml/simulator/logOutput.cc
alexdewar: Ok, fair enough. It's probably not a bad thing to force population sizes to be ``unsigned int`` on Windows and *nix, so that they're 32 bits on all platforms.

src/spineml/simulator/simulator.cc
alexdewar: You could do a check to see if the directory exists beforehand, but it's up to you. If the folder isn't creating I assume you'd just get an error message further down the line anyway.

src/spineml/simulator/logOutput.cc
alexdewar: I'll do a ``memmove`` if e.g. you're copying ``int``s from one array to another, but I think otherwise it'll just use the assignment operator, so you'll get an implicit cast. It's true that the documentation doesn't seem to mention this explicitly though.

src/spineml/simulator/logOutput.cc
neworderofjamie: So, in general I favour passing native-sized types around, but I have changed it to parse and subsequently pass the port around as ``unsigned int``  and cast to ``uint16_t`` at the ``htons`` call to hopefully appease fussy compilers

alexdewar: :+1:

PR #243, opened 2019-05-28
PyGeNN fix

neworderofjamie:
Clearly a GeNN 4.0.1 is going to be required! This includes:
- Fixes a dumb typo
- Adds some more error-checking to parameter and value initialisation. Previously, while PyGeNN converted parameters and values from Python dictionaries to C++ GeNN arrays, it didn't check if your dictionary didn't also have any extra unsupported parameters which made it very easy to make mistakes
- Previously you got a auto-generated SWIG error if you tried setting parameters to e.g. arrays - this adds a friendlier error earlier

### DIFF
--- a/pygenn/genn_model.py
+++ b/pygenn/genn_model.py
@@ -521,7 +521,7 @@ class GeNNModel(object):
         if not self._loaded:
             raise Exception("GeNN model has to be loaded before pulling")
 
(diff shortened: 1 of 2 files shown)


neworderofjamie:
@tnowotny  could you have a quick look at this? it would be nice to get some of these small things merged and maybe make a bugfix release before the BoB workshop

PR #244, opened 2019-05-30
Small fix for Visual Studio 2013

neworderofjamie:
Clearly I hadn't worked from home using this machine for a while as GeNN 4.0.0 doesn't build on Visual Studio 2013 (our lowest support Visual Studi) as it has slightly broken support for {} initialisation - basically it requires an extra set of curly brackets i.e. {{}}.


neworderofjamie:
Yeah we've done it in several other places for the same reason and all the tests pass.

PR #245, opened 2019-07-16
Synapse update strategy

neworderofjamie:
This P.R. provides a framework to start experimenting with new presynaptic update algorithms - it employs the [strategy design pattern](https://en.wikipedia.org/wiki/Strategy_pattern) to allow each approach to be implemented within its own class. For now it doesn't change anything, this is merely a refactor.


neworderofjamie:
No indeed - aside from the trivialest copy, paste and tweak you do require a relatively deep knowledge of what's going on.

PR #246, opened 2019-07-16
CUDA error checking improvement

neworderofjamie:
I was getting a mysterious lack of spikes when running a model and it turned out there were out of memory errors going uncaught. This P.R. wraps all ``cudaXXX`` functions in ``CHECK_CUDA_ERRORS`` and additionally checks errors after kernel launches using ``cudaPeekAtLastError()``


neworderofjamie:
I would imagine that the overhead is minimal - all CUDA host functions are returning the error code anyway and I'd guess peek is literally returning the CUDA equivalent of ``errno``

Issue #55, opened 2015-08-19
Refine global - register -global transfers


tnowotny:
within the learn code, there can be repeated reference to dd_remapXX[iprePos]. Whenever it is used more than once, it would probably be efficient to fetch it into a local register.

tnowotny:
1. load dd_remapXX[...] into a register by default.
2. But more generally, introduce an annotation "readonly XXX YYY" e.g. ``readonly scalar x;`` for variables. Those would not be written back to global memory; the register copy could be declared ``const``

tapaswenipathak:
Hi folks: Is this ticket still open?

neworderofjamie:
Well, the read-only part has been implemented by #247 and the equivalent of ``dd_remap[iprepos]`` is now loaded into a register [here](https://github.com/genn-team/genn/blob/master/src/genn/backends/cuda/backend.cc#L696). The issue is that that is done whether it is required or not as discussed in #248

tapaswenipathak:
All PRs are there then right, can you ref the ticket in the PRs?

PR #247, opened 2019-07-21
Read only vars

neworderofjamie:
This change allows you to specify whether state variables are readonly when you define a model e.g. in the ``IzhikevichVariable (using syntax similar to that suggested in  #55):
```c++
class IzhikevichVariable : public Izhikevich
{
public:
    DECLARE_MODEL(NeuronModels::IzhikevichVariable, 0, 6);

    SET_PARAM_NAMES({});
    SET_VARS({{"V","scalar"}, {"U", "scalar"}, 
              {"a", "scalar", VarAccess::READ_ONLY}, {"b", "scalar", VarAccess::READ_ONLY}, 
              {"c", "scalar", VarAccess::READ_ONLY}, {"d", "scalar", VarAccess::READ_ONLY}});
};
```
Knowing whether state variables are read-only is useful in two ways:
- Immediately: to remove unnecessary writes of neuron and postsynaptic; and pre and postsynaptic post weight update model state variable back to global memory (to some extent aiding #55) - depending on dumbness of CUDA compiler this could reduce memory bandwidth and reduce register pressure
- In future: this allows all kinds of useful 'analysis' e.g. to determine if state variables can be optimised out either in the style of ``GLOBALG`` 
 (this is something that I think is done in the PyNN, SpineML and, I suspect, Brian frontends) or by the long-dreamed-of Izhikevich style re-sampling

### DIFF
--- a/include/genn/genn/snippet.h
+++ b/include/genn/genn/snippet.h
@@ -119,16 +119,14 @@ public:
     //----------------------------------------------------------------------------
     // Structs
     //----------------------------------------------------------------------------
-    //! A variable has a name, a type and a readonly flag
-    /*! Through the wonders of C++ aggregate initialization, if readonly 
-    // is not initialized in the {}, it will be VALUE-initialised to false*/
-    struct Var
+    //! An extra global parameter has a name and a type
+    struct EGP
     {
         std::string name;
         std::string type;
-        bool readonly;
     };
 
+    //! Additional input variables, row state variables and other things have a name, a type and an initial value
     struct ParamVal
     {
         std::string name;
@@ -148,7 +146,7 @@ public:
     // Typedefines
     //----------------------------------------------------------------------------
     typedef std::vector<std::string> StringVec;
-    typedef std::vector<Var> VarVec;
+    typedef std::vector<EGP> EGPVec;
     typedef std::vector<ParamVal> ParamValVec;
     typedef std::vector<DerivedParam> DerivedParamVec;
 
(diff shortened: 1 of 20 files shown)


include/genn/genn/snippet.h
```diff
@@ -119,11 +119,14 @@ class GENN_EXPORT Base
     //----------------------------------------------------------------------------
     // Structs
     //----------------------------------------------------------------------------
-    //! A variable has a name and a type
+    //! A variable has a name, a type and a readonly flag
+    /*! Through the wonders of C++ aggregate initialization, if readonly 
+    // is not initialized in the {}, it will be VALUE-initialised to false*/
```
tnowotny: When I saw the syntax user-side, I was wondering whether a flag like "ro" vs "rw" in the initializer would have been more understandable ... but I can see how it looks very convenient from this side.
What do you think - any way to make this more explicit (i.e. that "true" in third position signifies readonly)?

neworderofjamie: Yeah, in #55 the suggested syntax was adding "ro" or "rw" to the start of the type but, this is indeed much more convenient :) You're right that it's not very explicit though - maybe an enum might be better?

tnowotny: An enum might work ...

tnowotny: ... and obviously I had forgotten #55 entirely.

neworderofjamie: Hah me too

neworderofjamie: The enum syntax is a little less terse but way clearer

tnowotny: Ok, yes, this is much clearer. I think it's good to go.

Issue #47, opened 2015-01-29
Avoiding creating unused variables


esinyavuz:
I'm renaming this thread as we have other instances of unused variable type of problems and it would be nice to take care of this in a more systematic way during code generation.

The previous thread consisted of the warnings on:
Unused variable warning for ipost, ipre for GLOBALG used with NSYNAPSE model

This happens in the SynDelay project. It is not as straightforward as it seems; indexing is not necessary when we use GLOBAL, but if we had a synapse model with a variable that uses these indices they would be needed. These are just warnings at the moment, but it would be more elegant to make a proper check. 

Another example of this kind of problem is unnecessary post-to-pre arrays for sparse connectivity. They are (presumably) not needed unless the learning kernel is used.

Creating unnecessary variables would fill the memory unnecessarily and we should avoid this.

tapaswenipathak:
Hi @esinyavuz: Can I work on the ticket?

neworderofjamie:
Hi @tapaswenipathak,

I'm afraid @esinyavuz has left the project, but you're more than welcome to work on this ticket. GeNN's code generator has been somewhat refactored since this issue was created so this is less of an issue but, looking through our build server's output, the following two unused variables are definitely still being created by GeNN's CUDA backend:

1. ``shSpk`` created [here](https://github.com/genn-team/genn/blob/master/src/genn/backends/cuda/backend.cc#L523) - as well as the existing tests, this should only be created if any of the synapse groups have their ``m_SpanType`` set to ``SynapseGroup::SpanType::POSTSYNAPTIC``
2. ``synAddress`` created [here](https://github.com/genn-team/genn/blob/master/src/genn/backends/cuda/presynapticUpdateStrategy.cc#L281) - this is rather more tricky as it depends upon whether ``$(id_syn)`` is referenced in the subsequently generated code.

tapaswenipathak:
Thanks for the notes @neworderofjamie.

tapaswenipathak:
@neworderofjamie: No `synAddress` can be removed as per me, can you take a look?

PR #248, opened 2019-07-27
Avoiding creating unused variables 

tapaswenipathak:
Close https://github.com/genn-team/genn/issues/47.
Ref https://github.com/genn-team/genn/issues/55.

### DIFF
--- a/src/genn/backends/cuda/backend.cc
+++ b/src/genn/backends/cuda/backend.cc
@@ -277,7 +277,6 @@ void Backend::genNeuronUpdate(CodeStream &os, const ModelSpecInternal &model, Ne
         if(std::any_of(model.getLocalNeuronGroups().cbegin(), model.getLocalNeuronGroups().cend(),
             [](const ModelSpec::NeuronGroupValueType &n){ return !n.second.getNeuronModel()->getThresholdConditionCode().empty(); }))
         {
-            os << "__shared__ volatile unsigned int shSpk[" << m_KernelBlockSizes[KernelNeuronUpdate] << "];" << std::endl;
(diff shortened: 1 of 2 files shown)


neworderofjamie:
Hi @tapaswenipathak, thanks for your work on this. It needs a little more thought though - ``shSpk`` is required e.g. [here](https://github.com/genn-team/genn/blob/master/src/genn/backends/cuda/presynapticUpdateStrategy.cc#L219) if any synapse groups have their synapse span type set to postsynaptic and thus the ``PostSpan`` strategy is used.

tapaswenipathak:
@neworderofjamie: Can you tell me the exact point where you see the synapse span is set as postsynaptic present in the `else` condition. Outside or before `else` condition, there are initialisation and use, I agree.

+ How does the PR look now? I will add separate for separate variables.

neworderofjamie:
Well, users are free to set the synapse span from their own models as well as everything else exposed via ``ModelSpec``, ``NeuronGroup``, ``SynapseGroup`` and ``CurrentSource`` (rather than their ``xxxInternal`` counterparts.

neworderofjamie:
I also noticed that our automated testing was not being run on P.R.s created from forks - I have now fixed that and you (should) be able to see a lot of compilation errors caused by the incorrect removal of ``shSpk`` in the warnings e.g. in https://gen-ci.inf.sussex.ac.uk/job/GeNN/job/genn/view/change-requests/job/PR-248/1/artifact/msg_INF900802.txt/*view*/

tapaswenipathak:
@neworderofjamie: review ping.

PR #250, opened 2019-08-13
Docs structure

### LINKED ISSUE #230: Read the docs
GeNN's doxygen-based documentation is currently deployed to a github page whenever there's a release. However this means:
1. We only have online documentation for the latest release of each major version (i.e. currently 3.2.0 and 2.2.3)
2. Documentation for master isn't currently available pre-built.

The best solution to this would be to use a service like [Read the Docs](https://readthedocs.org/) to automatically build documentation using a github webhook. However read the docs is designed for Sphinx (Python documentation tool) so a process like https://breathe.readthedocs.io/en/latest/readthedocs.html would be required.

Vigneswaran-Chandrasekaran:
Hi @neworderofjamie , I just added a basic minimal structure with `sphinx`  (in `source` directory) using `Breathe` extension to get the `Doxygen` generated `XML` files (currently in `documentation/xml/`). In a nutshell,

1. Edited `doxygen/genn-doxygen.conf` file to generate `XML` (line `GENERATE_XML = YES`).
2. Generated `XML` files are stored in `documentation/xml`.
3. Set `Sphinx` source directory, using `sphinx-quickstart`.
4. Edited `source/conf.py` file to add `Breathe` extension and added path of `doxygen` generated `XML` files. 
5. Edit `index.rst` to add `doxygenclass` etc. (which needs to be done with some discussions).
6. `Sphinx` build to generate `html` files (Saved in `Source/_build`)

Kindly have a look and as a newbie to `GeNN`, I am not certain about arranging `index.rst` file entities from `XML` files to determine how the final doc should look like. So I need your support here to finish this successfully.
fixes #230


source/_build/.buildinfo
```diff
@@ -0,0 +1,4 @@
+# Sphinx build info version 1
```
neworderofjamie: I don't think the ``_build`` folder should be added to git (in the same way that the contents of ``documentation`` is currently not

make.bat
```diff
@@ -0,0 +1,35 @@
+@ECHO OFF
```
neworderofjamie: The sphinx website suggests that the ``sphinx-quickstart`` should create a GNU makefile as well - could you add this? GeNN does fully support Windows, but our build server (and my desktop machine) run Linux so this would be convenient

neworderofjamie:
Thank you so much for looking at this! I've made a couple of suggestions but, in answer to your main point, in general we'd like the sphinx documentation to look much like [our current doxygen-based documentation](http://genn-team.github.io/genn/documentation/4/html/index.html). Things to consider are:

1. As well as auto-generated documentation, we also have a bunch of dox pages we have written - it doesn't SEEM like doxygen xml includes these so they'd need converting to rst by some other means.
2. Currently we're using doxypypy to generate documentation for GeNN's python interface and these are somewhat intermingled with the other code documentation. It might make sense to instead build these directly using sphinx.

neworderofjamie:
I also spotted https://github.com/vovkos/doxyrest which might be worth trying as an alternative to breathe

Vigneswaran-Chandrasekaran:
Hi, as you suggested, I used `doxyrest` as a bridge between `sphinx` and `doxygen`. All the `doxyrest` related files are located in `doxyrest/` directory. And I have added `doxyrest/build/` dir for your reference to the generated `html` files (which I'm worried about). I've integrated `doxyrest` pipeline to the project and finally when `sphinx` generate `html` files from `.rst`s (which is populated by `doxygen`'s XML file [in dir: `doxyrest/xml`] with the help of `doxyrest`), I wouldn't able to see any document elements :disappointed: . (However when I run `doxygen/genn-doxygen.conf`, I get `html`, `xml` and `pdf` files, in which when I try to open `index.html`, I end up with same empty documentation page). I'm pretty sure that, I am missing a small piece somewhere but hard to figure out. 

And regarding the `doxyrest` setup I followed the steps mentioned in: 
1) https://github.com/vovkos/doxyrest
2) https://vovkos.github.io/doxyrest/build-guide/

neworderofjamie:
Ok, after a lot of fiddling with readthedocs/Jenkins, this is [now basically working ](https://genn.readthedocs.io/en/readthedocs2/) on the readthedocs2 branch (the best thing may be to update your fork and work from there). Outstanding things that need investigating are:

- It would be nice if the version of GeNN was included in the sphinx documentation - this is specified globally in version.txt so you just need to read it in with Python
- Static images aren't getting copied from ``doxygen/images/`` into wherever sphinx wants them
- Release notes are somehow turned into top-level menu items - I imagine this needs fixing in the doxygen
- There are a lot of warnings being issued at various stages - are these causing any issues?

Vigneswaran-Chandrasekaran:
Oh! I didn't get a notification at all. I randomly surfed through here to check the updates on `ReadTheDocs` and surprised to see many things that I did not follow for many days. Apologies for that.  I'll look into the things mentioned :+1:

neworderofjamie:
No worries, github notifications can be a bit flaky at times!

PR #252, opened 2019-08-26
GCC 4.9 compilation fixed

costrau:
Hi,

for compilation with gcc 4.9 there is a single include missing in one of the header files. I know it is a rather old compiler, however, the support for the CPU simulator is easily added. If you want to further support gcc 4.9, I have also added a travis configuration file for continuous integration and would be happy to help setting travis up. (you can see an example at https://travis-ci.org/costrau/genn)


neworderofjamie:
Thank you so much for this! We do indeed intend to support GCC 4.9 and I _feel_ that I may well have manually made that change on our HPC system. We are currently using Jenkins for testing as Travis doesn't let us test on GPU which, as you might imagine, is pretty important for us so could you maybe revert that part?

costrau:
I did as you proposed (although the pull request is now really small :) )

neworderofjamie:
Best kind!

PR #255, opened 2019-08-29
Merge postsynaptic fix

neworderofjamie:
@costrau found this when integrating GeNN with [cypress](https://github.com/hbp-unibi/cypress/). Turns out the mechanism for merging postsynaptic model (#201) generated broken code when combined with synapse groups where the presynaptic update operates on ``linSyn`` or ``shLg``.

The reason for the broken code was a typo, but the fix wasn't entirely trivial. If you have merged postsynaptic models, multiple synapse groups may be trying to update ``dd_inSynXXXX`` simultaneously so both when using the ``linSyn`` or ``shLg`` optimisation I think you need to do something like this:

1. Rather than setting ``linSyn`` or ``shLg`` to the current value of ``dd_inSynXXXX`` before doing a presynaptic update, set them to zero
2. Rather than copying ``linSyn`` or ``shLg`` back over the top of ``dd_inSynXXXX`` after doing a presynaptic update, if merging is enabled, use an atomic add and otherwise just +=.

What do you think @tnowotny ?


neworderofjamie:
Yes, I'm in the process of adding some, I realised I never added any..

neworderofjamie:
I think these two tests cover both code paths (and should bump up the coverage somewhat) - what do you think>

neworderofjamie:
Do the tests look ok @tnowotny ?

tnowotny:
I find the display of checks a bit cryptic ... what test has failed?

neworderofjamie:
Hah yeah, I was pleased with the binary->one-hot encoding family of tests 😄 

That's a sorry tale - in adding the missing windows tests, I also added a batch file which broke the shell script that finds tests on *nix..However, it's all looking good now and all 63 tests are running on Windows and Linux (your old mac is still having CUDA problems)

tnowotny:
I will look into this now.

tnowotny:
Ok, I installed a driver update and the deviceQuery seems to be running in CUDA 9. Can you check whether things are working now?

PR #261, opened 2019-09-23
Make userproject/include useful again

neworderofjamie:
This change has been at the back of my mind for a while, but was spurred on by my attempt to decouple GeNN from BoB robotics where I had previously been stashing helpful utilities (see discussion [here](https://github.com/BrainsOnBoard/bob_robotics/pull/153)).

By default, once generated, GeNN 4 models have no dependency on GeNN itself (which is a good thing). But, there are some  utility classes in ``userproject/include`` that we use in the userprojects themselves, which you can use to make building simulations easier (spike recording code, timers etc). However, as GeNN 4 no longer needs a ``GENN_PATH`` environment variable, finding these is a little tricky. Therefore this change:

1. Adds ``-u`` option to the  ``genn-create-user-project`` tool which adds  ``userproject/include`` to the include path of the generated Makefile/MSBuild project (that part was fun...) based on the location of genn-buildmodel (which should be in the path).
2. Generalize the spike-recording classes a little to support nicer csv output and so you can add your own file format if you wish
3. Took this opportunity to add 'best' shared library model class to user project as I had ended up duplicating this in a lot of places - allows models to be loaded dynamically rather than statically linked making it possible for applications to load multiple GeNN models and to make handling models with many populations much easier. Because this was essentially the one used by PyGeNN, updated it to use derived version which just adds weird, numpy-style accessors


neworderofjamie:
Bearing in mind the mixture of SWIG and MSBuild horrors I can hardly blame you :smile: will merge and think about the default in future

PR #262, opened 2019-09-23
Nasty bug if derived parameters are used in spike-like-event threshold conditions

neworderofjamie:
Not sure how long this had been lurking but, the better error-checking in #256 made it more visible. Basically, in ``ModelSpec::finalize``, derived parameters were calculated **after** the spike-like-event threshold conditions were generated.

### DIFF
--- a/src/genn/genn/modelSpec.cc
+++ b/src/genn/genn/modelSpec.cc
@@ -148,45 +148,6 @@ void ModelSpec::setPrecision(FloatType floattype /**<  */)
 
 void ModelSpec::finalize()
 {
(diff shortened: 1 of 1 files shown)


neworderofjamie:
I worry our current tests don't actually actively cover derived parameters - these both show suspicious gaps:

https://codecov.io/gh/genn-team/genn/src/master/src/genn/genn/neuronGroup.cc#L195
https://codecov.io/gh/genn-team/genn/src/master/src/genn/genn/synapseGroup.cc#L352

I might merge this, but add an issue to add some feature tests for derived parameters in the various locations in which they can appear.

PR #269, opened 2019-10-04
Accessing queued vars

neworderofjamie:
As discussed, there was no way for users to easily access the contents of queued neuron state variables. This is particularly tricky as, whether state variables are queued or not depends on if there are any outgoing synapse groups with axonal delays and with weight update models which access the variables. 

This P.R. does the following:
1.  Generates ``pullCurrentXXXYYYFromDevice()`` and ``pushCurrentXXXYYYToDevice()`` functions (where, as usual, ``XXX`` is the name of the variable and ``YYY`` is the name of the neuron group)
2. Generates ``currentXXXYYY`` macros (in the style of ``spike_XXX`` macro) for accessing variable contents
3. Adds getter functions ``getCurrentXXXYYY()`` functions to retrieve these at runtime.

While I am going to merge this into the SpineML branch that original led to me realising this omission, I think this needs a little bit more thought and discussion, namely:

- In order to access these variables from the SpineML frontend or anything PyGeNN-based, getter functions are required, but should they always be generated? Although they are useful, they are _further_ increasing the size of runner.cc 
- Is there any point in generating macros at all? They _might_ slightly reduce overheads but I imagine it's insignificant.
- For consistancy/to tidy up a fair amount of recording code should we add gettes for current spikes and spike like events?


neworderofjamie:
> I think os far, so good. Regarding a more general improvement of data access tools:
> 
>     * Should we do the macros? - I can't see much harm in providing them; while not saving massive overhead, the queue pointers could be confusing to people not too familiar with the insides of the code.
> 
I definitely agree we need **something** to save people from having to deal with spike queues themselves but my point was more do we need macros **and** getter functions.
>     * Should we do getters for the spike queues? - Probably a good idea for consistency.
> 
I will add this
>     * you seem to be worried about the size of runner.cc and resulting object code: Has this been and issue or is becoming one? If so, one could add a preference to switch getter generation off?
> 
I am in general getting slightly worried about generated code size and compilation time but, as CPU only compilation is pretty speedy, I would imagine long compile times are more to do with CUDA kernel compilation.
>     * different question: do we have a slight syntax inconsistency with the spike_YYY macros versus currentXXXYYY (underscore vs not)? I remember when we did those macros for the first time, we started doing underscores for more user facing things versus the camelback names "inside"; but this was probably not well thought out because there is a continuum of use cases and one consistent naming scheme would probably make more sense.
Haha, I had no idea that that was the source of the inconsistency - by the time I got to it - that was definitely not clear :smile: I think more or less everything is now ``lowerCamelCase`` (aside from those macros). Personally I like ``SCARY_ALL_CAPS`` to make it clear that you're using macros and thus have to be careful but having the macros for accessing current variable state use this would then be _very_ inconsistent...

tnowotny:
* Should we do the macros? - I can't see much harm in providing them; while not saving massive overhead, the queue pointers could be confusing to people not too familiar with the insides of the code.
I definitely agree we need something to save people from having to deal with spike queues themselves, but my point was more do we need macros and getter functions.

oh, I see, I completely misunderstood. "getters only" perfectly fine by me (overhead indeed likely to be completely negligible). Only would have to phase out any existing macros slowly in case people have used them in the past ...

About naming inconsistency: If we get rid of the macros eventually, then that problem solves itself ... (i.e. everything lowerCamelCase ...)

neworderofjamie:
What do you think @tnowotny? I implemented those changes.

PR #272, opened 2019-10-10
Expose select cuda device

chanokin:
+ Merged master into branch


pygenn/genn_model.py
```diff
@@ -125,6 +128,17 @@ def __init__(self, precision=None, model_name="GeNNModel",
         self.current_sources = {}
         self.dT = 0.1
 
+    @property
+    def selected_gpu(self):
+        return self._selected_gpu
+
+    @selected_gpu.setter
+    def selected_gpu(self, v):
```
neworderofjamie: I don't think the behaviour of this setter makes sense - imo if you set selected GPU and the backend isn't CUDA it should fail rather than mess with your backend choice

chanokin: True, we should respect their choice.

PR #273, opened 2019-10-15
Device types

neworderofjamie:
Starting to put the pieces together to allow GeNN to take advantage of devices that support double throughput half-precision (Jetsons, newer Teslas and _seemingly_ all Turings). GeNN 4 is somewhat stricter in seperating what is host code i.e. compilable with a normal C++ compiler vs what is device code i.e. compilable with whatever the backend desires (NVCC in the case of the CUDA backend). This P.R. formalizes this distinction for types so e.g. ``curandState`` and ``half`` can only be used on device and thus cannot be referenced from (platfom-agnostic) host code.


neworderofjamie:
Also, out of interest, I re-ran the 'accuracy' analysis from the frontiers paper on a 16-bit version of the microcircuit:
Figure_1
if you zoom in some 'detail' is different but otherwise it would appear that the dynamics are maintained (perhaps unsurprising as half-precision floating point obviously has better dynamic range than 16-bit fixed point as used on SpiNNaker)

PR #274, opened 2019-10-15
Option to use 'narrower' types for sparse matrix indices

neworderofjamie:
With sparse, static synapses, the postsynaptic indices take up as much memory as the actual weights. However, if the postsynaptic population has less than 65535 neurons (common case), the postsynaptic indices can be stored in ``uint16_t`` halving memory. This P.R. implements this as an optional feature as it is likely to break any existing code that accesses the sparse structure on the host (generic solution for any such code in future is to use templates). Combining this feature with #273 means that the full sized microcircuit takes 1756 MB rather than 3146 MB.


include/genn/genn/modelSpec.h
```diff
@@ -545,6 +550,9 @@ class GENN_EXPORT ModelSpec
     //! What is the default location for sparse synaptic connectivity? Historically, everything was allocated on both the host AND device
     VarLocation m_DefaultSparseConnectivityLocation; 
 
+    //! The default for whether narrow i.e. less than 32-bit types are used for sparse matrix indices
+    bool m_DefaultNarrowSparseIndEnabled;
```
tnowotny: is this set to false by default somewhere?

src/genn/genn/modelSpec.cc
```diff
@@ -37,7 +37,8 @@
 ModelSpec::ModelSpec()
 :   m_TimePrecision(TimePrecision::DEFAULT), m_DT(0.5), m_TimingEnabled(false), m_Seed(0),
     m_DefaultVarLocation(VarLocation::HOST_DEVICE), m_DefaultExtraGlobalParamLocation(VarLocation::HOST_DEVICE),
-    m_DefaultSparseConnectivityLocation(VarLocation::HOST_DEVICE), m_ShouldMergePostsynapticModels(false)
+    m_DefaultSparseConnectivityLocation(VarLocation::HOST_DEVICE), m_DefaultNarrowSparseIndEnabled(false),
```
neworderofjamie: You worried me for a minute, but here!

tnowotny: Ah, I see - didn't catch it first time round.

Issue #277, opened 2019-11-04
Support for using 2D grids for large kernels broken


neworderofjamie:
When grids that were larger than the maximum size of a single grid dimension were created, GeNN 2 and 3 would create a 2D grid. This ability has been lost!

PR #278, opened 2019-11-04
Increase maximum supported grid size

neworderofjamie:
Due to #277, GeNN 4 only supports generating 1D grids but, as on devices with CC > 3.0, the maximum x dimensions can be 4294967295 this is unlikely to be problematic. However, for some reason, we were checking against the y-dimension instead of the x-dimension. This was the behaviour in GeNN < 4 as well but I don't know why - any memories @tnowotny? Are squarer grids better for some reason?


tnowotny:
Back then we did it to avoid the rather restrictive size limit on 1D grids. But I cannot recollect any reason why to check for the maximum y-dimension (truth be told, I think back then they always were the same).

neworderofjamie:
Ok cool - glad there wasn't some hidden insight here! In that case this makes for a good stop-gap solution until #277 gets fixed

PR #280, opened 2019-11-06
Bitmask connectivity optimizations

neworderofjamie:
This is the very indirect result of my attempts to implement the fabled vectorized half-precision mode...and finding out it performed exactly the same as non-vectorized (even on the Jetson) :cry: I'm still trying to fully understand the profiling results, but it would seem that this is because these kernels end up being constrained by the latency of the atomicAdd instructions - information about their implementation is hard to come by but, they are implemented within the L2 cache so _are_ fire-and-forget but _presumably_ with some sort of queuing occuring in the cache. ANYWAY, my thinking was then, what kind of synapse kernels can we optimise that don't need atomics so I thought about the good 'ole bitmask connectivity and did some profiling - basically if you have a beast of a GPU like a Titan V with loads of threads and fast memory the current implementation works really well (twice as fast as ``SPARSE_GLOBALG`` for large networks on the Titan V). However, it performs very badly on small GPUs (half as fast as ``SPARSE_GLOBALG`` on Jetson TX1). 

In this PR I add a new presynaptic update strategy to the CUDA backend (and function to the CPU backend where this approach gives an even more impressive 5x speedup) which uses postsynaptic parallelism, but with each thread responsible for processing a whole word of connectivity. Within this word, we use a count-leading-zeros intrininsic to loop efficiently through the synapses that actually exist. Each thread has 32 scalars of shared memory into which it accumulates the results in (with a block size of 32 and single-precision this is only 1kB per block which is not **that** much on a modern SM).

Doing some benchmarking with a scalable version of the Vogels Abbott benchmark model (https://github.com/neworderofjamie/genn_examples/blob/master/va_benchmark/model.cc), this gives a pretty decent 2-3x speedup  over the current BITMASK implementation on small GPUs but, on the Titan V, it's significantly slower - probably because you would need 160000 postsynaptic neurons to even occupy all 5000 of its CUDA cores let alone hide any latency.

GTX 1050ti (low-end Pascal)
Jetson TX1

Because this approach requires that each row of bitmask is padded to a word boundary, it could break existing code so is only enabled when ``GENN_PREFERENCES.enableBitmaskOptimisations`` is set.

Finally, I thought this would be a good PR for @jamesturner246 to look at as, adding the ``PostSpanBitmask`` presynaptic update strategy requires similar changes to what will be required to add support for the backend-side of convolutional connectivity.


neworderofjamie:
@tnowotny and @jamesturner246 - any chance of that review?

neworderofjamie:
Yeah, the surrounding minor refactoring obscures the core of it somewhat.As I said, it's off by default anyway...

jamesturner246:
Hi Jamie. The same story here, really... I am still catching up with all the changes made since I last used it. I will drop by today for a more low-level discussion, if that's OK with you.

PR #285, opened 2019-12-17
Python tweaks

neworderofjamie:
I was doing some stuff with PyGeNN and found a tweaked a few small things (mostly windows related)

- Updated the windows bit of the Jenkinsfile to use the same version of numpy as else where (this could probably be revisited as I suspect whatever the library that was breaking on 1.15 is now fixed but still)
- Used multiple cores when building generated code in PyGeNN
- Exposed GeNN kernel timing to Python


jamesturner246:
Hey Jamie, is there a reason pygenn can't use numpy > 1.15? TF uses higher and pip always complains. Also is the swig memory leak thing an issue, or not harmful?

neworderofjamie:
Apologies for totally missing this! The numpy restriction was due to an issue with a PyNN release which *may* now have been fixed - haven't had time to investigate though (feel free to install whatever version of numpy you like on top/change your setup.py locally). The SWIG memory leak isn't new and, in this case, isn't harmful as the backend object basically has the same lifetime as the python script, *but* it is symptomatic of me not fully understanding how SWIG memory management works.

neworderofjamie:
@jamesturner246  could you have a quick look at this? I think the kernel timings stuff'll be useful for you for a start! probably need to merge #301 first

jamesturner246:
I'm getting the following:

```
Traceback (most recent call last):
  File "conversion_example.py", line 99, in <module>
    main()
  File "conversion_example.py", line 77, in main
    tg_model.create_genn_model(dt=1.0, rng_seed=1, rate_factor=1.0, input_type='if_cs')
  File "/its/home/jt273/workspace/test_tensor_genn/tensor_genn/model.py", line 292, in create_genn_model
    self.g_model.build()
  File "/its/home/jt273/workspace/genn-master-pr/pygenn/genn_model.py", line 432, in build
    check_call(["make", "-j", cpu_count(logical=False), "-C", output_path])
  File "/usr/lib/python3.6/subprocess.py", line 306, in check_call
    retcode = call(*popenargs, **kwargs)
  File "/usr/lib/python3.6/subprocess.py", line 287, in call
    with Popen(*popenargs, **kwargs) as p:
  File "/usr/lib/python3.6/subprocess.py", line 729, in __init__
    restore_signals, start_new_session)
  File "/usr/lib/python3.6/subprocess.py", line 1295, in _execute_child
    restore_signals, start_new_session, preexec_fn)
TypeError: expected str, bytes or os.PathLike object, not int
swig/python detected a memory leak of type 'CodeGenerator::CUDA::Backend *', no destructor found.
```

Looks like it didn't like     check_call(["make", "-j", cpu_count(logical=False), "-C", output_path]).

neworderofjamie:
Sorry, fixed that in another branch, try now

neworderofjamie:
Have you set ``model.timing_enabled = True`` beforehand?

jamesturner246:
OK that did it. Note to others, this should be done before `build()` / `load()`.

So these times reported are for the previous call to `step_time()`?

neworderofjamie:
It's exactly the same as the C++ interface - the reported times are for the whole simulation so the totals get added to after every ``stepTime``/``initialize``/``initializeSparse`` call.

jamesturner246:
Great. All good. Thanks.

PR #287, opened 2019-12-24
Fix repeating code

9inpachi:
Hi,
The common data being sent to the code stream can be placed before the if statement avoiding code repetition.


neworderofjamie:
Looks good - thanks!

PR #291, opened 2020-01-06
Fixed extra global parameters in PyGeNN

neworderofjamie:
As, I fear, with many things in PyGeNN these were a little broken outside of the way PyNN GeNN uses them:
- Extra global parameters were entirely broken on synapse groups - these are now implemented for E.G.P.s associated with weight update models, postsynaptic models and sparse connectivity initialiser snippets
- 'Scalar' E.G.P.s were not implemented at all - this is really useful when you want to e.g. upload an entire epoch of training data to a pointer E.G.P. and set an index to somewhere within it.

This P.R. also hacks at SWIG to fix overriding the maximum row and column length calculation functions in sparse connectivity initialisers from Python.


neworderofjamie:
@jamesturner246 ping!

jamesturner246:
Hi Jamie. pygenn/genn_model.py: line 120 should be ```self._selected_gpu = selected_gpu```.

I'm also having a mismatch between numpy weights added via

```
syn = add_synapse_population('syn01', 'SPARSE_INDIVIDUALG', genn_wrapper.NO_DELAY,
                    neuron_pop_trg, neuron_pop_src,
                    'StaticPulse', {}, {'g': np_w.copy()}, {}, {},
                    'DeltaCurr', {}, {})
syn.set_sparse_connections(pre_inds.copy(), post_inds.copy())
```

and the generated pygenn weights, where ```syn.vars['g'].view.shape``` does not equal ```np_w.shape```.

The ```add_extra_global_param``` API looks fine, but I'm getting the following:

```
The Traceback (most recent call last):
  File "./conversion_example_2.py", line 112, in <module>
    main()
  File "./conversion_example_2.py", line 90, in main
    tg_model = TGModel(tf_model)
  File "/its/home/jt273/workspace/tensor_genn_new/model.py", line 37, in __init__
    self.create_genn_model(dt=dt)
  File "/its/home/jt273/workspace/tensor_genn_new/model.py", line 358, in create_genn_model
    self.neuron_pops[-1].add_extra_global_param('thr', 1.0)
  File "/its/home/jt273/workspace/genn/pygenn/genn_groups.py", line 264, in add_extra_global_param
    self._add_extra_global_param(param_name, param_values, self.neuron)
  File "/its/home/jt273/workspace/genn/pygenn/genn_groups.py", line 64, in _add_extra_global_param
    assert param_type is not None
AssertionError
```

jamesturner246:
nrn model code is

```Python
    if_model = genn_model.create_custom_neuron_class(
        'if_model',
        param_names=['Vres'],
        var_name_types=[('Vmem', 'scalar'), ('Vmem_peak', 'scalar'), ('nSpk', 'unsigned int'), ('Vthr', 'scalar')],
        sim_code='''                                                                                                
        $(Vmem) += $(Isyn) * DT;                                                                                    
        $(Vmem_peak) = $(Vmem);                                                                                     
        ''',
        reset_code='''                                                                                              
        $(Vmem) = $(Vres);                                                                                          
        $(nSpk) += 1;                                                                                               
        ''', 
        threshold_condition_code='''                                                                                
        $(Vmem) >= $(thr)                                                                                           
        '''
    )
```

neworderofjamie:
Ahh yes, the data structure underneath the sparse connectivity is not the same as it 'used' to be (mostly for reasons of on-GPU initialization) https://github.com/genn-team/genn/pull/180 might be a useful reference. There is a helper to extract weight update model variables in the correct shape https://github.com/genn-team/genn/blob/master/pygenn/genn_groups.py#L402-L419

neworderofjamie:
That model doesn't seem to have an extra global parameter called 'thr'? You would need to add it to the model with a ``extra_global_params=[("thr", "scalar")],`` kwarg.

neworderofjamie:
The PyGeNN syntax is all a bit ad-hoc so, if you have any suggestions, feel free to raise issues/make a PR!

jamesturner246:
I see. So the add_global_param sets the param, but it still needs to first be defined in the group model. Will look at the staggered weights thing, and converting them to non-staggered.

neworderofjamie:
``set_extra_global_param`` may be a better name :worried:

tnowotny:
I have to agree ... ```set_extra_global_param``` would probably be more intuitive...

neworderofjamie:
As I said, PyGeNN is still rather adhoc, as long as you update PyNN GeNN (as @chanokin relies on it) feel free to rename/improve anything!

jamesturner246:
Have a commit changing it to ```set_..._global_param```, and fixing ```_selected_gpu```. Shall I PR, or just merge it to this (python_scalar_egp) branch?

neworderofjamie:
Sweet! Just merge it in - we probably have enough branches right now :)

jamesturner246:
Actually, hold on. ```neuron_pop.set_var('Vmem', 0.0)``` is not resetting membrane potentials. Neither on host nor device.

neworderofjamie:
That's another confusing bit of API I'm afraid, once the model is loaded you need to use views to change variable values

neworderofjamie:
Also,  so as to be a bit nicer to any users we might have, it might make sense to keep the old method as well (just have it fall the new one) and mark it as deprecated with https://deprecation.readthedocs.io/en/latest/

jamesturner246:
OK. I remember you mentioned separating model generation and execution. I guess this is one of those situations where it would be beneficial - i.e. have separate getter/setter functions for buildtime and runtime.

neworderofjamie:
Yeah exactly - this was basically the thing that made me think it would be a good idea

jamesturner246:
I'm having trouble pushing an extra global param to device. I use the following.
```Python
            neurons.extra_global_params['Vthr'].view[:] = new_Vthr
            genn_model.push_var_to_device(neurons.name, 'Vthr')
```
and get
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  You cannot push var 'Vthr' to population 'if1'
```
Is there another function for adding extra global vars?

neworderofjamie:
If you set up your egp values before you load your model, PyGeNN will automatically push extra global params if they are pointers (as you say, pushing isn't required for non-pointer extra global params). If you want to update pointer EGPs at run time you need to delve a little and call ``genn_model._slm.push_extra_global_param(pop_name, egp_name, len(egp_data.values))``

neworderofjamie:
I added the deprecated warnings I mentioned and merged in master so I could get this merged. If you want to use your previous commit for Tensor GeNN development (which I think would be a reasonable idea) you probably wanna make a new branch

Issue #292, opened 2020-01-07
Slow merging for large networks with custom models defines in Python


neworderofjamie:
Because, for example:
```c++
bool WeightUpdateModels::Base::canBeMerged(const Base *other) const
{
    return (Models::Base::canBeMerged(other)
            && (getSimCode() == other->getSimCode())
            && (getEventCode() == other->getEventCode())
            && (getLearnPostCode() == other->getLearnPostCode())
            && (getSynapseDynamicsCode() == other->getSynapseDynamicsCode())
            && (getEventThresholdConditionCode() == other->getEventThresholdConditionCode())
            && (getSimSupportCode() == other->getSimSupportCode())
            && (getLearnPostSupportCode() == other->getLearnPostSupportCode())
            && (getSynapseDynamicsSuppportCode() == other->getSynapseDynamicsSuppportCode())
            && (getPreSpikeCode() == other->getPreSpikeCode())
            && (getPostSpikeCode() == other->getPostSpikeCode())
            && (getPreVars() == other->getPreVars())
            && (getPostVars() == other->getPostVars())
            && (isPreSpikeTimeRequired() == other->isPreSpikeTimeRequired())
            && (isPostSpikeTimeRequired() == other->isPostSpikeTimeRequired()));
}
```
Requires a large number of C++->Python calls, complexity aside, merging becomes very slow. Possible solutions would be:
- A hash-based merging algorithm where all required properties are hashed and these are used as the initial basis for compatibility testing
- Adding an intermediate C++ class between ``WeightUpdateModel::Base`` and Python custom models which caches the results after making calls for the first time

PR #293, opened 2020-01-08
Microcircuit userproject

neworderofjamie:
I've been thinking we could do with some new userprojects for a while so, in this PR, I've added a slightly tidied up version of the microcircuit model which demonstrates various newer GeNN features including dynamically loading models into simulations. In the process of doing this I also added the following models:

* ``InitVarSnippet::NormalClipped`` - For initialising weights, clipping the normal distribution is often required to stop the 'tail' resulting in sign changes - This is a standard PyNN connector
* ``InitVarSnippet::NormalClippedDelay`` - When specifying delays, not only do you want it clipped but, ideally, you'd like to specify stuff in ms but have them converted to integer timesteps automatically
* ``InitSparseConnectivitySnippet::FixedNumberTotalWithReplacement`` - It's a standard PyNN connector and comes with the helper function to calculate row lengths in userproject/include
* ``InitSparseConnectivitySnippet::FixedNumberPostWithReplacement`` - Again it's a standard PyNN connector type which isn't trivial to implement correctly (was previously implemented in the IzhSparse user project)
* ``CurrentSourceModels::PoissonExp`` - The generality of this is a little more dubious, but basically this adds an input current to a neuron which should match that delivered by a population of Poisson sources connected one-to-one with exponential synapses. Advantages are no spike delivery overhead and support for isi < dt as multiple spikes can be summed per timestep.

This has a slight ulterior motive as, adding the models to GeNN, works around #292 somewhat :smile:


userproject/PotjansMicrocircuit_project/model/runner_guid.txt
```diff
@@ -0,0 +1 @@
+62182A2D-DE80-4612-BF02-C3FF910EF67B
```
tnowotny: What is this?

userproject/include/generateRun.h
```diff
@@ -154,7 +155,8 @@ class GenerateRunBase
             cmd += " -c";
         }
 
-        cmd += " && make clean all";
+        const unsigned int numThreads = std::thread::hardware_concurrency();
+        cmd += " && make -j " + std::to_string(numThreads) + " clean all";
```
tnowotny: interesting idea ... I doubt it will speed up massively?

userproject/PotjansMicrocircuit_project/model/runner_guid.txt
neworderofjamie: Magic that means GeNN will re-generate a visual studio project with the magic number as the solution

userproject/include/generateRun.h
neworderofjamie: Well, now the generated code is split across multiple modules (synapse update, neuron update, init and runner), there's at least 4 decently independent jobs so it makes a reasonable difference actually

include/genn/genn/currentSourceModels.h
```diff
@@ -71,4 +71,31 @@ class GaussianNoise : public Base
...
+        "$(current) += $(Init) * (scalar)(numSpikes - 1);\n"
+        "$(injectCurrent, $(current));\n"
+        "$(current) *= $(ExpDecay);\n");
+
+    SET_PARAM_NAMES({"weight", "tauSyn", "rate"});
+    SET_VARS({{"current", "scalar"}});
+    SET_DERIVED_PARAMS({
+        {"ExpDecay", [](const std::vector<double> &pars, double dt){ return std::exp(-dt / pars[1]); }},
+        {"Init", [](const std::vector<double> &pars, double dt){ return pars[0] * (1.0 - std::exp(-dt / pars[1])) * (pars[1] / dt); }},
+        {"ExpMinusLambda", [](const std::vector<double> &pars, double dt){ return std::exp(-(pars[2] / 1000.0) * dt); }}});
```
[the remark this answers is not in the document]
neworderofjamie: The documentation was definitely lacking there - done a bit of updating

include/genn/genn/initSparseConnectivitySnippet.h
```diff
@@ -212,4 +212,107 @@ class FixedProbabilityNoAutapse : public FixedProbabilityBase
...
+            return (unsigned int)pars[0];
+        });
+
+    SET_CALC_MAX_COL_LENGTH_FUNC(
+        [](unsigned int numPre, unsigned int numPost, const std::vector<double> &pars)
+        {
+            // Calculate suitable quantile for 0.9999 change when drawing numPost times
+            const double quantile = pow(0.9999, 1.0 / (double)numPost);
+
+            // In each row the number of connections that end up in a column are distributed
```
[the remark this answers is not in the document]
neworderofjamie: thanks

PR #296, opened 2020-01-23
Improve README formatting and add source command to Linux/Mac installation

9inpachi:
Hi,

The source command helps in actually updating the $PATH variable after putting the value for GeNN's bin directory in `~/.bash_profile`.  This helps in avoiding a restart of the system or re-logging of the user.


neworderofjamie:
That looks much nicer - well done! My only comment wold be that maybe the bit about sourcing should be a note at the end like the Windows equivalent as it's not a _required_ step. Maybe something like 
"Note that you must either logout and in again or run ``source ~/.bash_profile`` for the changes to ``.bash_profile`` to take effect."

9inpachi:
I had that in mind initially. But since the source command is a go-to way, I preferred it.
I will add it to the end as you suggest.

PR #298, opened 2020-01-24
Smarter row build state

neworderofjamie:
This P.R. slightly extends the syntax for defining connectivity initialization models so you can initialize row-wide state variables with expressions (which get substitutions applied etc) as well as with constants. This lets you express the fixed total number of synapses connectivity in such a way that the row length is only read once per row. This actually makes a pretty sizable difference to the performance - 20% faster procedural synapse update on my home machine. Because it uses the same code path, this functionality also now applies to neuron additional inputs although I can't think of a use-case off hand...

**NOTE** this is currently set to merge into #293 as it updates the built in models

### DIFF
--- a/include/genn/genn/code_generator/codeGenUtils.h
+++ b/include/genn/genn/code_generator/codeGenUtils.h
@@ -1,17 +1,12 @@
 #pragma once
 
 // Standard includes
-#include <iomanip>
-#include <limits>
 #include <string>
-#include <sstream>
 #include <unordered_map>
 #include <vector>
 
 // GeNN includes
 #include "gennExport.h"
-#include "models.h"
-#include "snippet.h"
 #include "variableMode.h"
 
 // GeNN code generator includes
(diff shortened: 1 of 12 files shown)


neworderofjamie:
Sadly that 20% doesn't translate to the multi-area simulation - presumably as, for once, there is enough computation occuring to hide memory latency!

include/genn/genn/code_generator/codeGenUtils.h
```diff
@@ -106,44 +101,6 @@ bool regexFuncSubstitute(std::string &s, const std::string &trg, const std::stri
 void functionSubstitute(std::string &code, const std::string &funcName,
                         unsigned int numParams, const std::string &replaceFuncTemplate);
 
-//--------------------------------------------------------------------------
-//! \brief This function writes a floating point value to a stream -setting the precision so no digits are lost
-//--------------------------------------------------------------------------
-template<class T, typename std::enable_if<std::is_floating_point<T>::value>::type* = nullptr>
```
tnowotny: I am not sure I can parse this ... is it the same as
```std::enable_if_t<std::is_floating_point<T>::value, int> = 0``` ?
Oh boy - feels like I have gotten totally out of touch with more recent C++ standards.

neworderofjamie:
Yeah, the enable_if syntax is particularly awful - I seem to remember this article helped me grasp it https://eli.thegreenplace.net/2014/sfinae-and-enable_if/. All that's happening with that function is that I'm moving it to a different utils module though! Give me a shout when you're around tomorrow and I can attempt to explain though!

tnowotny:
Thanks for the link. I (almost) get it now. Yes, I realised that this had nothing to do with the pull request. I didn't really find the meat of the pull request in the file changes, though.

neworderofjamie:
Hurrah! What 'meat' there is as follows:

1. Switch the ``ParamVal`` structure, used for specifying these state variables, to use a string rather than a double as it's value type and add a constructor taking a double for backward compatibility - https://github.com/genn-team/genn/blob/smarter_row_build_state/include/genn/genn/snippet.h#L138-L143
2. Apply suitable substitutions to the string when writing it out in synaptic matrix initialisation code (https://github.com/genn-team/genn/blob/smarter_row_build_state/src/genn/genn/code_generator/generateInit.cc#L344-L348) and procedural connectivity code (https://github.com/genn-team/genn/blob/smarter_row_build_state/src/genn/genn/code_generator/generateSynapseUpdate.cc#L161-L165)
2. Use the new features in some sparse connectivity init snippets - https://github.com/genn-team/genn/blob/smarter_row_build_state/include/genn/genn/initSparseConnectivitySnippet.h#L277-L287

PR #302, opened 2020-02-15
Use correct max function in models

neworderofjamie:
@AdityaKapoor74 found this issue when building MBody1 on CPU -  ``max`` doesn't resolve to anything useful in <cmath> so we should use ``fmax``


neworderofjamie:
Either self-defined max functions or ``using names pace std`` I would wager

PR #304, opened 2020-02-17
Automatically delete old builds on Jenkins

neworderofjamie:
I noticed that the Jenkins server is basically out of disk space. I did some manual deleting of old build but this should set some automatic build discarding rules based on https://stackoverflow.com/questions/47778019/how-to-use-options-in-a-jenkins-scripted-pipeline - if you think archiving the 100 latest builds isn't ideal, feel free to change


tnowotny (approved):
Yeah ... 100 should definitely be enough I'd think.

PR #305, opened 2020-02-17
Don't #include runner in modules

neworderofjamie:
One of the hacky bits of getting GeNN 4's multiple module CUDA build to work was that, when we were compiling each module independently to cubin while optimizing block sizes, there were persistent warnings about extern variables being treated as static. These didn't matter as the code was never run, but I couldn't find any way of suppressing them as they came from the (undocumented) **device** compiler rather than gcc or whatever so I ended up #included the runner in each module when optimizing block sizes.

However, with large models, the runner becomes very large and compiling it an extra 6 times or whatever is the last thing one wants to do. I finally found the (undocumented) nvcc command line arguments on a dark corner of stack overflow and incorporated them.

**Note** on Windows nvcc is not normally invoked directly, but in the special case of optimizing block sizes, it is so the same approach (irritatingly the warning numbers are different) works on Windows as well.

### DIFF
--- a/include/genn/genn/code_generator/generateInit.h
+++ b/include/genn/genn/code_generator/generateInit.h
@@ -15,5 +15,5 @@ class ModelSpecMerged;
 namespace CodeGenerator
 {
 void generateInit(CodeStream &os, const MergedEGPMap &mergedEGPs, const ModelSpecMerged &modelMerged,
-                  const BackendBase &backend, bool standaloneModules);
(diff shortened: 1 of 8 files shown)


neworderofjamie:
@tnowotny - any chance of a quick review on this one?

neworderofjamie:
Yeah, sadly, CUDA has a lot of undocumented corners....

PR #313, opened 2020-04-09
Move CHECK_CUDA_ERRORS to definitionsInternal

9inpachi:
Hi,
Since all device specific functionality is supposed to be in definitionsInternal. I think `CHECK_CUDA_ERRORS` should be there as well unless the user is supposed to have access to it.


9inpachi:
Looks like the tests had passed on Jenkins.

Issue #314, opened 2020-04-09
Behaviour of $(sT_pre) and $(sT_post) is not ideal


neworderofjamie:
In some event-driven learning rules, you want to be able to calculate ``$(t) - $(sT_pre)`` in the presynaptic spike-driven sim_code. However, currently ``$(sT_pre)`` gets set in the neuron kernel so essentially refers to the current spike being processed in the sim_code i.e. it'll always be ``$(t) - DT``.  The only current solution is to add an extra per-synapse variable to track the time of the last pre spike but this is very wasteful of memory and innefficient. When these variables **should** get updated needs more thought!

neworderofjamie:
While, if using presynaptic parallelism, updating $(sT_pre) could be added at the end, all other cases probably need an additional kernel. Implementation could be an additional virtual on ``WeightUpdateModel::Base`` like ``isPrevPreSpikeTimeRequired``

tnowotny:
naively (can't fully recollect the order of things) but is this not a case where neurons could register whether they are spiking or not in some flag 'spiking' when the spike is first detected and then update sT only at the beginning of the next time step (resetting the "spiking" flag to false)?

neworderofjamie:
Actually, that's a much better idea, maaaybe a flag isn't even required and the spike arrays from the previous time step can be re-used

PR #317, opened 2020-04-17
Fix for nasty bug in single-threaded CPU backend

neworderofjamie:
The standard pattern for implementing STDP traces using pre or postsynaptic variables is to use code something like:

```c++
$(C) = ($(C) * exp(-dt / $(tauC))) + 1.0;
```

This worked fine in the CUDA backend but, in the CPU backend, ``sT`` was updated **before** this code was run meaning that dt was always 0. Note this is being merged into the 4_2_0 branch as I'm planning a 4.2.1 release - intended to be the last release before kernel merging finally gets merged - containing a few fixes.


neworderofjamie:
Combination of lack of tests of this particular case and no one testing models which use it on CPU I guess 😟 

#314 discusses the more general order of update issue

PR #318, opened 2020-04-20
Windows fixes

neworderofjamie:
Another one for the 4.2.1 release! This solves a couple of Brian2 issues:
- There's some dubious stuff in randomlib which would normally cause warnings but, if [SDL checks](https://docs.microsoft.com/en-us/cpp/build/reference/sdl-enable-additional-security-checks) are enabled (which by default they are), these are promoted to errors
- For the semicolon-seperated list of include paths to get through to the build system, it needs to be quoted (otherwise semicolons are treated as whitespace). Also quoted all other arguments for future-safety!


neworderofjamie:
Hah, who was the author of the terrifying batch script version of genn-buildmodel?

tnowotny:
ah ... yes ... that would have been me ...🙄

PR #321, opened 2020-04-24
Fix compilation with gcc 4.9

costrau:
Hi,

recent changes in userproject's SpikeRecorder will fail to compile with gcc 4.9. Although gcc4.9 is fairly old, I think its worth to keep it compatible as long as possible.

Cheers,
Christoph

### DIFF
--- a/userproject/include/spikeRecorder.h
+++ b/userproject/include/spikeRecorder.h
@@ -27,7 +27,7 @@ public:
     }
 
 // GCC 4.x does not provide a move constructor for ofstream
-#if defined(__GNUC__) && (__GNUC__ > 4)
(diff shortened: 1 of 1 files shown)


userproject/include/spikeRecorder.h
```diff
@@ -26,11 +26,14 @@ class SpikeWriterText
         }
     }
 
+// GCC 4.x does not provide a move constructor for ofstream
+#if defined(__GNUC__) && (__GNUC__ > 4)
```
neworderofjamie, line 30: I'm afraid a more complex test is need to deal with Visual C++ and clang (which for extra fun also defines ``__GNUC__``):

```c++
!defined(__GNUC__) ||     __clang__ ||  __GNUC__ > 4
```
maybe

PR #323, opened 2020-04-29
Support for weight sharing in GeNN

neworderofjamie:
C++ syntax looks like:
```c++
model.addSynapsePopulation<WeightUpdateModels::StaticPulse, PostsynapticModels::DeltaCurr>(
        "Syn1", SynapseMatrixType::SPARSE_INDIVIDUALG, NO_DELAY, "Pre", "Post1",
        {}, staticSynapseInit,
        {}, {});
model.addSlaveSynapsePopulation<PostsynapticModels::DeltaCurr>(
        "Syn2", "Syn1", NO_DELAY, "Pre", "Post2",
        {}, {});
```

And Python syntax looks like:
```python
syn1 = model.add_synapse_population(
    "Syn1", "DENSE_INDIVIDUALG", NO_DELAY,
    pre, post1,
    "StaticPulse", {}, {"g": w.flatten()}, {}, {},
    "DeltaCurr", {}, {}))
model.add_slave_synapse_population(
    "Syn2", "Syn1",
    NO_DELAY, pre, post2,
    "DeltaCurr", {}, {})
```
Testing on the (very small) three layer network from PyGeNN tutorial on my GeForce 1050ti (low-end Pascal):
weight sharing sim time  s  and no weight sharing sim time  s

only shows speedup at very large batch sizes but, this model only has around 400kb of weights so, at small scales, the whole model can fit into L2 cache.


jamesturner246:
Awesome! :D I will take it for a spin and report some memory saving data

neworderofjamie:
I'm not too happy about the Python syntax - I think all the population parameters should take either a string or a reference to a population

neworderofjamie:
Yeah, please do! I'm not super-happy about the Python syntax but otherwise I think it should work fine (I added a bunch of feature tests that seem fine anyway) - I might take this opportunity to improve PyGeNN so you can pass either strings or ``*Group`` objects to functions that take references to populations

neworderofjamie:
Ok, as a major user of PyGeNN, I'd very much appreciate your feedback on the last two commits @jamesturner246

jamesturner246:
Looks good. Being able to add synapses and input groups seems useful for interactive sessions. I'm using the group objects directly, right now.

I especially like being able to push and pull variables directly from groups, though.  Looks less awkward / feels more encapsulated.

It might be cool to be able to initialise EGP in the add_population calls, rather than (forgetting to do) using set_extra_global_param calls. Similarly, it might be nice to include sparse connection setting with add_synapse_population. They seem pretty related.

Just getting data from shared weights now. Will update here,

jamesturner246:
Oh, one bug:  `group.push_state_to_device` still expects a population name.

neworderofjamie:
> I especially like being able to push and pull variables directly from groups, though. Looks less awkward / feels more encapsulated.
> 
Yeah, that had been bugging me every time I actually _used_ PyGeNN 😄 This is what happens when you start with an automatically-generated intermediate layer and then realise that it's actually quite useful in its own right...

> It might be cool to be able to initialise EGP in the add_population calls, rather than (forgetting to do) using set_extra_global_param calls. Similarly, it might be nice to include sparse connection setting with add_synapse_population. They seem pretty related.
Agreed - this would totally make sense - will have a go in another PR. In the short-term (until me or @chanokin bothers to update PyNN) we can probably remind you to pass these parameters via ``FutureWarning`` 😄 
> 
> Just getting data from shared weights now. Will update here,
👍

neworderofjamie:
@tnowotny, @jamesturner246 have finished this off and fixed the typos james found. Any chance of a quick review?

pygenn/genn_groups.py
```diff
@@ -44,6 +45,48 @@ def set_var(self, var_name, values):
...
+
+    def pull_var_from_device(self, var_name):
+        """Wrapper around GeNNModel.pull_var_from_device
+
+        Args:
+        var_name    --  string with the name of the variable
+        """
+        self._model.pull_var_from_device(self.name, var_name)
+
+    def pull_extra_global_param_to_device(self, egp_name, size=1):
```
jamesturner246, line 60: Nitpick: pull_e_g_p_ *from* _device?

neworderofjamie: Haha, looks like I copy-and-pasted [your typo](https://github.com/genn-team/genn/commit/4c440628b169203b10eac56b089c153208e5e80c#diff-608634e87bcf7f48cd21ee48051d988fR541-R547)

jamesturner246: whoops! xD

PR #325, opened 2020-05-05
Expose memory estimate to Python

### LINKED ISSUE #324: Expose memory estimates and merging stats properly
Currently these are printed out [here](https://github.com/genn-team/genn/blob/master/src/genn/genn/code_generator/generateAll.cc#L75-L98). As the comment suggests, this is not only a hack but also means they aren't really accessible from Python/SpineML. ``generateAll`` should instead return them in some sort of SWIGable struct.

neworderofjamie:
``GeNNModel.build`` now returns a ``MemAlloc`` object which lets you query GeNN's memory usage estimate:

```python
mem_alloc = model.build()
print("Device mem: %uMB" % mem_alloc.get_device_mbytes())
```

The following methods are also available:
```python
get_host_bytes()
get_device_bytes()
get_zero_copy_bytes()
get_host_mbytes()
get_device_mbytes()
get_zero_copy_mbytes()
```

In #324 I also talked about exposing the kernel merging stats but getting that out ended up being a bit of a mess and I think that's ok only being outputted to the log. As such fixes #324

### DIFF
--- a/generate_swig_interfaces.py
+++ b/generate_swig_interfaces.py
@@ -406,8 +406,9 @@ def generateBackend(swigPath, folder, namespace):
         # Add plog severity header so logging severities can be passed to backend
         mg.addSwigInclude('<plog/Severity.h>')
         
-        # Parse backend base, ignore BackendBase itself to get PreferencesBase definition
+        # Parse backend base, ignore BackendBase itself and MemAlloc to get PreferencesBase definition
         mg.addSwigIgnore("BackendBase")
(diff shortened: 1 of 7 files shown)


neworderofjamie:
Bit of both in this case - C++11 generally guarantees that returning complex objects won't actually result in a copy via some an optimization called [copy elision](https://zpjiang.me/2018/08/01/Copy-Elision-in-C/) and this was the route of least resistance to get this into Python

PR #326, opened 2020-05-05
Check if getSynapseDynamicsCode() is empty for KernelSynapseDynamicsUpdate

9inpachi:
Hi,

Removed repeating if statements.


neworderofjamie:
Haha, actually it looks like you found a bug. The second if should be checking whether ``getSynapseDynamicsCode()`` is empty rather than ``getLearnPostCode()``.

PR #329, opened 2020-05-14
Extra global parameters for variable initialisation

neworderofjamie:
Previously only connectivity initialisation snippets had extra global parameters. This has now been extended to variable initialisation snippets. This is generally helpful as it let you use extra global parameter arrays as little lookup tables:
```c++
class Decoder : public InitVarSnippet::Base
{
public:
    DECLARE_SNIPPET(Decoder, 0);

    SET_EXTRA_GLOBAL_PARAMS({{"rowWeights", "scalar*"}});
    SET_CODE("$(value) = $(id_pre) * $(rowWeights)[$(id_post)];\n")
};
IMPLEMENT_SNIPPET(Decoder);
```
These extra global parameters can then be set from C++ (this should be done between ``allocateMem`` and ``initialize`` calls), where the variable being set is called "g" and the synapse population is called "Syn":
```c++
allocaterowWeightsgSyn(4);
std::iota(&rowWeightsgSyn[0], &rowWeightsgSyn[4], 1);
pushrowWeightsgSynToDevice(4);
```
or Python (this should be done before ``GeNNModel.load`` call):
```python
syn.vars["g"].set_extra_global_init_param("rowWeights", np.arange(1, 5))
```

Discussions with @tnowotny and @jamesturner246 also made me realize that a dense matrix of procedurally-generated weights might be useful for some stuff so this P.R. also introduces this as ``DENSE_PROCEDURALG``. This should be super-quick as postsynaptic inputs get summed into a register. One caveat is that this can't currently be used with snippets that use random number generators as stream would have to be advanced by ``num_kernel_rng_calls * (next_spike - current_spike)`` which is a little tricky.


neworderofjamie:
Thanks - I think I'm *finally* getting into the test-writing habit! Big changes are mostly just because some of the bits of code had gone beyond acceptable levels of copy-and-paste and needed tidying tbh. Let's talk rngs on Tuesday!

PR #335, opened 2020-06-08
Another nasty Python corner case

neworderofjamie:
Another omission in my Python knowedge came to light when trying to figure out why PyNN GeNN tests were failing. Basically, my recent addition of cross-references to some classes resulted in a structure resembling:

```python
class A(object):
    def __init__(self, parent):
        print("Creating A")
        self.parent = parent

    def __del__(self):
        print("Destroying A")

class B(object):
    def __init__(self):
        self.children = []
        print("Creating B")

    def __del__(self):
        print("Destroying B")

b = B()
b.children.append(A(b))
b.children.append(A(b))
b = None
```
Which, suprisingly (at least to me as a C++ programmer), does not result in any objects being destroyed. This is because, unlike e.g. Java, Python doesn't automatically run a 'proper' garbage collector and relies with simple refernce counting which can't deal with such loops. The solution is to redefine A as:

```python
from weakref import proxy

class A(object):
    def __init__(self, parent):
        print("Creating A")
        self.parent = proxy(parent)

    def __del__(self):
        print("Destroying A")
```

Which I have done where required in PyGeNN.

### DIFF
--- a/pygenn/genn_groups.py
+++ b/pygenn/genn_groups.py
@@ -7,6 +7,7 @@ try:
 except NameError:  # Python 3
     xrange = range
 
+from weakref import proxy
 from deprecated import deprecated
 from six import iteritems
 import numpy as np
@@ -32,7 +33,7 @@ class Group(object):
         name    --  string name of the Group
         """
         self.name = name
-        self._model = model
+        self._model = proxy(model)
         self.vars = {}
         self.extra_global_params = {}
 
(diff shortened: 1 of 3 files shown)


neworderofjamie:
My thoughts exactly!

tnowotny:
Hum ... does that mean, on the other hand, that a.parent may be pointing to nowhere, once b is deleted in the following scenario:
```
b= B()
a= A(b)
b= None
```
Is that a problem?

neworderofjamie:
So what happens is that a ``ReferenceError`` gets thrown which, as this isn't the intentional usage of the (contrived) code and as python doesn't have ``private`` etc to prevent the programmer from doing things like that, is about as a good a result as you can hope for 😄

PR #339, opened 2020-06-26
get connectivity from device for on-device init

chanokin:
changes to allow required data to be accessed after on-device connectivity initialization


pygenn/genn_groups.py
```diff
@@ -774,11 +786,15 @@ def load(self):
                         # Copy row from non-padded indices into correct location
                         ind[i:i + r] = self.ind[syn:syn + r]
                         syn += r
+                elif self.connectivity_initialiser is not None:
+                    self.ind = ind
```
neworderofjamie, line 790: I think using ``self.ind`` and ``self.row_lengths`` for both data set from ``set_sparse_connections`` and for views of GeNN data is a bit dodgy - maybe add ``self._ind`` and ``self._row_lengths`` for the GeNN views and **always** set them in ``load`` (this also maintains backward compatibility)

pygenn/genn_groups.py
```diff
@@ -632,6 +632,18 @@ def get_sparse_post_inds(self):
...
 
+            if self.connectivity_initialiser is not None:
+                # download the connectivity details from device
+                self.pull_connectivity_from_device()
+
+                # the ind array still has some non-valid data so we remove them
+                # with the row_lengths data
+                ind = [self.ind[start_idx + j]
+                        for i, start_idx in enumerate(xrange(0, len(self.ind), self.max_row_length))
+                            for j in range(self.row_lengths[i])]
```
neworderofjamie, line 643: Doing this in python will be mega slow and memory-eating as these arrays can be gigabytes. Maybe try some numpy like:

```python
return np.hstack([self._ind[i * self._max_row_length, (i * self._max_row_length) + r] 
                  for i, r in enumerate(self._row_lengths)])
```

pygenn/genn_groups.py
```diff
@@ -747,19 +759,19 @@ def push_connectivity_to_device(self):
...
+        if not self.is_dense and self.weight_sharing_master is None:
+            if self.is_ragged:
+                # Get pointers to ragged data structure members
+                ind = self._assign_ext_ptr_array("ind",
+                                                 self.weight_update_var_size,
+                                                 "unsigned int")
+                row_length = self._assign_ext_ptr_array("rowLength",
+                                                        self.src.size,
+                                                        "unsigned int")
+                # If data is available
```
neworderofjamie, line 773: you should also get the max row length GeNN is using with:
```python
max_row_length = self._assign_ext_ptr_single("maxRowLength", "unsigned int")
```
Otherwise, you won't be able to read back variables or connectivity correctly from synapse groups with on-device connectivity

pygenn/genn_groups.py
```diff
@@ -632,6 +632,18 @@ def get_sparse_post_inds(self):
                 raise Exception("only manually initialised connectivity "
                                 "can currently by accessed")
 
+            if self.connectivity_initialiser is not None:
+                # download the connectivity details from device
+                self.pull_connectivity_from_device()
```
neworderofjamie, line 637: The rest of the API assumes that the user knows when to call "pull" functions so this shouldn't be done automatically in this one case.

chanokin: so this should be called from the PyNN side?

pygenn/genn_groups.py
chanokin: does this actually change when something is computed on-device?

pygenn/genn_groups.py
```diff
@@ -604,15 +611,22 @@ def get_sparse_pre_inds(self):
...
-            if self.ind is None or self.row_lengths is None:
-                raise Exception("only manually initialised connectivity "
-                                "can currently by accessed")
-
-            # Expand row lengths into full array
-            # of presynaptic indices and return
-            return np.hstack([np.repeat(i, l)
-                              for i, l in enumerate(self.row_lengths)])
-
+            if self.connectivity_initialiser is None:
```
neworderofjamie, line 614: you could probably reduce the code duplication here with something like:

```python
rl = self.row_lengths if self.connectivity_initialiser is None else self._row_lengths
               
if rl is None:
    raise Exception("Cannot access connectivity")
...
```

pygenn/genn_groups.py
neworderofjamie: Actually, forget that, I was totally wrong. ``pop.get_max_row_length()`` is totally fine to use however the connectivity is calculated.

pygenn/genn_groups.py
neworderofjamie: Exactly! That also has the advantage that you only need to make that (potentially expensive) call once if you, for instance, want both the pre and post indices

chanokin:
Any other observations to the code?

neworderofjamie:
I think the only outstanding thing (which is totally my fault) is that the ``self._max_row_length`` stuff is uneccessary - you can just call ``pop.get_max_row_length()`` however the connectivity was configured.

chanokin:
That's already in the last commit :smiley:

neworderofjamie:
Are you sure you pushed, I see:

- https://github.com/genn-team/genn/blob/pynn_sparse_init_on_device/pygenn/genn_groups.py#L789
- https://github.com/genn-team/genn/blob/pynn_sparse_init_on_device/pygenn/genn_groups.py#L502
- https://github.com/genn-team/genn/blob/pynn_sparse_init_on_device/pygenn/genn_groups.py#L659

chanokin:
forgot to push the button :man_facepalming:

neworderofjamie:
Hate it when that happens 😄 merging!

PR #340, opened 2020-07-02
Open CL support

neworderofjamie:
Here it is! @9inpachi and colleagues got to the point of a working GeNN 4.2 OpenCL backend and I have updated it to support GeNN 4.3.0 i.e. merged kernels. It now supports pretty much all GeNN features. Significant changes to GeNN itself are:

1. GeNN now has a /share directory (I believe this is a standard Unix convention) which contains resources that can get copied into generated code. We use it here to copy in [clRNG](https://github.com/clMathLibraries/clRNG) (what seems like the most viable alternative to cuRAND) and Khronos-supplied C++ headers which, for magical reasons of their own, Apple don't include in their frameworks.
2. OpenCL 1.2 does not have the equivalent of Unified Virtual Addressing (CUDA 4 feature) so device pointers are innaccesible to host code (device memory is handled indirectly via ``cl::Buffer``). This means some extra flexibility is required in the generation of merged structures but the resulting refactor has, I think, improved the nastiness of ``GroupMerged`` and ``MergedStructGenerator``
3. Slightly changed Jenkins testing strategy - stopped testing on the master node (i.e. the poor web-serving VM itself) and instead test all supported backends on all nodes. Also added some hacky code to all tests to allow OpenCL device to be selected using environment variables (On Macs with discrete GPUs, the CPU will otherwise be selected and the driver is super-limited)
4. In various bits of code we were using references but OpenCL is C-based so have switched over to using pointers (no functional difference)
5. While OpenCL is C-based it doesn't support maths functions with explicit single or double-precision, instead it has some weird special case operator-overloading. Now (as of _some_ time ago) CUDA is C++ there's also no need to use the explicit single-precision versions anywhere. Have repurposed that code path to get rid off any lingering single-precision functions from code snippets and replace them with the generic versions.
6. At the algorithmic level, OpenCL and CUDA (and presumably any other SIMT abstraction) are the same so I have added ``BackendSIMT`` between ``BackendBase`` and the CUDA and OpenCL backends. This and accompanying ``PresynapticUpdateStrategySIMT`` provide algorithms for both backends and the backends themselves just deal with setting up kernels, allocating memory etc. 
7. Now we have an AMD GPU to play with I have been able to isolate the cause of the issues we've been experiencing. Essentially a combination of over-zealous optimization and bugs in the AMD runtime. I have implemented a rather brute-force workaround by inserting flush events between all kernel launches (approximately equivalent to ``cudaStreamSynchronize()``) but AMD have confirmed the bug and are investigating (https://community.amd.com/message/2995651) so hopefully these can be removed after a driver update.

Before this is merged:
- [ ] Documentation/credits update

To address in future PRs (issue links in brackets):
- 'Zero-copy' is probably key to getting decent performance out of devices with shared memory (Intel GPUs) (#369).
- We're currently including the kernels as big strings. This is good for debugging but slows down model loading. OpenCL can also 'compile' kernels to produce binary blobs. These could either be compiled by the backend and written into the source code or built the first time the model is run (#366).
- I have made some effort at optimizing performance of OpenCL but it still lags behind CUDA on the same hardware. I am suspicious that this is due to something sub-optimal happening in clRNG (as there's little else magical that happens in neuron kernels) (#370).
- More useful device and platform picking algorithms in optimiser e.g. favour discrete device (#365)
- On some devices (ARM) local memory isn't really local so using it to cache e.g. neuron parameters doesn't help (#368)
- OpenCL does not support 'fire-and-forget' floating point add which is critical to good GeNN performance in a lot of models. On NVIDIA cards, we use inline ptx to generate the correct instruction but non-NVIDIA hardware does not support this operation. However, converting the inSyn calculation to integer would be relatively simple and should provide a massive boost (#367).
- Extremely irritatingly, as it would be really useful in the absence of modern NVIDIA-equipped Macs, OpenCL is irredeemably broken on Mac OS. On AMD devices [this bug](https://feedbackassistant.apple.com/feedback/8250497) prevents our code from working at all and on NVIDIA everything seems to work aside from clEnqueueFillBuffer which [randomly fails](https://stackoverflow.com/questions/38556710/clenqueuefillbuffer-fills-a-buffer-correctly-only-at-random/44402248). Spending more time on this seems seems like a waste of my time.


9inpachi:
Hi Jamie!

This looks wonderful!
I will most probably look into support code this weekend once I get off of some work. :)

Peace.

neworderofjamie:
Thank you @9inpachi ! One advantage of getting this merged sooner rather than later is that if you feel like continuing to contribute to the project you can work in smaller pull requests off master which is a bit nicer.

9inpachi:
@neworderofjamie Sure. I will try testing it out on the machine and let you know.

9inpachi (commented):
This looks nice and clean.

9inpachi:
Wonderful!

@neworderofjamie Thanks for following up on this from the 4.2.0 version!
You improved the code so much along with implementing the newer features over time.

Extremely glad to have taken up and worked on this project. :)

Hoping you and GeNN best for the future! Let me know if you need any help anytime.

neworderofjamie:
Thank you for all your help with this @9inpachi - this definitely wouldn't have happened without you guys! If you ever fancy some smaller OpenCL-related GeNN tasks, I've turned a number of potential improvements into issues.

9inpachi:
Great - the recorded issues will help. I will try contributing whenever I get some free time.

PR #344, opened 2020-07-07
Device only connectivity fix

neworderofjamie:
Missed an issue with #339 - PyNN doesn't do this but, you can mark on-device initialised connectivity (like variables) as being device only if you _never_ want to download  connectivity, potentially letting you save loads of host memory. However, this means the variables we were querying don't exist!


chanokin:
This would only apply for PyGeNN stuff, right? All my tests ran fine after merging this so it seems fine :)

PR #345, opened 2020-07-13
Generation of support code for backends that do not support namespaces

9inpachi:
Hi,

This adds generation of support code for backends that do not support namespaces. It works by substituting functions with a namespace prefix in the function name.
Also added test for the new function in `codeGenUtils.cc` and added generation of host buffers (`h_`) for extra global parameters.

### DIFF
--- a/include/genn/genn/code_generator/codeGenUtils.h
+++ b/include/genn/genn/code_generator/codeGenUtils.h
@@ -85,7 +85,7 @@ GENN_EXPORT void checkUnreplacedVariables(const std::string &code, const std::st
 /*! \brief 
  */
  //--------------------------------------------------------------------------
-GENN_EXPORT std::string getNamespaceFunction(const std::string supportCode, const std::string code, std::string namespaceName);
+GENN_EXPORT std::string substituteNamespaceFunction(const std::string supportCode, const std::string code, std::string namespaceName);
 
 //-------------------------------------------------------------------------
 /*!
--- a/include/genn/genn/code_generator/codeGenUtils.h
+++ b/include/genn/genn/code_generator/codeGenUtils.h
@@ -82,7 +82,7 @@ GENN_EXPORT std::string ensureFtype(const std::string &oldcode, const std::strin
 GENN_EXPORT void checkUnreplacedVariables(const std::string &code, const std::string &codeName);
 
 //--------------------------------------------------------------------------
-/*! \brief 
+/*! \brief This function substitutes function names in a code with namespace as prefix of the function name for backends that do not support namespaces by checking that the function indeed exists in the support code and returns the substituted code.
  */
  //--------------------------------------------------------------------------
 GENN_EXPORT std::string substituteNamespaceFunction(const std::string supportCode, const std::string code, std::string namespaceName);
--- a/include/genn/genn/code_generator/codeGenUtils.h
+++ b/include/genn/genn/code_generator/codeGenUtils.h
@@ -85,7 +85,7 @@ GENN_EXPORT void checkUnreplacedVariables(const std::string &code, const std::st
 /*! \brief 
  */
  //--------------------------------------------------------------------------
-GENN_EXPORT std::string getNamespaceFunction(std::string supportCode, const std::string code, std::string namespaceName);
+GENN_EXPORT std::string getNamespaceFunction(const std::string supportCode, const std::string code, std::string namespaceName);
 
 //-------------------------------------------------------------------------
 /*!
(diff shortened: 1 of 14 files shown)


include/genn/genn/code_generator/codeGenUtils.h
```diff
@@ -80,6 +81,12 @@ GENN_EXPORT std::string ensureFtype(const std::string &oldcode, const std::strin
 //--------------------------------------------------------------------------
 GENN_EXPORT void checkUnreplacedVariables(const std::string &code, const std::string &codeName);
 
+//--------------------------------------------------------------------------
+/*! \brief This function substitutes function names in a code with namespace as prefix of the function name for backends that do not support namespaces by checking that the function indeed exists in the support code and returns the substituted code.
+ */
+ //--------------------------------------------------------------------------
+GENN_EXPORT std::string substituteNamespaceFunction(const std::string supportCode, const std::string code, std::string namespaceName);
+
```
neworderofjamie, line 89: For consistency with other functions like this in GeNN, I think:

```c++
GENN_EXPORT void disambiguateNamespaceFunction(std::string &supportCode, const std::string &namespaceName);
```

might be a better signature and name.

src/genn/backends/opencl/presynapticUpdateStrategy.cc
```diff
@@ -125,6 +121,10 @@ void PreSpan::genCode(CodeStream &os, const ModelSpecMerged &modelMerged, const
 
             std::string code = threshOsStream.str();
 
+            if (!wu->getSimSupportCode().empty()) {
```
neworderofjamie, line 124: I think this belongs in [this function](https://github.com/genn-team/genn/blob/opencl/src/genn/genn/code_generator/generateSynapseUpdate.cc#L113) like the neuron stuff (you might need to add the namespace as a parameter or something).

src/genn/genn/code_generator/codeGenUtils.cc
```diff
@@ -486,4 +486,33 @@ void checkUnreplacedVariables(const std::string &code, const std::string &codeNa
...
     }
 }
+
+//--------------------------------------------------------------------------
+/*! \brief This function substitutes function names in a code with namespace as prefix of the function name for backends that do not support namespaces by checking that the function indeed exists in the support code and returns the substituted code.
+ */
+ //--------------------------------------------------------------------------
+std::string substituteNamespaceFunction(const std::string supportCode, const std::string code, std::string namespaceName) {
+    // Regex for function call
+    std::regex funcCallRegex("\\w+(?=\\(.*\\))");
```
neworderofjamie, line 496: In C++ 11 you can make these slightly nicer by using 'raw' strings i.e. 
```c++
std::regex funcCallRegex(R"\w+(?=\(.*\))");
```
Also some comments explaining that the regexes do would be nice as they can be tricky to decode 😄 (e.g. I'm struggling to remember how forward lookeahead ``?=`` means)

src/genn/genn/code_generator/generateSynapseUpdate.cc
```diff
@@ -2,6 +2,7 @@
 
 // Standard C++ includes
 #include <string>
+#include <regex>
```
neworderofjamie, line 5: it doesn't look like you're not actually using ``<regex>`` in this file so, if it's required, that suggests you need it include in "codeGenUtils.h" instead.

src/genn/backends/opencl/presynapticUpdateStrategy.cc
9inpachi: Do you mean we should just substitute the same way but in `applySynapseSubstitutions()` before the `synapseSubs.apply(code);` call so we don't repeat the code here in `presynapticUpdateStrategy`?

If so, then I didn't do that because the backend variable is not available there to check if the backend does not support namespace. Although we can pass the backend itself or a boolean (`backend.supportsNamepsace()`) to `applySynapseSubstitutions()` from [here](https://github.com/genn-team/genn/blob/opencl/src/genn/genn/code_generator/generateSynapseUpdate.cc#L172).

Is that it?

neworderofjamie: Yeah, exactly. Feel free to change that function however seems best.

9inpachi:
> I think you still need to update Backend::genExtraGlobalParamAllocation to deal with host buffers

I tried but it seems the groups are not set up for host buffers. Because even if we generate the host buffer allocation for groups - there is no actual definition of them in the program somehow (should be generated from [here](https://github.com/9inpachi/genn/blob/opencl/src/genn/backends/opencl/backend.cc#L1868)) so they the code ends up not working.

May be you can take a look since I am not too familiar with this new group stuff. :)

Issue #289, opened 2020-01-03
PyGeNN only really works with precision set to float


neworderofjamie:
This is almost entirely due to laziness and just requires the scalar type 'threading' through various bits of code

neworderofjamie:
Feature tests now test PyGeNN with "double" and "float"

PR #356, opened 2020-08-19
PyGeNN time precision

neworderofjamie:
By default GeNN represents time using the standard "scalar" type which is often 32-bit float. However, as this demonstrates:

```python
import numpy as np
d = np.asarray([2260350.0],dtype=np.float32)
print(d[0])
d[0]+=0.1
print(d[0])
```

Float stops being able to represent the timestep and the current time in long simulations. GeNN has supported seperate time precision since #207 but it's never made its way into PyGeNN. When you create a ``GeNNModel`` you can now control the time precision like this:

```python
model = genn_model.GeNNModel("float", "va_benchmark", time_precision="double")
```
A quick review and test this doesn't break PyNN or TensorGeNN is exciting ways would be much appreciated!

Additionally, this also goes some way to helping with #289 but more testing is required.

### DIFF
--- a/pygenn/genn_groups.py
+++ b/pygenn/genn_groups.py
@@ -13,7 +13,7 @@ from six import iteritems
 import numpy as np
 from . import genn_wrapper
 from . import model_preprocessor
-from .model_preprocessor import ExtraGlobalVariable, Variable, genn_types
+from .model_preprocessor import ExtraGlobalVariable, Variable
 from .genn_wrapper import (SynapseMatrixConnectivity_SPARSE,
                            SynapseMatrixConnectivity_BITMASK,
                            SynapseMatrixConnectivity_DENSE,
@@ -109,7 +109,7 @@ class Group(object):
 
         assert param_type is not None
         egp_dict[param_name] = ExtraGlobalVariable(param_name, param_type,
-                                                   param_values)
+                                                   param_values, self)
 
     def _assign_ext_ptr_array(self, var_name, var_size, var_type):
         """Assign a variable to an external numpy array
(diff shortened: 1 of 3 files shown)


pygenn/model_preprocessor.py
```diff
@@ -312,7 +289,8 @@ def __init__(self, variable_name, variable_type, values=None):
         else:
             self.is_scalar = True
             self.type = variable_type
-
+        
+        self.group = proxy(group)
```
chanokin, line 293: what does proxy do?

neworderofjamie: it's basically a weak reference to prevent circular-dependencies

neworderofjamie: https://docs.python.org/3/library/weakref.html

chanokin: aaah, I remember you fixing other stuff with this :)

pygenn/genn_groups.py
```diff
@@ -109,7 +109,7 @@ def _set_extra_global_param(self, param_name, param_values, model, egp_dict=None
 
         assert param_type is not None
         egp_dict[param_name] = ExtraGlobalVariable(param_name, param_type,
-                                                   param_values)
+                                                   param_values, self)
```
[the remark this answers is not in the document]
neworderofjamie: Hah, my GSoC student found the same classic typo at the same time - this is fixed now right?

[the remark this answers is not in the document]
chanokin: yup, that's the thing fixed on the next commit

PR #357, opened 2020-08-20
RNG substream fix

neworderofjamie:
This is actually a really nasty one that shows up in a couple of places only when synapse groups are merged together. 

1. Each presynaptically parallelised procedural connectivity thread needs to select the correct RNG substream - these should be calculated by adding the presynaptic neuron index to the starting thread index of the merged group but was instead using the starting thread index of the whole block of merged groups. This meant if you had multiple projections between populations with the same number of neurons which got merged together, their connectivity would be exactly the same.
2. There's some slightly involved logic in the optimised bitmask connectivity implementation (#280) where the writing back of synaptic input from shared memory also used the starting thread index.

The code that generates the merged group binary search now also adds "group_start_id" to the substitution stack so this can be accessed in both cases. I have also added new feature tests to check neither of these bugs recur.


tnowotny (approved):
For what it's worth, this seems to be simplifying things while correcting the error; Hard to see on a quick scan what's right or wrong but looks sensible in principle.

neworderofjamie:
Yeah, I removed the possibility for further errors by not passing through the problematic parameter 😄

PR #361, opened 2020-09-25
Fix for old CUDA versions and impove warning suppression

neworderofjamie:
Slightly embarassingly, this came to light when @agarwalmanvi was trying to show her supervisor GeNN. 

Nonetheless, the undocumented nvcc option I was using to hide annoying warnings when optimizing block sizes aren't implemented in CUDA 7.5 or 8 and cause nvcc to keel over with a fatal warning. This change only tries to hide these warnings on CUDA 9 and later. Also, after realising the magic numbers I was using to repress these warnings changed not only between platforms but between minor CUDA versions and are thus entirely unmaintainable, I figured out the string equivalent by dissambling the ``cudafe`` executable and grepping at the strings section.

Also, after testing this on the K40 still somehow hanging in on apollo, I can confirm that GeNN still works with GCC 4.9.4 and CUDA 7.5 😄

### DIFF
--- a/src/genn/backends/cuda/backend.cc
+++ b/src/genn/backends/cuda/backend.cc
@@ -2040,7 +2040,13 @@ std::string Backend::getNVCCFlags() const
 #ifdef _WIN32
     nvccFlags += " -Xcudafe \"--diag_suppress=2961\"";
 #else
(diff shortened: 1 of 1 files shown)


tnowotny (approved):
Good.

PR #364, opened 2020-09-28
Added support for SM 8.0 and 8.6

neworderofjamie:


tnowotny (approved):
Ok ... the world keeps turning, and so we keep updating. Thanks for staying on top.

PR #372, opened 2020-10-12
Efficient spike/spike-like event recording

neworderofjamie:
The standard GeNN idiom for recording spikes was to use synchronous ``cudaMemcpy`` to read the number of spikes for each population in the current timestep, then use another synchronous ``cudaMemcpy`` to read that many spikes and then copy the host data into a host-side data structure. Especially with simulation with a 0.1ms timestep and lots of populations, this pretty much prevents any chance of real-time. This PR introduces very simple system which lets you allocate any remaining GPU memory for spike recording meaning that, in many simulations, you can run for a large number of timesteps without any device->host memory transfers.

1. Mark a neuron population for spike recording in C++:
    ```c++
    e->setSpikeRecordingEnabled(true);
    ```
    or Python:
    ```python
    excitatory_pop.spike_recording_enabled = True
    ```
2. Allocate a number of timesteps of spike recording buffer in C++:
    ```c++
    allocateRecordingBuffers(10000);
    ```
    or Python:
    ```python
    model.load(num_recording_timesteps=10000)
    ```
3. Pull the spike recording buffers in C++:
    ```c++
    pullRecordingBuffersFromDevice();
    ```
    or Python:
    ```python
    model.pull_recording_buffers_from_device()
    ```

4. Access recording data in C++ (where this helper function is in userprojects - you can do whatever you want with the data):
    ```c++
    writeTextSpikeRecording("spikes.csv", recordSpkE,
                                            Parameters::numOutput, 10000, 0.1,
                                            ",", true);
    ```
    or Python:
    ```python
    spike_times, spike_ids = excitatory_pop.spike_recording_data
    ```

I think recording spikes is the most common and most inefficient case but this could easily be extended in future to record subsets of neurons/state variables..

From the upcoming PyGeNN paper, this disproportionately helps Python and slow CPUs where the copying of data and sticking it in a host-side data structure component is particularly costly:
microcircuit_overheads


jamesturner246:
I'm getting 
```
In file included from /its/home/jt273/workspace/genn-recording/pygenn/genn_wrapper/include/../../../userproject/include/sharedLibraryModel.h:24:0,
                 from /its/home/jt273/workspace/genn-recording/pygenn/genn_wrapper/include/sharedLibraryModelNumpy.h:4,
                 from pygenn/genn_wrapper/generated/SharedLibraryModelNumpy_wrap.cpp:3132:
/its/home/jt273/workspace/genn-recording/pygenn/genn_wrapper/include/../../../userproject/include/spikeRecorder.h: In function ‘void writeTextSpikeRecording(const string&, const uint32_t*, unsigned int, unsigned int, double, const string&, bool, bool)’:
/its/home/jt273/workspace/genn-recording/pygenn/genn_wrapper/include/../../../userproject/include/spikeRecorder.h:223:43: error: invalid conversion from ‘int’ to ‘std::ios_base::openmode {aka std::_Ios_Openmode}’ [-fpermissive]
     std::ofstream stream(filename, append ? std::ofstream::app : 0);
                                    ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~
In file included from /its/home/jt273/workspace/genn-recording/pygenn/genn_wrapper/include/../../../userproject/include/spikeRecorder.h:5:0,
                 from /its/home/jt273/workspace/genn-recording/pygenn/genn_wrapper/include/../../../userproject/include/sharedLibraryModel.h:24,
                 from /its/home/jt273/workspace/genn-recording/pygenn/genn_wrapper/include/sharedLibraryModelNumpy.h:4,
                 from pygenn/genn_wrapper/generated/SharedLibraryModelNumpy_wrap.cpp:3132:
/usr/include/c++/7/fstream:715:7: note:   initializing argument 2 of ‘std::basic_ofstream<_CharT, _Traits>::basic_ofstream(const string&, std::ios_base::openmode) [with _CharT = char; _Traits = std::char_traits<char>; std::__cxx11::string = std::__cxx11::basic_string<char>; std::ios_base::openmode = std::_Ios_Openmode]’
       basic_ofstream(const std::string& __s,
       ^~~~~~~~~~~~~~
```

when compiling PyGeNN. Isn't openmode a c-string?

neworderofjamie:
Well that last minute addition was clearly not great 😄 can you get latest and try?

pygenn/genn_groups.py
```diff
@@ -278,6 +300,14 @@ def delay_slots(self):
     def size(self):
         return self.pop.get_num_neurons()
```
jamesturner246, line 302: Would be cool to have a per-layer `pull_recording_buffers_from_device` method, so you can selectively pull recording buffers without pulling anything else.

jamesturner246 (commented):
Okay, it compiles and the results look sensible. Thought they looked short, but I wasn't setting save steps high enough.

API is fine, but maybe convenient for lazy people to set layer recording in the neuron population constructor? Also see the per-neuron push/pull recording buffer note.

I see you plan to do it for neuron state vars (and synapse state?) as well, which could be handy.

pygenn/genn_groups.py
neworderofjamie: What would be the use case for doing that? If you've enabled recording on a neuron population wouldn't you always want the data at the end of the recording period?

jamesturner246: Good point, I suppose you would.

neworderofjamie:
Turning over the metaphorical rock by adding tests uncovered a number of loose ends/bugs (including what appears to be a bug in NVIDIA's OpenCL implementation which needs further investigation). These are all now addressed and this should be good to go.

tnowotny (approved):
Did I miss it or should we add a few words about this to the manual? - otherwise approved.

PR #373, opened 2020-10-13
Kernel-based connectivity

neworderofjamie:
After several abortive attempts, this PR slightly extends existing syntax to enable ``InitSparseConnectivitySnippet`` snippets to be used to initialize kernel-based connectivity like convolutions (see ``InitSparseConnectivitySnippet::Conv2D`` for an example):

1. The ``SET_CALC_KERNEL_SIZE_FUNC`` macro in a C++ snippet definition / the ``calc_kernel_size_func`` kwarg in a Python snippet creation call let you define a function which calculate the kernel size (with arbitrary dimensions) from the model parameters using similar syntax to maximum row length calculation etc.
2. Snippets which define a kernel size extend the ``$(addSynapse, idPost)`` function substitution to take corresponding kernel indices e.g.  ``$(addSynapse, idPost, kernRow, kernCol, $(inChan), outChan)``.

GeNN then turns these indices into a flat index which is accessible to ``InitVarSnippet`` used to initialise per-synapse variables as ``$(id_kernel)`` (see ``InitVarSnippet::Kernel``). This whole system works both with ``SPARSE_INDIVIDUALG`` connections where it can initialise sparse matrices with convolutional connectivity on the GPU or with ``PROCEDURAL_PROCEDURALG`` to do the whole thing procedurally. The only additional complexity is that, previously, per-synapse variables in ``SPARSE_INDIVIDUALG`` connections were initialised in the ``initializeSparse`` kernel but, so that ``$(id_kernel)`` is still in existence, any variable initialisation that requires it must be done at the same time as the synapses are initialised (hence the refactor of my perpetual bugbear the ``GroupMerged`` class hierarchy).

I have added tests for both modes of operation which, in case you're interested in what the magic ``conv_data.h`` contains, basically use Conv2D to apply different Sobel operations to an input image.

Finally, in the process of testing this, I found a bug which I think has been lurking for many years. When you use presynaptic parallelism with < block size postsynaptic neurons, shared memory was used to accumulate postsynaptic input. The contents of the shared memory was then added to global memory without using atomic operations so was only correct when there was also only < block size presynaptic neurons as well (the case that the tests cover). Could have added global memory atomics here as well but the whole "optimisation" seemed dubious as all the threads in the block are likely to be contending for the same shared memory and then all blocks will be contending for the same global memory so I simplified and removed this code path entirely for presynaptic parallelism.


jamesturner246 (approved):
Looks good!

tnowotny (approved):
This seems to make sense overall even though things are getting increasingly hard to assess ...

PR #374, opened 2020-10-22
Column-parallel connectivity initialisation

neworderofjamie:
One annoying limitation of the row-parallel connectivity initialisation is that it can't be used to implement "FixedNumberPre" (PyNN-speak) or "fixed_indegree" (NEST-speak) connectivity where each postsynaptic neuron has a fixed number of synapses connecting it to presynaptic neurons. This is widely used and nice as you can scale up a network without changing weights and each neuron will receive similar input. There's no obvious way to implement such a scheme for procedural connectivity but, for initialisation, you can simply use a thread per postsynaptic neuron and add entries to the matrix being careful to use atomics to update the row length so you don't overwrite a synapse another thread is mid-way through adding. Column-parallel connectivity initialisation code can now be added with the ``SET_COL_BUILD_CODE`` macro in C++ and ``row_build_code`` kwarg to the Python ``create_custom_sparse_connect_init_snippet_class`` function. 

Only slightly gnarly bit is the additional complexity caused by merging this with #373 which has led to a bit of a re-write of some code in the SIMT and single-threaded CPU backends.


tnowotny (approved):
Also looks good. Are we keeping track of all these improvements in the documentation? I know it's additional pain on top of having to do tests and all but it's key if people are to use it ...

PR #376, opened 2020-10-30
Spike time improvement

### LINKED ISSUE #314: Behaviour of $(sT_pre) and $(sT_post) is not ideal

neworderofjamie:
As discussed in #314, in some event-driven learning rules, you want to be able to calculate ``$(t) - $(sT_pre)`` in the presynaptic spike-driven sim_code (and  ``$(t) - $(sT_post)`` in postsynaptic spike-driven code). However, the existing spike times got set in the neuron kernel so actually contain the time of the _current_ spike being processed in the sim_code/postsynaptic code i.e. it'll always be ``$(t) - DT``. Having access to the _previous_ spike times would also be very useful. Implementing this has two parts:

1. Setting spike times _after_ synapse kernels run. Because there's already a kernel run before the neuron update to reset spike counts etc, I've extended this to also update last spike times based on the contents of the last timestep's ``glbSpk`` and ``glbSpkCnt``. Because there's a kernel here already, this shouldn't cause extra latency and as previously this kernel only had a thread per _population_ there were likely to be a lot of hardware sitting idle.
2. If delays are involved, the fact that the spike times are updated after the synapses during the timestep they are emitted in still leaves them in the 'wrong' delay slot if synapse code reads them during subsequent timesteps. To handle this case, synapse code needs to read spike times from the delay slot _before_ other variables.

In C++ these new 'previous spike time' variables can be generated by adding ``SET_NEEDS_PREV_PRE_SPIKE_TIME(true)`` and ``SET_NEEDS_PREV_POST_SPIKE_TIME(true)`` macros to a weight update model and, in Python, by setting the ``is_prev_pre_spike_time_required=True`` and ``is_prev_post_spike_time_required=True`` kwargs. As an example of the usage of these new variables, you can now implement the fusi rule as:
```python
fusi_model = create_custom_weight_update_class(
    "fusi_model",
    param_names=["tauC", "a", "b", "thetaV", "thetaLUp", "thetaLDown", "thetaHUp", "thetaHDown",
                 "thetaX", "alpha", "beta", "Xmax", "Xmin", "JC", "Jplus", "Jminus"],
    var_name_types=[("X", "scalar")],
    post_var_name_types=[("C", "scalar")],
    sim_code="""
    $(addToInSyn, (($(X) > $(thetaX)) ? $(Jplus) : $(Jminus)));
    const scalar decayC = $(C) * exp(-dt / $(tauC));
    if ($(V_post) > $(thetaV) && $(thetaLUp) < decayC && decayC < $(thetaHUp)) {
        $(X) += $(a);
    }
    else if ($(V_post) <= $(thetaV) && $(thetaLDown) < decayC && decayC < $(thetaHDown)) {
        $(X) -= $(b);
    }
    else {
        const scalar X_dt = $(t) - $(prev_sT_pre);
        if ($(X) > $(thetaX)) {
            $(X) += $(alpha) * X_dt;
        }
        else {
            $(X) -= $(beta) * X_dt;
        }
    }
    $(X) = fmin($(Xmax), fmax($(Xmin), $(X)));
    """,
    post_spike_code="""
    $(C) = ($(C) * exp(-dt / $(tauC))) + $(JC);
    """,
    is_prev_pre_spike_time_required=True,
    is_post_spike_time_required=True)
```
which halves the memory usage of our previous effort:
```
fusi_model = create_custom_weight_update_class(
    "fusi_model",
    param_names=["tauC", "a", "b", "thetaV", "thetaLUp", "thetaLDown", "thetaHUp", "thetaHDown",
                 "thetaX", "alpha", "beta", "Xmax", "Xmin", "JC", "Jplus", "Jminus"],
    var_name_types=[("X", "scalar"), ("last_tpre", "scalar")],
    post_var_name_types=[("C", "scalar")],
    sim_code="""
    $(addToInSyn, (($(X) > $(thetaX)) ? $(Jplus) : $(Jminus)));
    const scalar decayC = $(C) * exp(-dt / $(tauC));
    if ($(V_post) > $(thetaV) && $(thetaLUp) < decayC && decayC < $(thetaHUp)) {
        $(X) += $(a);
    }
    else if ($(V_post) <= $(thetaV) && $(thetaLDown) < decayC && decayC < $(thetaHDown)) {
        $(X) -= $(b);
    }
    else {
        const scalar X_dt = $(t) - $(last_tpre);
        if ($(X) > $(thetaX)) {
            $(X) += $(alpha) * X_dt;
        }
        else {
            $(X) -= $(beta) * X_dt;
        }
    }
    $(X) = fmin($(Xmax), fmax($(Xmin), $(X)));
    $(last_tpre) = $(t);
    """,
    post_spike_code="""
    $(C) = ($(C) * exp(-dt / $(tauC))) + $(JC);
    """,
    is_pre_spike_time_required=True,
    is_post_spike_time_required=True)
```
Fixes #314


tnowotny (approved):
Looks good. A nice improvement and not disrupting anything existing (unless there are unintended side-effects of setting the spike times at a different time in the update cycle that we did not spot - but not too nervous about that).

PR #378, opened 2020-11-25
Allocate mem params

neworderofjamie:
This is a really simple one which enables multiple instances of the same executable to be run on multiple GPUs
* Adds new generic mechanism to allow backends to specify parameters to pass via ``allocateMem``
* Add a new ``DeviceSelectMethod`` to the CUDA backend which uses this mechanism to allow devices to be specified via ``allocateMem`` parameter


tnowotny (approved):
Yeah ... I think that looks reasonable and reasonably tidy.

Issue #379, opened 2020-12-01
Repeat of spike-like-event conditions in synapse code flawed


neworderofjamie:
In some cases e.g. "$(V_pre) > -50.0" this is likely to work but e.g. a condition that is only true _before_ the spike reset code is run or one triggered using an extra global parameter will fail as the condition won't be true when it's re-checked next timestep (/after axonal delay) in the synapse code. There is also a much more minor issue that the threshold condition code is not considered when determining which neuron variables require queues in ``ModelSpec::finalize``.

I think the best fix is for each spike-like-event condition (give or take some merging which is already done) to generate its own array of events.

PR #380, opened 2020-12-11
Tracking of spike-like-event times

neworderofjamie:
This is a small extension of #376 to improve support for fully event-driven three-factor learning using spike-like events to deliver the third factor. By adding SET_NEEDS_PRE_SPIKE_EVENT_TIME(true) or SET_NEEDS_PREV_PRE_SPIKE_EVENT_TIME(true) macros to a weight update model in C++ or setting the is_pre_spike_event_time_required=True or is_prev_pre_spike_event_time_required=True kwargs in Python, the same sort of code used for tracking spike times is generated and ``$(seT)`` and ``$(prev_seT)`` can be used from weight update model code.


tnowotny (approved):
That is an easy one. Looks like the right thing to do and should be unproblematic. Well done.

neworderofjamie:
Thanks! The only problem relates to #379.

PR #384, opened 2020-12-29
Added access to $(num_pre) in row_build_code  ...

tnowotny:
… and to $(num_post) in col_build_code.

I found this useful in an application for a specific connectivity. In particular, I was trying to build a simple rectangular matrix where neuron 0 was connected to neuron 0 to n-1 and neuron 1 to neuron n to 2n-1 and so on.
So I tried
```python
 row_build_code=
        """
        const unsigned int row_length= $(num_post)/$(num_pre);
        const unsigned int offset= $(id_pre)*row_length;
        for (unsigned int i= 0; i < row_length; i++) {
            $(addSynapse, (offset + i));
        }
        $(endRow);
        """
```
... but it failed because in row_build_code ```num_pre``` was not translated. I don't see why one would not allow this and how the above code could be done as easily without it, so suggest to do the translation for ```num_pre``` as in this pull request. I hope I have covered all backends. I have tested it on the cuda backend where the above code now works fine in pygenn.

### DIFF
--- a/src/genn/backends/single_threaded_cpu/backend.cc
+++ b/src/genn/backends/single_threaded_cpu/backend.cc
@@ -594,6 +594,7 @@ void Backend::genInit(CodeStream &os, const ModelSpecMerged &modelMerged, Memory
                     popSubs.addVarSubstitution("id_post_begin", "0");
                     popSubs.addVarSubstitution("id_thread", "0");
                     popSubs.addVarSubstitution("num_threads", "1");
+                    popSubs.addVarSubstitution("num_pre", "group->numSrcNeurons");
                     popSubs.addVarSubstitution("num_post", "group->numTrgNeurons");
                 }
                 // Otherwise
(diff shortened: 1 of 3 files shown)


neworderofjamie (commented):
One more num_post is required at https://github.com/genn-team/genn/blob/master/src/genn/genn/code_generator/presynapticUpdateStrategySIMT.cc#L554 for procedural connectivity but, otherwise, this is a good idea

tnowotny:
Did you mean one more num_pre?

neworderofjamie:
Yes I did 😄

tnowotny:
ok ... like so?

neworderofjamie:
you could have got away with only a single additional substitution outside the if but that's totally fine 🤷

neworderofjamie (approved):

tnowotny:
good point ... but on the other hand it looks pretty tidy this way ... I will merge when the checks have finished (bot that they are testing this I am quite sure).

PR #387, opened 2020-12-30
Neuron initialisation merge fix

neworderofjamie:
When we're determining if the initialisation of neuron groups can be merged together (i.e. done by the same kernel), the main thing we care about is whether the state variables associated with the neuron group (and stuff like postsynaptic models associated with it) can be initialised in the same way. What the neuron simulation code is is generally irrelevant at this point. Therefore, previously, two neuron groups with neuron/postsynaptic models which differ only in that one requires an RNG would get merged. This is bad because, based on their order, either the RNG for the group that needs it wouldn't get seeded or code would be generated to seed a non-existent RNG. 

Anyway, simple fix is to only merge neuron groups where ``isSimRNGRequired() == other.isSimRNGRequired()``

Will add some unit tests around this sort of issue before merging.


tnowotny (approved):
Aha, ... I see. That was the problem. This merging business is a tricky one.

tnowotny:
PS: I also can confirm that this fixes the problem when I run the previously problematic AL model.

PR #388, opened 2021-01-04
Ensure PyGeNN never select OpenCL automatically

neworderofjamie:
Following my initial concern on your Mac. Assuming the single-threaded CPU backend always gets built, this change effectively means that, for now, you have explicitly create your ``GeNNModel`` with ``backend="OpenCL"`` if you want to try out OpenCL.


tnowotny (approved):
Yep ... definitely.

PR #395, opened 2021-01-25
Don't call cudaFree on zero copy variables

neworderofjamie:
Not sure when this one crept in - the comments suggest that, at one point, it was right 😄


tnowotny (approved):
Sorry for late review! No-brainer.

PR #396, opened 2021-02-03
PyGeNN debug builds

neworderofjamie:
One issue with PyGeNN has been that you can't debug into GeNN. In the face of a horrible divide by zero, I implemented this (basically just linking to the right libraries) in setup.py so now you can build a debug PyGeNN using:

```
make DEBUG=1 DYNAMIC=1 LIBRARY_DIRECTORY=`pwd`/pygenn/genn_wrapper/
python setup.py build_ext --debug develop
```

Then run your model via ``gdb --args python model.py``. For reasons I don't fully understand, Anaconda hsa no means of producing debug Python libraries on Windows so that's untested but otherwise it should work,

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -10,6 +10,10 @@ from setuptools.command.build_ext import build_ext
 from shutil import copytree, rmtree
 from generate_swig_interfaces import generateConfigs
 
+# Determine is this is a debug build
+# **YUCK** this is not a great test
+debug_build = "--debug" in sys.argv
+
 # Get CUDA path from environment variable - setting this up is a required CUDA post-install step
 cuda_path = os.environ.get("CUDA_PATH")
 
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -59,11 +69,12 @@
     "extra_link_args": []}
 
 # Always package LibGeNN
-package_data = ["genn_wrapper/genn_Release_DLL.*"] if windows else ["genn_wrapper/libgenn_dynamic.*"]
+package_data = ["genn_wrapper/genn" + genn_lib_suffix + ".*"] if windows else ["genn_wrapper/libgenn" + genn_lib_suffix + ".*"]
 
 # Copy dictionary and add libGeNN to apply to all modules that link against GeNN
 genn_extension_kwargs = deepcopy(extension_kwargs)
-genn_extension_kwargs["libraries"] =  ["genn_Release_DLL"] if windows else ["genn_dynamic"]
+genn_extension_kwargs["libraries"] =  ["genn" + genn_lib_suffix] if windows else ["genn" + genn_lib_suffix]
```
tnowotny, line 76: the distinction between windows and not seems pointless now.

tnowotny (changes requested):
See comment plus should this be mentioned in the documentation?

jamesturner246 (approved):

tnowotny (approved):
Looking good!

PR #401, opened 2021-02-23
Build wheels using Python3

### LINKED ISSUE #334: Stop building Python 2.7 wheels
Jenkins still uses Python 2.7 to build PyGeNN on machines where ``python`` points to Python 2.7 - if machines have a Python 3 environment installed (probably to be indicated with a Jenkins flag) use it instead.

neworderofjamie:
I've added environment variables to Jenkins so ``PYTHON`` specifies the python executable. I've set this to ``python3`` on the Ubuntu 18 machines, ``python`` on the arch machine and ``python3.8`` on the mac so Jenkins now builds Python 3 rather than 2 wheels

Fixed #334

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -227,7 +227,7 @@ for(b = 0; b < builderNodes.size(); b++) {
                             ${env.PYTHON} -m venv virtualenv
                             . virtualenv/bin/activate
 
-                            pip install "numpy>=1.17"
+                            pip install wheel "numpy>=1.17"
 
                             python setup.py clean --all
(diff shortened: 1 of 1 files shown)


tnowotny (approved):
Can't actually see where the python versions are set now ... i.e. ${env.Python} ... but I assume that's just not covered by the displayed diff?

neworderofjamie:
They're set through the Jenkins GUI for each builder

jamesturner246:
Could even use the standard python3 module venv, now you don't need to support python2. Doesn't really make a difference I suppose.

jamesturner246 (approved):

PR #404, opened 2021-03-19
izhikevich neurons shouldn't use auto-refractory

neworderofjamie:
Otherwise, if you inject sufficient current for them to spike every timestep it doesn't work due to the voltage reset being at the start of the sim code (presumably to support reading the spike voltage from learning rules). Do you think the automatic refractory mechanism should be turned off for all models with reset code?


tnowotny (approved):
I am not sure whether spiking every timestep should be a valid usage scenario but I agree with switching it off.

PR #405, opened 2021-03-22
Custom update

neworderofjamie:
I'm afraid this is quite a big one but I think it's going to be quite useful in spike-based machine learning land. Fundamentally, it lets you define 'custom updates' as GeNN models which can be run on demand (rather than every timestep). For example, this toy custom update which sets both a variable and a variable reference to the current time whenever it is run is used in several tests:
```c++
class SetTime : public CustomUpdateModels::Base
{
public:
    DECLARE_CUSTOM_UPDATE_MODEL(SetTime, 0, 1, 1);
    
    SET_UPDATE_CODE(
        "$(V) = $(t);\n"
        "$(R) = $(t);\n");

    SET_VARS({{"V", "scalar"}});
    SET_VAR_REFS({{"R", "scalar", VarAccessMode::READ_WRITE}})
};
IMPLEMENT_MODEL(SetTime);
```
We want custom updates to have their own state e.g. momentum variables in an optimizer but we also want them to be able to interact with parts of existing model. This requires a new 'variable reference' syntax in the model definition and these can be configured within your model like:
```c++
SetTime::VarReferences neuronVarReferences(createVarRef(ng, "V"));
SetTime::VarReferences wuPostVarReferences(createWUPostVarRef(sparseSG, "Post")); 
```
where the first parameter to the ``createXXXVarRef`` functions is a pointer to the neuron/synapse/current source group returned when you add them to the model. Because synapse groups contain many types of variable (postsynaptic model, presynaptic and postsynaptic), the functions taking a synapse group pointer need unique names. The only fly in this ointment is that per-synapse variables have different properties to the other kinds so custom updates need to be applied to them using different kernels and thus references to them are made using a different type:
```c++
SetTime::WUVarReferences wuDenseVarReferences(createWUVarRef(denseSG, "g"));
```
These 'weight update variable references' also have the additional feature that they can be used to define a link to a 'transpose' variable. These (currently limited to one per custom update) get updated at the end of the update using the algorithm described in https://developer.nvidia.com/blog/efficient-matrix-transpose-cuda-cc/ on SIMT backends:
```c++
CustomUpdateModels::Transpose::WUVarReferences wuDenseVarReferences2(createWUVarRef(denseSG2, "g", transposeSG2, "g")); 
```
Custom updates of all types can then be added to the model with:
```c++
model.addCustomUpdate<SetTime>("NeuronSetTime", "Test", {}, {0.0}, neuronVarReferences);
```
where the second parameter is the name of the update 'group' - these let you split custom updates amongst multiple kernels so, in this case you would launch one of these updates from your simulation code by calling ``updateTest()``. To summarise, the key parts of the design are:
- A single custom update model can be used in a multiple contexts - updating neuron or synapse state variables and potentially also calculating a transpose
- The custom update models have their own state so changing optimizer doesn't mean re-writing all the neuron/synapse models to include additional state.

To demonstrate this in the context of gradient-based learning, I have updated the superspike demo in https://github.com/neworderofjamie/genn_examples/tree/custom_update/superspike_demo and added a PyGeNN implementation both using custom updates. This not only makes the code a lot nicer than the custom CUDA kernels, but it makes it backend-agnostic and allows gradient-based learning to be applied to more complex models where kernel merging of custom update kernels is going to be necessary. The final feature required to complete this functionality will be custom updates which perform a reduction either across populations or across batches - this will let models like the multi-GPU eProp s-MNIST model to be defined entirely in GeNN.

Beyond the ability to add optimizers to models, I think this system has potential to be useful in lots of other contexts as well. I also think this functionality will allow Brian2 run-regularly statements to be implemented efficiently.


tnowotny:
Ok, this looks intriguing.
The bit I didn't get is how ``` SET_VAR_REFS({{"R", "scalar", VarAccessMode::READ_WRITE}})``` relates to the later creation of var refs using ```SetTime::VarReferences neuronVarReferences(createVarRef(ng, "V"));``` and so on.
Is the first declaring the existence of the var ref for the purpose of processing the code snippet and the second is making the actual connection to a neuron/synapse model?

neworderofjamie:
Yeah so the bit in the model description is telling the model that it requires a single variable reference (in the same way as it might require variables or parameters) and the second bit is the equivalent of _initialising_ the variables/parameters but, in the case of a variable reference, you initialise it by pointing it at a variable belonging to another population (https://github.com/neworderofjamie/genn_examples/blob/custom_update/superspike_demo/model.cc#L391-L410 where it has some context might make better sense)

tnowotny (commented):
Unless I missed it - should there be a brief expose (similar to the PR explanation) of this in the documentation?

neworderofjamie:
Ok, warnings and typo fixed and I've updated the documentation in several places

tnowotny (approved):
Looks like you did due diligence with the documentation now, thanks!
I will approve this even though I can't review the details on the basis that the approach makes sense and you have included appropriate tests. So hopefully no skeletons left in the closet.

PR #407, opened 2021-03-26
Ensure that 64-bit compiler is used on Windows

neworderofjamie:
For reasons best known to itself, without explicitly preferring a 64-bit toolchain, MSBuild was using a 32-bit version of the 64-bit compiler (if that makes sense) with resulting 4GB memory limit. This prevented very large models like the multi-area model compiling.


tnowotny (approved):
yep ... the wisdom of the Windows toolchain is unmatched.

PR #414, opened 2021-04-21
Fixed nasty race condition

neworderofjamie:
In #376 I added support for tracking 'previous' spike times. However, this generated ``preNeuronResetKernel`` code like this:

```c++
if(lid < group->spkCnt[0]) {
    group->prevST[group->spk[lid]] = t - DT;
}
        
if(lid == 0) {
    group->spkCnt[0] = 0;
}
```

Which has a nasty race condition in that the spike count may have been reset by the time other blocks try and access it. Presumably, this wasn't hit in our tests or the Pavlovian conditioning model because they were small enough that the whole kernel ran simultaneously. Better performing (this adds approximately 1.5 microseconds per-timestep to the Pavlovian conditioning model), more complex solutions might be possible but for now I've just split the two stages into two separate kernels which has had the side benefit of tidying up several nasty bits of the previous implementation)

### DIFF
--- a/include/genn/genn/code_generator/backendSIMT.h
+++ b/include/genn/genn/code_generator/backendSIMT.h
@@ -28,8 +28,9 @@ enum Kernel
     KernelSynapseDynamicsUpdate,
     KernelInitialize,
     KernelInitializeSparse,
-    KernelPreNeuronReset,
-    KernelPreSynapseReset,
+    KernelNeuronSpikeQueueUpdate,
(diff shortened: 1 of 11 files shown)


tnowotny (approved):
Classic trap. Glad you spotted it. Trying to think of more efficient solutions but coming up blank atm.

neworderofjamie:
Sadly not the cause of the bug that led me to fix it but hey 😄

PR #415, opened 2021-04-21
Tidied up handling of Extra Global Parameters in PyGeNN

### LINKED ISSUE #342: ``push_extra_global_param_to_device`` method is counter-intuitive
The prototypes of these functions are currently:
```python
push_extra_global_param_to_device(self, egp_name, size=1):

push_extra_global_param_to_device(self, pop_name, egp_name, size=1)
```
in ``Group`` and ``GeNNModel`` respectively. While this matches the underlying C++ API, the default of 1 was a really bad idea as, if you leave out ``size``, the calls _look_ like ``push_var_to_device`` calls but are often incorrect as they only copy the first element of the extra global parameter array. The size is known in Python so either use that or remove the default entirely.

neworderofjamie:
This fixes two issues:

1. You needed to pass the size of extra global parameter arrays to ``Group.pull_extra_global_param_from_device`` and ``Group.push_extra_global_param_to_device`` even though it was already know (#342)
2. When using non-array extra global parameters, you needed to call ``Group.set_extra_global_param`` before you loaded your model in order for them to be accessible via their views later.

I've solved both of these problems by, essentially handling them more like normal variables i.e. populating each group's extra global parameters dictionary from the model as soon as it is created. The ``Group.set_extra_global_param`` style methods simple call the ``set_values`` method on the appropriate ``ExtraGlobalParameter`` object in this dictionary and ``Group.pull_extra_global_param_from_device`` and ``Group.push_extra_global_param_to_device`` use the length of the values array (they shouldn't be used on non-array extra global parameters and this is now checked). I've tried to do this in a backward compatible manner and I think I've succeeded aside from the new requirement that postsynaptic model extra global parameters need to be pushed and pulled via ``SynapseGroup.pull_psm_extra_global_param_from_device`` and ``push_psm_extra_global_param_to_device`` which is pretty niche and thus I think is ok.

Fixes #342

### DIFF
--- a/pygenn/genn_groups.py
+++ b/pygenn/genn_groups.py
@@ -48,6 +48,25 @@ class Group(object):
         """
         self.vars[var_name].set_values(values)
 
+    @deprecated("This function was poorly named, use 'set_extra_global_param' instead")
+    def add_extra_global_param(self, param_name, param_values):
+        """Set extra global parameter
+
+        Args:
+        param_name      --  string with the name of the extra global parameter
+        param_values    --  iterable or a single value
+        """
+        self.set_extra_global_param(param_name, param_values)
+
+    def set_extra_global_param(self, param_name, param_values):
+        """Set extra global parameter
+
+        Args:
+        param_name      --  string with the name of the extra global parameter
+        param_values    --  iterable or a single value
+        """
+        self.extra_global_params[param_name].set_values(param_values)
+
     def pull_state_from_device(self):
         """Wrapper around GeNNModel.pull_state_from_device"""
         self._model.pull_state_from_device(self.name)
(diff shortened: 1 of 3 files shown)


jamesturner246 (approved):
Looks good. Nice backward compatibility. Doesn't break ml genn.

chanokin (approved):
Not breaking more tests than the ones already broken so I'll give it a pass

pygenn/genn_groups.py
```diff
@@ -60,14 +81,14 @@ def pull_var_from_device(self, var_name):
...
-    def pull_extra_global_param_from_device(self, egp_name, size=1):
+    def pull_extra_global_param_from_device(self, egp_name, size=None):
         """Wrapper around GeNNModel.pull_extra_global_param_from_device
 
         Args:
         egp_name    --  string with the name of the variable
         size        --  number of entries in EGP array
         """
-        self._model.pull_extra_global_param_from_device(self.name, egp_name, size)
+        self._pull_extra_global_param_to_device(egp_name, size)
```
tnowotny, line 91: I will admit I am not seeing fully to the bottom but is this really "pull ... to_device"? - seems against our normal naming conventions.

tnowotny (changes requested):
yeah ... I think you should change the name of that pull *to* device wrapper function to pull *from* device ... otherwise it looks all sensible.

pygenn/genn_groups.py
neworderofjamie: good spot - definitely something bad going on there

tnowotny: I think you just got the name wrong ... but as you matched it in the definition of the wrapper, it's not actually a bug. You can just refactor to the right name I think.

tnowotny (approved):
Yeah ... I think it is all good.

PR #417, opened 2021-04-22
Fixing OpenCL backend for AMD devices

### LINKED ISSUE #390: OpenCL correctness
After discussing our AMD problems with their OpenCL team, turns out that AMD drivers do some magic in the background to try and overlap kernel execution etc if they can't detect any dependencies. Because of the kernel merging, this detection doesn't work so we need synchronisation between each kernel for correct execution. However, they suggest a doable workaround using 'sub-buffers'. Essentially:
- We need to create a ``cl::Buffer``, large enough to contain all static allocations (i.e. everything beside extra global parameters and recording data) padded to ``CL_​DEVICE_​MEM_​BASE_​ADDR_​ALIGN``
- When we allocate individual variables, the _device_ buffers need to be created as sub-buffers from this buffer using ``cl::Buffer::createSubBuffer``
- The main ``cl::Buffer`` needs to be passed to **all** kernels.
- This approach doesn't work for runtime-sized allocations but, _hopefully_, there's fewer of these so passing the individual buffers to kernels that require them won't hit the ``CL_DEVICE_MAX_PARAMETER_SIZE`` limit

However, there are the following caveats:
> However the things are a bit more complicated than that. On gfx10  HW(navi) it should work. On gfx9 HW it may still require extra clFlush(), since compiler can detect if the global resource has R/W access and then runtime could consider it as a nop. To avoid that the app may need some arbitrary single DWORD write to the global mem object in the kernel that has access to the hidden objects.

neworderofjamie:
As described in #390, AMD drivers do some magic in the background to try and overlap kernel execution etc if they can't detect any inter-dependencies based on buffers passed to each kernel (there are no global variables in OpenCL). However, due to kernel merging, GeNN doesn't pass all the buffers required to each kernel. Doing this would bring back many of the scaling problems kernel merging was trying to address so instead we do the following (as suggested by AMD support):
- Create a single ``cl::Buffer``, large enough to contain all static allocations (i.e. everything beside extra global parameters and recording data) padded to ``CL_​DEVICE_​MEM_​BASE_​ADDR_​ALIGN``
- When we allocate individual variables, the _device_ buffers need to be created as sub-buffers from this buffer using ``cl::Buffer::createSubBuffer``
- Pass the main ``cl::Buffer`` to **all** kernels to indicate that they need to be run in sequence.

We already track memory allocations in order to estimate device memory usage so this is actually pretty trivial to do. The only flies in the ointment are that, NVIDIA's OpenCL implementation doesn't like the resultant kernels so we can't entirely replace the old codepath. The second tricky issue is that, to handle dynamic allocations like pointer extra global parameters and recording buffers, we allocate as large a buffer as our device allows (on our modern AMD card this is almost all it's memory but, supposedly, on older devices this might be a limitation) and dish these out of that 'above' the static allocations.

**Note** the meaningful changes are all in the OpenCL backend.cc. The remainder involve updating the ``BackendBase`` methods which generate static allocations to take ``MemAccess`` objects by reference rather than returning ``MemAccess`` objects which then get added to the global total. As well as allowing the start point of allocations in the static buffer to be accessed when required this is arguably a more fool-proof API.

Fixes #390

### DIFF
--- a/include/genn/backends/cuda/backend.h
+++ b/include/genn/backends/cuda/backend.h
@@ -172,7 +172,7 @@ public:
     virtual void genDefinitionsPreamble(CodeStream &os, const ModelSpecMerged &modelMerged) const override;
     virtual void genDefinitionsInternalPreamble(CodeStream &os, const ModelSpecMerged &modelMerged) const override;
     virtual void genRunnerPreamble(CodeStream &os, const ModelSpecMerged &modelMerged) const override;
-    virtual void genAllocateMemPreamble(CodeStream &os, const ModelSpecMerged &modelMerged) const override;
+    virtual void genAllocateMemPreamble(CodeStream &os, const ModelSpecMerged &modelMerged, const MemAlloc &allocations) const override;
     virtual void genStepTimeFinalisePreamble(CodeStream &os, const ModelSpecMerged &modelMerged) const override;
 
     virtual void genVariableDefinition(CodeStream &definitions, CodeStream &definitionsInternal, const std::string &type, const std::string &name, VarLocation loc) const override;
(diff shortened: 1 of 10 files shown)


tnowotny (approved):
Yeah, if that is what it takes to make OpenCL work correctly on AMD devices, it is not a horrible price to pay.

PR #418, opened 2021-04-26
Allow no self conns all to all

chanokin:
When All-to-All connections from PyNN GeNN disable self-connections (autapses?) there's a discrepancy between the size of variable values and the size of the variable view. Instead of just assigning the values to the view, I generate appropriate indices to set the values.

### DIFF
--- a/pygenn/genn_groups.py
+++ b/pygenn/genn_groups.py
@@ -1014,6 +1014,8 @@ class SynapseGroup(Group):
             # copy variables  directly into view
             # **NOTE** we assume order is row-major
             if self.is_dense:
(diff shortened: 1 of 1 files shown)


neworderofjamie:
I'm afraid I don't think this is the correct solution. If self-connections are removed, the connectivity is no longer dense (zero'd weights are different than no connection with many STDP rules for example). This issue should be fixed on the PyNN-GeNN side probably by using sparse connectivity if connectivity is all-to-all but with no self-connections (sadly this is a bit sub-optimal but I can't think of an alternative)

PR #419, opened 2021-05-10
Fixed very nasty bug in column-wise init

neworderofjamie:
The "connect_init" test was only testing with 'square' matrices so this wasn't being detected...


tnowotny (approved):
Seems like a straightforward fix.

PR #420, opened 2021-05-11
Speed up block size optimization

### LINKED ISSUE #259: Block size optimisation is slow
Both the inner and outer loops here https://github.com/genn-team/genn/blob/master/src/genn/backends/cuda/optimiser.cc#L177-L245 could be parallelised easily (personally I would use ``std::async``) - only slightly tricky thing is that ``kernelsToOptimise`` would either need to be changed to a more thread-safe data structure or accessed with a lock.

neworderofjamie:
As the first step in speeding up code generation, I made the block size optimisation multi-threaded. This seems to half build time on the machines I've tried it on which is definitely noticeable. I tried to parallelise the outer loop (over the block sizes) as well but, this got messier than the speedup I achieved justified because:
- It required the code for the different block sizes to be generated into separate directories and these to then be cleaned up.
- On Windows, parallel invocations of nvcc operating on files with the same name (seemingly the path isn't taken into account), clash so you would need to use the --keep-dir option to give each block size it's own temporary directory and clean _that_ up as well.

Fixes #259


tnowotny (approved):
Looks fairly straightforward but need to trust you that it works as expected.

PR #422, opened 2021-05-25
Fixes spike recorder

### LINKED ISSUE #421: MBody1 example project broken
Something went wrong with changes in SpikeRecorder so that the MBody1 example doesn't compile any more.
I suspect @neworderofjamie will know immediately what happened. But if not, let me know if you want me to dig into it. The call doesn't match any of the overloaded constructors:
```
MBody1Sim.cc:134:110: error: no matching function for call to ‘SpikeRecorder<>::SpikeRecorder(unsigned int* (*)(unsigned int), unsigned int& (*)(unsigned int), std::__cxx11::basic_string<char>)’
```

neworderofjamie:
In #392, the prototype of the ``getXXXCurrentSpikes`` and ``getXXXCurrentSpikeCount`` functions gained a ``batch`` parameter. The idea was that this wouldn't affect 'normal' users as it defaulted to zero but it did break the ``SpikeRecorder`` helper class which take function pointers to these functions.

Fixes #421

### DIFF
--- a/userproject/include/spikeRecorder.h
+++ b/userproject/include/spikeRecorder.h
@@ -150,14 +150,21 @@ template<typename Writer = SpikeWriterText>
 class SpikeRecorder : public Writer
 {
 public:
-    typedef unsigned int& (*GetCurrentSpikeCountFunc)();
-    typedef unsigned int* (*GetCurrentSpikesFunc)();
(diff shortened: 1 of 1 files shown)


tnowotny (approved):
yep that explains it & fixes that.

PR #426, opened 2021-06-14
Alt binomial

tnowotny:
The ```binomialInverseCDF``` is failing for large n, observed for instance, at n=5e8 with p approximately 1e-3. Here I suggest an alternative direct calculation of an inverse binomial CDF that appears to hold up longer and seems to give values similar to those provided by SciPy with different methods (only a spot test).
Probably we don't want to merge it like this but change the code that is using ```binomialInverseCDF``` to use the new function ``` directBinomialInverseCDF```. Feel free to push appropriate changes into this branch before merging.

### DIFF
--- a/include/genn/genn/initSparseConnectivitySnippet.h
+++ b/include/genn/genn/initSparseConnectivitySnippet.h
@@ -375,7 +375,7 @@ public:
     SET_CALC_MAX_ROW_LENGTH_FUNC(
         [](unsigned int numPre, unsigned int numPost, const std::vector<double> &pars)
         {
-            // Calculate suitable quantile for 0.9999 change when drawing numPre times
+            // Calculate suitable quantile for 0.9999 chance when drawing numPre times
             const double quantile = pow(0.9999, 1.0 / (double)numPre);
 
             // In each column the number of connections that end up in a row are distributed
--- a/src/genn/genn/binomial.cc
+++ b/src/genn/genn/binomial.cc
@@ -124,7 +124,9 @@ unsigned int binomialInverseCDF(double cdf, unsigned int n, double p)
 }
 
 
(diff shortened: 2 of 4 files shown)


tnowotny:
Actually, while we are working on this code, I am not sure why the "quantile" is calculated by taking the power of p here: https://github.com/genn-team/genn/blob/7cd8c549f45a6c6a3840e6aaa51c8ae6e849c20d/include/genn/genn/initSparseConnectivitySnippet.h#L379

tnowotny:
One more thought: We also can make it hybrid:
```
binomialInvertedCDF()
if (n < 400000) {
    betaBinomialInvertedCDF();
}
else {
    directBinomialInvertedCDF();
}
```
Furthermore the beta function version should probably use an interval halving "binary search" strategy as it's calculating the CDF from scratch every iteration anyway.

neworderofjamie:
Do the permutations of probabilities and population sizes at https://github.com/genn-team/genn/blob/alt_binomial/tests/unit/binomial.cc#L15-L16 cover the nasty permutations @tnowotny ?

neworderofjamie (approved):

tnowotny:
The cases seem to cover relevant combinations. I briefly thought about p= 0.0 being another interesting case but is it too artificial?
My other thought would be, your test is using n for both .9999^(1/n) and as the n in the binomial - that would normally not necessary be coupled ...
For the binomial, we might want to test even bigger n (as it can be number of synapses ...) - so maybe add n= 1e8?

tnowotny:
Final thought, I think an error of 1 would be acceptable/ expected for the worst corner cases

neworderofjamie:
Yes - this is a good point. I will add another test which imitates a fixed number pre/post connector (I don't think it matters from the pov of the binomial)

PR #427, opened 2021-06-21
Model hashing for incremental CUDA block size optimisation and merging

### LINKED ISSUE #292: Slow merging for large networks with custom models defines in Python

neworderofjamie:
This is following the general approach we discussed - building (somewhat irregular) [Merkle trees ](https://en.wikipedia.org/wiki/Merkle_tree) of the model based on SHA-1 hashes (common choice for data integrity checking - used in Git, BitTorrent and others).

To prevent a lot duplication and to allow existing unit tests to be used to validate the correctness of the leaf node comparisons, I started this work by reimplementing the model merging checks using hash comparisons (compare https://github.com/genn-team/genn/blob/master/src/genn/genn/synapseGroup.cc#L592-L625 to https://github.com/genn-team/genn/blob/module_hashing_scope_reduce/src/genn/genn/synapseGroup.cc#L601-L629 for a representative comparison).

Stupidly, I then got most of the way through implementing model comparison and realised that it's a harder problem than I thought as all the modules ``#include "definitionsInternal.h"`` which means pretty much any model change leads to all modules getting recompiled. However, this doesn't apply to the CUDA block size optimisation where only changes which affect the actual kernel code (the contents of the data structures going into it doesn't matter at compile time) matter (essentially the same criteria as mergeability) meaning this was trivial to implement.

Therefore, while it doesn't quite go as far as I originally hoped, I think this 'reduced scope version' is still very much worth merging because:
- The hash-based merging implementation is not only more elegant  but also only O(N) which should fix #292.
- When using CUDA, the block size optimisation time dominate (as it compiles each module twice) so just the speedups we achieve here are going to improve the user experience.
- This PR includes some tidying that will lay the groundwork for incremental model building: you can now generate modules independently (before the code for dishing out constant cache required modules to be built in order)


tnowotny (commented):
Yes, I think I get most of it and it seems like a good approach. It's a shame that the different "modules" are so interlinked that one can't yet go further but what about the case where the model is literally unchanged? Is that now automatically covered as not needing a recompile?

neworderofjamie:
Excellent! I think, in future, some de-coupling of the modules would be possible which would have the side benefit of reducing the size of the runner but, that's definitely one for another day. 

This branch currently only uses the hashes to determine whether CUDA block size optimization is required but, I think you have a good point in that not regenerating any code when the model is unchanged would be sensible. I might restore the appropriate bits of code in a second PR though if that's ok?

tnowotny (approved):
Sure, agreed.

PR #429, opened 2021-06-28
Python multithread fixes

neworderofjamie:
I would normally just wait until the rest of the hash-based model comparison stuff is done but Tom is using master (as maybe are you for the mushroom body model) so I cherry picked out these fixes for issues I broke in #427 which I discovered while trying to use PyGeNN. Basically:

- If you're using CUDA and have never built a model before, the output directory wasn't get created and thus everything else failed
- We were calling ``ModelSpecMerged::getXXXArchetypeHashDigest`` from multiple worker threads in the block size optimiser. However, if you're using PyGeNN, this function will call into Python implementations of models and Python is essentially single threaded. SWIG does have a mechanism for generating thread-safe code but I couldn't get it to work and all it would do is serialise these calls so the simpler solution was to calculate hashes on the main thread.


tnowotny (approved):
Ok, thanks for fixing!

PR #430, opened 2021-06-28
Use hashes to detect any changes to model and recompile

neworderofjamie:
This follows on from #427 and hashes together additional model properties and backend preferences so _any_ model changes can be detected. If the model hasn't changed, no code is regenerated and therefore the compiler doesn't recompile anything. This _could_ clearly cause really insidious bugs but I've added a lot of unit tests to test various sorts of small model change so **hopefully** it's reasonably safe. Also I've added an option to force rebuilding of the model (triggered by the ``-f`` command line argument to genn-buildmodel, the ``force_rebuild`` flag to ``pygenn.GeNNModel.build`` and the ``--force-rebuild`` option for the SpineML generator. Notable parts of the change are:

- In GroupMerged, as part of the merging logic, there was previously functionality to determine whether parameters were referenced in a given part of the model before adding the values of parameters which vary across the merged groups to the group struct. We need this functionality without the test for varying across the group as both hardcoded and varying parameters ultimately change the model so there has been some refactoring here.
- The CUDA block-size optimiser now works on files with different names than those used for the actual model so it doesn't overwrite anything and thus trigger rebuilding

The following graph shows the evolution of genn-buildmodel and compilation times of (don't worry, Visual C++ definitely isn't a faster compiler, my Windows machine was just recently upgraded with a fancy NVMe SSD and the Linux machine is still using an actual hard disk):

GeNN code generation performance (1)


tnowotny (approved):
Thanks for your hard work on this! I think it is a really good step towards better usability for exploration type applications - at least if only variable values change.

neworderofjamie:
Thanks 😃 It has made building the mushroom body notebooks much more of a pleasant experience

PR #431, opened 2021-07-01
Var location hashing

neworderofjamie:
Whether variables are located on the host/device does not effect the actual simulation code (i.e. neuronUpdate.cc or whatever) but it does change the runner so needs to be taken into account in the top-level hash used for detecting model changes.


tnowotny (approved):
Yep ... that broadly makes sense.

Issue #359, opened 2020-09-09
Exception handling issues on Windows


neworderofjamie:
There seem to be several issues with exception handling on Windows which need further investigation:

1. Exceptions raised in C++ boilerplate generated by SWIG aren't getting back to Python
2. Exceptions raised within runner aren't get back to application if runner is loaded dynamically. You can catch them in a ``catch(...)`` handler but any attempt to call e.g. ``what()`` fails. This is perhaps related to ``warning C4297: 'updateSynapses': function assumed not to throw an exception but does`` warnings. Setting ``/EHs`` should solve this but the CUDA Visual Studio integration doesn't expose this and sets ``/EHsc`` which "assumes that functions declared as extern "C" never throw a C++ exception."

PR #432, opened 2021-07-02
PyGeNN helpers for accessing model internals

neworderofjamie:
Namely:
- Spike times (and all the thrilling variants)
- inSyn (this will also be a useful building block for https://github.com/genn-team/pynn_genn/issues/16)

All of these things are accessible already if you know the right thing to hack at, but the names are inconsistent etc. The only nuance is making ``SynapseGroup::isPSModelMerged`` public rather than only accessible in the code generator. This is less than idea as it's value is meaningless at the point the non-PyGeNN user is likely to call it (merging only happens during ``ModelSpec::finalize`` but the only alternative would just be to rely on catching the exception raised when you try and access a merged-away inSyn which #359 means isn't possible


tnowotny (approved):
I am all for it!

PR #433, opened 2021-07-06
PyGeNN exception handling

### LINKED ISSUE #359: Exception handling issues on Windows

neworderofjamie:
This has been annoying me for ages - C++ exceptions weren't making it through to Python so debugging errors was a nightmare. As discussed in #359, I _thought_ this was a Windows issue but, I think, this is just because, on Windows, there's no default exception handler so a C++ program that throws e.g. a ``std::exception`` into the ether, displays no output and terminates. This is what was happening with all exceptions emanating from C++ (aside for, for some reason, exceptions generated by ``create_backend`` which I had manually added an exception handler for previously 🤷‍♂️). After reading the documentation again, there is some additional SWIG directives which can be added to add a blanket mapping of standard C++ exceptions to SWIG (which in turn maps them to Python) (https://github.com/swig/swig/blob/master/Lib/exception.i#L289-L307). This seems to work very nicely even from the CUDA code that was emitting the weird warnings discussed in #359!

Fixed #359


tnowotny (approved):
Nice one ... this should make life a little easier in the future.
I haven't looked into why all the tests seem to have failed though ...

neworderofjamie:
Amusingly, it seems to be because whatever awful things the IJRAs are doing to @jamesturner246 's machine is using all the GPU memory and the CUDA out of memory errors are not being caught 😄

PR #436, opened 2021-07-12
Include versions in hash

neworderofjamie:
I had the realisation over the weekend that, models should be rebuilt if the GeNN version or the model changes. There's basically two use cases here:

1. Downloading zip files
2. Checking out of git

In the case of 1, it's reasonably safe to assume that it's a release version of GeNN so including the hash of the version string is fine. When GeNN's been checked out of git, we can use the git head revision hash. This is pretty simple using both make (aside from that the ``$(file)`` function is only supported on super-bleeding edge versions) and MSBuild - the only tricky bits are handling the error cases (we don't want the build failing if there's no git or git/shell stderr messages making their way into ``GIT_HASH``)

### DIFF
--- a/src/genn/MakefileCommon
+++ b/src/genn/MakefileCommon
@@ -4,10 +4,14 @@ OS_UPPER			:=$(shell uname -s 2>/dev/null | tr [:lower:] [:upper:])
 DARWIN				:=$(strip $(findstring DARWIN,$(OS_UPPER)))
 
 # Get directory of this makefile i.e. GeNN directory (means make can be invoked else where)
-GENN_DIR 			:= $(abspath $(dir $(lastword $(MAKEFILE_LIST)))../../)
+GENN_DIR			:= $(abspath $(dir $(lastword $(MAKEFILE_LIST)))../../)
+
+# Read GeNN version and git hash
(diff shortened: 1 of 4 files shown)


tnowotny (commented):
Very good point ... easily forgotten and could have been causing trouble down the road.
I assume, if either the the version string or the hash don't yield anything, this doesn't break?

neworderofjamie:
The version string should always be present but, if git fails, it will just result in an empty string.

tnowotny (approved):
Ok, I see ... if git fails that would then simply mean no recompilation (as the empty strings presumably produce the same hash) ... I guess that's right if genn wasn't checked out via git.
If the failure was for another reason one probably would have wanted to default to "always recompile" ... hum ...
I guess in the balance it is probably ok.

neworderofjamie:
Yeah, if you aren't using git then the the hash will only change if the GeNN version changes. It might fail are if git isn't in your path but making it recompile every time in that case seems a little opaque

tnowotny:
There is no ideal solution to this unless you wanted to parse error messages but that seems out of the question. So this is fine to merge I'd think.

PR #438, opened 2021-07-19
std::function wrapping

neworderofjamie:
When wrapping row length and kernel size calculation functions in PyGeNN we missed one step which lead to large numbers of irritating warnings. This PR rectifies this and, additionally, I spotted that setting backend preferences which were strings (e.g. ``userCxxFlagsGNU="-march=native"``) didn't work from PyGeNN - this simply required the SWIG ``<std_string.i>`` interface to be included in all backend SWIG interface files.


tnowotny (approved):
Looks good. I am glad we fixed this one. Good catch with the other bug as well.

PR #439, opened 2021-07-20
Nasty bugs in custom update kernels

neworderofjamie:
Two really nasty bugs have come to light as I've started building models which do more complex stuff with custom updates:

Firstly, we were building the group start indices (the arrays that get binary searched to find the index to the right data if groups get merged) across all custom update groups. This is clearly wrong as separate kernel(s) are generated for each update group.

Secondly, in the CUDA kernel, the group start indices for custom updates and custom weight updates weren't being generated together so if there were enough of either to require binary searching there would be problems.


neworderofjamie:
Ironically, in parallel, I was adding some new functionality and extending the tests to cover it and these were _starting_ to fail because of the second of these issues.

tnowotny (commented):
Sorry for delay. I will have to trust you on the details on this one.
Is there a very typical case you can add as a test?

tnowotny (approved):
Then I would say, let's go ahead and merge these.

PR #440, opened 2021-07-20
Fix issue where newly calculated memory estimate was added to cached value if model was changed

neworderofjamie:


tnowotny (approved):
Aha .. that makes a lot of sense ...

PR #444, opened 2021-07-23
Use symbolic links in /tmp to fix "path name spaces problem"

### LINKED ISSUE #424: genn-buildmodel.sh appears to fail when spaces in file path
It appears that genn-buildmodel.sh fails when the model source is in a location with a path name that contains spaces.

tnowotny:
Fixes #424.

### DIFF
--- a/bin/genn-buildmodel.sh
+++ b/bin/genn-buildmodel.sh
@@ -52,13 +52,20 @@ done
 if [[ -z "$MODEL" ]]; then
     genn_error $LINENO 2 "no model file given"
 fi
-pushd $OUT_PATH > /dev/null
-OUT_PATH="$PWD"
+
+# Use pushd to get an absolute path and symbolic links in /tmp to avoid problems
+# with spaces in the path names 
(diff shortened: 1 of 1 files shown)


bin/genn-buildmodel.sh
```diff
@@ -52,13 +52,20 @@ done
 if [[ -z "$MODEL" ]]; then
     genn_error $LINENO 2 "no model file given"
 fi
-pushd $OUT_PATH > /dev/null
-OUT_PATH="$PWD"
+
+# Use pushd to get an absolute path and symbolic links in /tmp to avoid problems
+# with spaces in the path names 
+pushd "$OUT_PATH" > /dev/null
+OUT_PATH=/tmp/genn_outpath$RANDOM
```
neworderofjamie, line 59: It would be _marginally_ nicer to use ``OUTPUT=$(mktemp)`` (and ``MODEL=$(mktemp)`` instead). Better chance of being robust to multiple users running GeNN simultaneously on the same machine.

neworderofjamie (approved):

PR #445, opened 2021-07-23
Setting spikes manually from PyGeNN

neworderofjamie:
_Getting_ current spikes was exposed via the ``pygenn.genn_groups.NeuronGroup.current_spikes`` property but setting these wasn't implemented. This PR does that!

### DIFF
--- a/pygenn/genn_groups.py
+++ b/pygenn/genn_groups.py
@@ -346,8 +346,8 @@ class NeuronGroup(Group):
         """Current spikes from GeNN"""
         # Get current spike queue pointer
         d = self.spike_que_ptr[0]
-        
-        # If batch size is zero, return single slice of spikes
+
+        # If batch size is one, return single slice of spikes
         if self._model.batch_size == 1:
             return self.spikes[0, d, 0:self.spike_count[0, d]]
         # Otherwise, return list of slices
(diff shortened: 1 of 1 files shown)


pygenn/genn_groups.py
```diff
@@ -346,15 +346,40 @@ def current_spikes(self):
...
+            if len(spikes) != self._model.batch_size:
+                raise Exception("When using a batched model, you must "
+                                "set current spikes using a list of spikes "
+                                "for each batch")
+
+            # Loop through batches and set spike counts and spike data
+            for b, batch_spikes in enumerate(spikes):
+                num_spikes = len(batch_spikes)
+                self.spike_count[b, d] = num_spikes
+                self.spikes[b, d, 0:num_spikes] = spikes
```
tnowotny, line 381: shouldn't this be ```self.spikes[b, d, 0:num_spikes] = batch_spikes```?

tnowotny (changes requested):
I *think* I have spotted a mistake there for the first time since forever.

pygenn/genn_groups.py
neworderofjamie: very good spot - serves me right for being too lazy to test out the functionality on a batched model

tnowotny (approved):
That should do it.

PR #446, opened 2021-08-02
Variable references to custom update variables

neworderofjamie:
This basically lets you 'chain' custom updates so they can process each other's results. This is used for implementing batch reductions, the results of which are used to update weights in next PR :smile:


tnowotny (approved):
I *think* it's all good.

PR #443, opened 2021-07-22
Validate population and variable names

neworderofjamie:
GeNN really breaks if you have variable/population names that aren't valid C identifier names but we never checked this.

Fixes #423

PR #448, opened 2021-08-10
Less aggressive name validation

neworderofjamie:
I realised as my microcircuit model failed to build that #443 was a little over-zealous - **population** names are fine to start with a number as, when they're used in generated code, they are always prefixed by a variable name (which can't start with a number)


tnowotny (approved):
Good point - even though I would never encourage people to start a name with a number.
Also good idea to test against use of dashes ... some people might well have that idea.

PR #449, opened 2021-08-11
NCCL multi-GPU reductions

neworderofjamie:
As hoped, the step from #447 to multi-GPU training was pretty small. Basically, we re-use the code from #316 which allows the generation of merged _host_ code (so big models don't end up with hundreds of lines of NCCL calls) to generate [NCCL](https://docs.nvidia.com/deeplearning/nccl/) reductions after the batch reduction kernels. These operations use standard CUDA stream semantics so won't cause any synchronisation unless they're followed by a synchronisation or memcpy operation. The only slightly gnarly bit is that we don't want platform-specific CUDA (or in this case NCCL) types to bleed out of the generated code so the unique id (which is used to identify the 'clique' of nodes that are communicating) is exposed as a pointer (which is then exposed to PyGeNN as a numpy view).

You need to combine the code this PR generates with _something_ that can spawn multiple processes and communicate between them. I have used [mpi4py](https://mpi4py.readthedocs.io/en/stable/) on JUWELS booster where you can do something like:
```python
from mpi4py import MPI

# Get communicator
comm = MPI.COMM_WORLD

# Get our rank and number of ranks
rank = comm.Get_rank()
num_ranks = comm.Get_size()

# Generate unique ID for our NCCL 'clique' on first rank
if rank == 0:
    model._slm.nccl_generate_unique_id()

# Broadcast our  NCCL clique ID across all ranks
nccl_unique_id_view = model._slm.nccl_assign_external_unique_id()
comm.Bcast(nccl_unique_id_view, root=0)

# Initialise NCCL communicator
model._slm.nccl_init_communicator(rank, num_ranks)
```
But you could also use the Python multiprocessing module e.g. on a multi-GPU workstation where you might not want to use MPI (I envisage this abstraction being handled by mlGeNN as you're likely to also want to use MPI/multi-processing to share e.g. number of correct classifications across all nodes)

Results from JUWELS booster (each node has 4*A100 GPU) and JADE2 (each node has 8*V100 GPU):
Figure_1


tnowotny (approved):
The approach seems sound and as far as I can tell it should be alright (not checked each implementation detail).

PR #458, opened 2021-09-15
Postsynaptic model target

### LINKED ISSUE #451: Make pointing synapse groups at additional input variables easier
Currently you need to define a new postsynaptic model for each additional input variable i.e. you should be able to use a standard "DeltaCurr" rather than having to define:
```python
feedback_psm_model = genn_model.create_custom_postsynaptic_class(
    "feedback_psm",
    apply_input_code="""
    $(ISynFeedback) += $(inSyn);
    $(inSyn) = 0;
    """)
```
It would be much nicer if the output variable of the apply input code was just a symbol and you could just call a method on the synapse group to set which variable it targets.

neworderofjamie:
As described in #451, previously, you needed to define a new postsynaptic model for each additional input variable you might want to connect synapses to e.g. you should be able to use a standard ``DeltaCurr`` rather than having to define:
```python
feedback_psm_model = genn_model.create_custom_postsynaptic_class(
    "feedback_psm",
    apply_input_code="""
    $(ISynFeedback) += $(inSyn);
    $(inSyn) = 0;
    """)
```
just to connect synapses to a neuron's ``ISynFeedback`` additional input variable. This PR instead allows you to do this in C++:
```c++
auto *syn= model.addSynapsePopulation<WeightUpdateModels::StaticPulse, PostsynapticModels::DeltaCurr>(
    "Syn", SynapseMatrixType::DENSE_INDIVIDUALG, NO_DELAY,
    "Pre", "Post",
    {}, { 1.0 },
    {}, {});
syn->setPSOutputVar("ISynFeedback");
```
or in Python
```python
syn = model.add_synapse_population("Syn", "DENSE_INDIVIDUALG", NO_DELAY,
    pre_pop, post_pop,
    "StaticPulse", {}, {"g": 1.0}, {}, {},
    "DeltaCurr", {}, {})
syn.ps_output_var = "ISynFeedback"
```

The implementation is trivial, my only question is whether "PS output variable" is a good names for these things rather than, for instance, "Target input variable" or "Postsynaptic input variable". Any thoughts much appreciated!

Fixes #451

### DIFF
--- a/include/genn/genn/synapseGroup.h
+++ b/include/genn/genn/synapseGroup.h
@@ -66,7 +66,7 @@ public:
 
     //! Set name of neuron input variable postsynaptic model will target
     /*! This should either be 'Isyn' or the name of one of the postsynaptic neuron's additional input variables. */
-    void setPSOutputVar(const std::string &varName);
+    void setPSTargetVar(const std::string &varName);
     
     //! Set location of sparse connectivity initialiser extra global parameter
     /*! This is ignored for simulations on hardware with a single memory space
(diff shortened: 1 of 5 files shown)


tnowotny (commented):
Yes, all good but could consider a different name. I think input/output maybe not ideal concepts as one's input is the other one's output. 
How about "PSTargetVar"?
I think you have used "PS" for post-synaptic before so is in good keeping with that practice.

neworderofjamie:
Excellent suggestion!

PR #460, opened 2021-09-20
Fix multiple issues with sparse synapse index narrowing

neworderofjamie:
In #274 I implemented support for 'narrow' sparse synapse indices so, if the postsynaptic population was < 65536 neurons, ``uint16_t`` would be used and if < 256 neurons, ``uint8_t``. However:

1.  PyGeNN always tried to access  the sparse synapse indices as ``uint32_t`` - it now calls``SynapseGroup::getSparseIndType`` to get correct type
2. In the generated code for ``pushXXConnectivityToDevice`` and ``pullXXConnectivityToDevice``, ``unsigned int`` was always being used - this also now uses the correct type.

### DIFF
--- a/include/genn/genn/synapseGroup.h
+++ b/include/genn/genn/synapseGroup.h
@@ -142,7 +142,13 @@ public:
     //! Is this synapse group a weight-sharing slave
     bool isWeightSharingSlave() const { return (getWeightSharingMaster() != nullptr); }
 
+    //! Has this synapse group's postsynaptic model been merged with others
+    /*! NOTE: this can only be called after model is finalized but needs to be public for PyGeNN */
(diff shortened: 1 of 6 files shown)


tnowotny (approved):
Oh dear ... good catch there. I suspect this came out from the guy's report about huge post-indices fro a small post-population?
Can you tell that you have caught everything now (I unfortunately can't ...)?

neworderofjamie:
It did indeed...The updated feature test definitely now covers the second bug but, PyGeNN is currently not well tested as most of the testing (aside from that done at a higher level by PyNNGeNN or mlGeNN) happens at the C++ level. For the low-level unit tests I think that is advantageous but, as part of the test revamp we've been talking about, maybe porting all the feature tests to PyGeNN is the solution.

PR #377, opened 2020-11-24
Time-driven pre and postsynaptic traces

neworderofjamie:
#152 added support for pre and postsynaptic weight update model variables to allow better encapsulation of learning vs neuron dynamics. However, these were purely event-driven dynamics which, retrospectively:

1. Even with "classic" event-driven STDP, is an inefficient approach for GPUs as loads of threads are going to waiting while a small subset are calculating exponentials - having all threads performing a much cheaper multiply every timestep is much more GPU-friendly (clearly, back in the heady days of 2018, optimising for SpiNNaker was too recent a memory)
2. Delays are annoying as you need to copy the event-driven traces between delay buffers every timestep anyway
3. Event-based traces are extra inefficient with continuous learning as you would need to calculate exponentials every timestep
4. Purely event-driven dynamics don't let you use traces to capture e.g. low-pass filtered neuron voltage

In this P.R. we introduce new syntax for time-driven pre and postsynaptic traces:

### Python
For example, the following simple additive STDP rule:
```python
stdp_additive = genn_model.create_custom_weight_update_class(
    "STDPAdditive",
    param_names=["tauPlus", "tauMinus", "aPlus", "aMinus", "wMin", "wMax"],
    var_name_types=[("g", "scalar")],
    pre_var_name_types=[("preTrace", "scalar")],
    post_var_name_types=[("postTrace", "scalar")],
    derived_params=[("aPlusScaled", genn_model.create_dpf_class(lambda pars, dt: pars[2] * (pars[5] - pars[4]))()),
                    ("aMinusScaled", genn_model.create_dpf_class(lambda pars, dt: pars[3] * (pars[5] - pars[4]))())],

    sim_code="""
        $(addToInSyn, $(g));
        const scalar dt = $(t) - $(sT_post);
        if(dt > 0) {
            const scalar timing = exp(-dt / $(tauMinus));
            const scalar newWeight = $(g) - ($(aMinusScaled) * $(postTrace) * timing);
            $(g) = fmin($(wMax), fmax($(wMin), newWeight));
        }
        """,

    learn_post_code="""        
        const scalar dt = $(t) - $(sT_pre);
        if(dt > 0) {
            const scalar timing = exp(-dt / $(tauPlus));
            const scalar newWeight = $(g) + ($(aPlusScaled) * $(preTrace) * timing);
            $(g) = fmin($(wMax), fmax($(wMin), newWeight));
        }
        """,

    pre_spike_code="""
        const scalar dt = $(t) - $(sT_pre);
[...]

PR #461, opened 2021-09-21
Merge pre and postsynaptic update

neworderofjamie:
GeNN has supported pre and postsynaptic weight update models variables and associated update code since #152 (with some extra flexibility introduced in #377). This is nice in terms of modularity **but**, the implementation is very sub-optimal for models with complex connectivity when compared to the hackier approach. If you have a neuron population with multiple incoming or outgoing synapse populations, all with the same learning rule, the variables get duplicated and, in the neuron kernel, they all get read from memory, updated using the same code and written back to memory. 

Postsynaptic models suffered from the same problem which I solved in #201. This PR basically does the same for these pre and postsynaptic weight updates. Essentially two outgoing synapse groups with presynaptic update code (incoming with postsynaptic is basically equivalent) will get merged together if:

1. They have the same weight update model (this could potentially be relaxed a little as most of the model can be ignored)
2. Any parameters/derived parameters referenced in the presynaptic update code are the same
3. There are no extra global parameters referenced in the presynaptic update code
4. All their presynaptic variables are initialised to the same constant value (variables associated with merged groups are not accessible to the user so, when combined with the other constraints, this means they will evolve in the same way)

The previous change for postsynaptic models significantly improved performance on the microcircuit model so I would imagine this change will be similarly beneficial once we start simulating models which combine more complex connectivity and learning.

### DIFF
--- a/include/genn/genn/code_generator/groupMerged.h
+++ b/include/genn/genn/code_generator/groupMerged.h
@@ -986,7 +986,7 @@ private:
                        Models::Base::VarVec(WeightUpdateModels::Base::*getVars)(void) const,
                        bool(NeuronUpdateGroupMerged::*isParamHeterogeneous)(size_t, size_t) const,
                        bool(NeuronUpdateGroupMerged::*isDerivedParamHeterogeneous)(size_t, size_t) const,
-                       const std::string&(SynapseGroupInternal::*getVarMergeSuffix)(void) const);
+                       const std::string&(SynapseGroupInternal::*getFusedVarSuffix)(void) const);
 
     //! Is the incoming synapse weight update model parameter referenced?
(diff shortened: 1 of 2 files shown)


tnowotny:
Looking at the thesaurus, maybe we should call this process "fuse", "fused".
And reserve "merge", "merged" for the kernel merging?
I also think "fused" presynaptic vars has the right sound to it.

neworderofjamie:
I like it - will save _that_  change for when I'm at home and can throw the Visual Studio magic rename tool at it :smile:

neworderofjamie:
I'm afraid the renaming somewhat dwarves the actual changes but still...The most meaningful new bits are probably the rules which define when updates can be merged and are in ``SynapseGroup.cc``

PR #471, opened 2021-10-25
Automatically find Visual Studio

### LINKED ISSUE #466: Automatically find Visual Studio on Windows
Especially for people using Python IDEs, it's very annoying having to use PyGeNN from a Visual Studio environment. Brian2 does the following:

```python
from setuptools import mscv
_msvc_env = msvc.msvc14_get_vc_env(arch_name)
```

Caches ``_msvc_env`` as the lookup is heavyweight and then just adds the result into the environment passed to ``subprocess.check_call``

neworderofjamie:
This needs two pieces of nastiness:
* ``setuptools`` to configure MSVC environment (same as Brian2)
* ``distutils`` to find msbuild within MSVC environment (due to weirdness of whatever API ``subprocess.check_call`` uses underneath)

This means you can use PyGeNN from a standard Python prompt e.g. an Anaconda prompt.

Fixes #466


neworderofjamie:
Tested on:
- Windows 10, Python 3.9.12 and VS 2017 15.9

PR #472, opened 2021-10-25
Sync PyGeNN version to main GeNN version

### LINKED ISSUE #442: Synchronise PyGeNN version with GeNN version
The pre-release version numbers of PyGeNN aren't properly semantically versioned which is annoying and, furthermore, I think we can now class PyGeNN as being released at this point. With 4.6.0 will switch PyGeNN to using the same version as GeNN itself.

neworderofjamie:
By doing the following:
* Read version.txt to provide package version in setup.py
* Query package metadata to provide version at runtime (pygenn.__version__)

The PyGeNN version will be synchronised with the main GeNN version (PyGeNN versions were previously manually set to 0.X.Y)

Fixes #442

### DIFF
--- a/pygenn/__init__.py
+++ b/pygenn/__init__.py
@@ -1,3 +1,13 @@
-# Import pygenn interface
+# python imports
+import sys
+
+# pygenn interface
 from .genn_groups import SynapseGroup, NeuronGroup, CurrentSource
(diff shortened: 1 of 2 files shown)


tnowotny (approved):
I think this is indeed a good idea.

PR #474, opened 2021-10-26
4.6.0 release

neworderofjamie:
A few updates to the documentation for the next GeNN release. I also realised that lots of bits of pygenn can be accessed via a shorter syntax so grepped that everywhere in the documentation.


doxygen/09_ReleaseNotes.dox
```diff
@@ -1,4 +1,30 @@
...
+3. Previously, weight update model pre and postsynaptic updates and variables got duplicated in the neuron kernel. This was very innefficient and these can now be 'fused' together by setting ModelSpec::setFusePrePostWeightUpdateModels.
+4. PyGeNN now shares a version with GeNN itself and this will be accessible via ``pygenn.__version__``. 
+5. The names of populations and variables are now validated to prevent code with invalid variable names being generated.
+6. As well as being able to read the current spikes via the pygenn.NeuronGroup.current_spikes property, they can now also be set.
+7. Spike-like events were previously not exposed to PyGeNN. These can now be pushed and pulled via pygenn.NeuronGroup.pull_spike_events_from_device, pygenn.NeuronGroup.push_spike_events_to_device, pygenn.NeuronGroup.pull_current_spike_events_from_device and pygenn.NeuronGroup.push_current_spike_events_to_device; and accessed via pygenn.NeuronGroup.current_spike_events.
```
tnowotny, line 16: stupid question: Can one access only ```current_spike_events``` or also the full ring buffer through something like ```spike_events```? And should one be able to do that?

tnowotny (commented):
Looks all good but one question (see comment).

doxygen/09_ReleaseNotes.dox
neworderofjamie: Yes you can. ``NeuronGroup.spike_events`` and ``NeuronGroup.spikes`` exist (and are what the corresponding ``NeuronGroup.current_spikes`` and ``NeuronGroup.current_spikes_events`` access)

tnowotny: Ah, then I think we can consider this PR approved.

PR #476, opened 2021-11-09
Include model name in DLL name on Windows

neworderofjamie:
Due to the inflexibility of Windows DLLfinding, ``runner_Release.dll`` (or ``runner_Debug.dll``) is created in the same directory as the script (rather than in the ``***_CODE`` sub-directory with everything else). Previously this was only a minor annoyance but, now the caching system (#430)  means that models don't always get rebuilt, it can lead to an issue where, if you have multiple models in one directory, a DLL for the wrong model will get loaded.

This PR solves this by adding a flag which includes the model name in the DLL name (for backward compatibility, for now, this is not the default). In PyGeNN, as we're in control of both building and loading the DLL, this is now the default.

### DIFF
--- a/generate_swig_interfaces.py
+++ b/generate_swig_interfaces.py
@@ -601,7 +601,7 @@ def generateConfigs(gennPath, backends):
             #ifdef _WIN32
                 // Create MSBuild project to compile and link all generated modules
                 std::ofstream makefile((outputPath / "runner.vcxproj").str());
-                CodeGenerator::generateMSBuild(makefile, backend, "", output.first);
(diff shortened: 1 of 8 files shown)


tnowotny (commented):
Looks good. What's the reasoning behind not including the name by default outside of pygenn? Simply backward compatibility?

neworderofjamie:
If you were using GeNN in 'classic' C++ mode with this enabled, all your existing projects would be left pointing at libraries with the old style names which would no longer get updated. Will make this the default behaviour next major version.

tnowotny (approved):

PR #478, opened 2021-11-11
Proper implementation of kernel weights

neworderofjamie:
Previously, the way kernel weights were implemented was a bit of hack. You used procedural connectivity with a special variable initialisation snippet which pulled through the weights from an extra global parameter. This has a number of disadvantages:
* As the kernels are extra global parameters, GeNN has no idea how large they should be
* There's no way to support learning in this configuration as, from the point of view of the weight update model, the weights are produced by an opaque initialisation snippet.
* You can't take advantage of standard initialisation syntax for the kernels as they are extra global parameters.
* If you wanted to do something where different kernels were used across batches, you'd need to figure out the structure in memory yourself within a custom variable initialisation snippet.

This all stands in the way of both learning in convolutions and implementing a more efficient convolution implementation. This change hopefully solves all these problems via a new ``PROCEDURAL_KERNELG`` matrix mode. When this mode is used, all weight update model variables are allocated to the kernel size specified by the sparse connectivity initialisation model and can be initialised using the full functionality of the variable initialisation system. Only downside is that **slightly** different code paths need to be used for initialising sparse connectivity using a kernel and using the kernel directly (https://github.com/genn-team/ml_genn/commit/03fda90321e5942465fe08ce8a2ace6410996c55)


src/genn/backends/opencl/backend.cc
```diff
@@ -1145,14 +1149,16 @@ void Backend::genInit(CodeStream &os, const ModelSpecMerged &modelMerged,
     const bool anyCustomUpdateInitGroups = !modelMerged.getMergedCustomUpdateInitGroups().empty();
     const bool anyCustomWUUpdateDenseInitGroups = !modelMerged.getMergedCustomWUUpdateDenseInitGroups().empty();
     const bool anyDenseInitGroups = !modelMerged.getMergedSynapseDenseInitGroups().empty();
+    const bool anyKerneInitGroups = !modelMerged.getMergedSynapseKernelInitGroups().empty();
     const bool anyConnectivityInitGroups = !modelMerged.getMergedSynapseConnectivityInitGroups().empty();
     genMergedGroupKernelParams(initializeKernels, modelMerged.getMergedNeuronInitGroups(), 
-                               anyCustomUpdateInitGroups || anyCustomWUUpdateDenseInitGroups || anyDenseInitGroups || anyConnectivityInitGroups || additionalParamsRequired);
+                               anyCustomUpdateInitGroups || anyCustomWUUpdateDenseInitGroups || anyDenseInitGroups || anyKerneInitGroups || anyConnectivityInitGroups || additionalParamsRequired);
```
tnowotny, line 1155: should `anyKerneInitGroups` be `anyKernelInitGroups`?

src/genn/backends/opencl/backend.cc
```diff
@@ -1145,14 +1149,16 @@ void Backend::genInit(CodeStream &os, const ModelSpecMerged &modelMerged,
...
     genMergedGroupKernelParams(initializeKernels, modelMerged.getMergedCustomUpdateInitGroups(), 
-                               anyCustomWUUpdateDenseInitGroups || anyDenseInitGroups || anyConnectivityInitGroups || additionalParamsRequired);
+                               anyCustomWUUpdateDenseInitGroups || anyDenseInitGroups || anyKerneInitGroups || anyConnectivityInitGroups || additionalParamsRequired);
```
tnowotny, line 1157: here, too

src/genn/backends/opencl/backend.cc
```diff
@@ -1145,14 +1149,16 @@ void Backend::genInit(CodeStream &os, const ModelSpecMerged &modelMerged,
...
     genMergedGroupKernelParams(initializeKernels, modelMerged.getMergedCustomWUUpdateDenseInitGroups(), 
-                               anyDenseInitGroups || anyConnectivityInitGroups || additionalParamsRequired);
-    genMergedGroupKernelParams(initializeKernels, modelMerged.getMergedSynapseDenseInitGroups(), anyConnectivityInitGroups || additionalParamsRequired);
+                               anyDenseInitGroups || anyKerneInitGroups || anyConnectivityInitGroups || additionalParamsRequired);
```
tnowotny, line 1159: and here ...

tnowotny (changes requested):
Looking pretty good.
Probably better to correct the typo in the variable name (see comments).
What is the reason behind the empty files "skip_SingleThreadedCPU"?

neworderofjamie:
So the ``skip_BACKEND_NAME`` files are just used to tell the test runner that this feature test shouldn't be run on a given backend e.g. single-threaded CPU doesn't support procedural connectivity and there's a few CUDA-specific tests as well

tnowotny (approved):
all good.

PR #482, opened 2021-11-16
References cf. values can be used in a couple of places

alexdewar:
My compiler warned me about the first of these. There might be a few others in this function but I didn't look too hard.


neworderofjamie (approved):
Good spot! thank you

PR #484, opened 2021-11-18
Toeplitz KernelG

neworderofjamie:
Building on #478, this adds the ``TOEPLITZ_KERNELG`` matrix type which implements convolution-adjacent types of connectivity using a doubly-blocked Toeplitz approach. Essentially this PR involves:

- A new snippet ``CodeGenerator::InitToeplitzConnectivitySnippet::Base`` snippet type containing the required code strings, state variables etc and ``CodeGenerator::InitToeplitzConnectivitySnippet::Init`` wrapper type for bundling snippets with required parameters. I don't really see users creating too many of these but it's a much nicer way of implementing the connectivity than adding loads of special stuff to ``SynapseGroup``.
- Yet more overrides for ``ModelSpec::addSynapsePopulation`` which take a ``CodeGenerator::InitToeplitzConnectivitySnippet::Init`` object as the last parameter
- In code generator (where it won't break backward compatibility) have generally tried to rename ``ConnectivityInitialiser`` to ``SparseConnectivityInitialiser`` to make clear separation with new ``ToeplitzConnectivityInitialiser``.
- New ``CodeGenerator::PresynapticUpdateStrategySIMT::PostSpanToeplitz`` with first-pass implementation (no separating of input channels)

At optimal batch-size, performance is approximately 3x faster than ``PROCEDURAL_KERNELG`` on VGG16 models (both CIFAR10 scale and ImageNet). However, at ImageNet scale, this equates to 7 hours rather than 24 hours to performance inference over the whole validation set so still pretty slow. Because there are only ``kernel width * kernel height * kernel out channels`` threads launched for each synapse group, this currently (splitting spikes and getting a factor of ``kernel input channels`` more threads is likely to solve this) performs less well than ``PROCEDURAL_KERNELG`` with batch size 1.

Figure_1
## TODO
- [x] Unit tests
- [x] CPU implementation (unlike procedural, I think this should have one)
- [ ] Add support for striding
- [ ] Profiling
- [ ] Documentation


tnowotny (approved):
After thinking things through & helpful discussion I believe I understand the scheme and the central calculations for the "on-demand" synaptic connections are right; including a cursory inspection of corner cases for different padding which seem consistent with TF conventions. I think this can be merged.

PR #489, opened 2021-11-27
Fixed for strided convolutions

neworderofjamie:
Two fixes here:
* We weren't calling ``ensureFtype`` on code used to initialise state variables e.g. the row-build state variables for procedural connectivity. In cases where all calculation couldn't be optimised out, this could lead to double-precision maths happening.
* When calculating convolutions with stride, you perform a calculation like ``1 + ((inRow - kernelSize) / stride)`` to obtain the lower bound of output rows to loop through. That division **should** be rounding towards -infinity rather than towards zero to handle cases like ``inRow=1``, ``kernelSize=3``, ``stride=2`` where you want to be starting at row 0 but the code was incorrectly giving 1. I have fixed by moving the calculation to floating point and using floor.

This is cherry-picked from the toeplitz_kernelg branch as it's not really related

### DIFF
--- a/include/genn/genn/initSparseConnectivitySnippet.h
+++ b/include/genn/genn/initSparseConnectivitySnippet.h
@@ -415,9 +415,9 @@ public:
     SET_ROW_BUILD_STATE_VARS({{"inRow", "int", "($(id_pre) / (int)$(conv_ic)) / (int)$(conv_iw)"},
                               {"inCol", "int", "($(id_pre) / (int)$(conv_ic)) % (int)$(conv_iw)"},
                               {"inChan", "int", "$(id_pre) % (int)$(conv_ic)"},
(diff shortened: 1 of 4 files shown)


tnowotny (approved):
Ha ... tricky business with that rounding issue but looks like the right way to fix it. I suspect not less fast if not faster than before.

jamesturner246 (commented):
The Conv2D fix solves the issue, and no adverse effects. 

I just thought I should try a test for AvePool2DConv2D as well, and I think it will also need the same fix. I just pushed the avepool conv test to mlgenn: procedural_kernelg, in any case.

neworderofjamie:
AvgPool will indeed need the same fix in ml genn (although I think average pooling followed by a conv with stride would be decidedly odd). Agreed, this is likely to be quicker.

PR #491, opened 2021-12-10
Another convolution fix

neworderofjamie:
We have been calculating max row length for building sparse matrices from Conv2D connectivity by multiplying the number of output channels by ones of these terms for each dimension (where X is unknown):
```
floor((X + pad) / stride) - floor((X + pad - kern_size) / stride)
```
Previously I had simplified this to:
```
floor(kern_size / stride)
```
But, I think this is correct:
```
ceil(kern_size / stride)
```
The cases that were crashing now don't but I'm not quite sure what the floor/ceiling function relation 'proves' it.....

### DIFF
--- a/include/genn/genn/initSparseConnectivitySnippet.h
+++ b/include/genn/genn/initSparseConnectivitySnippet.h
@@ -441,12 +441,12 @@ public:
     SET_CALC_MAX_ROW_LENGTH_FUNC(
         [](unsigned int, unsigned int, const std::vector<double> &pars)
         {
-            const unsigned int conv_kh = (unsigned int)pars[0];
-            const unsigned int conv_kw = (unsigned int)pars[1];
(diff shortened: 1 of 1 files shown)


tnowotny (approved):
As @neworderofjamie explained, this comes down to whether 
floor((x+p)/s) - floor((x+p-k)/s) <= ceil(k/s)
where X as stated, p pad_size, k is kern_size and s is stride.
I believe the following argument is a mathematical proof for this:
\[
floor((x+p)/s) - floor((x+p-k)/s) 
= floor((x+p)/s) + ceil(-(x+p)/s + k/s)
<= floor((x+p)/s) + ceil(-(x+p)/s) + ceil(k/s)
= floor((x+p)/s) - floor((x+p)/s) +ceil(k/s)
= ceil(k/s) 
\]
So, based on this I think the PR can be approved.

PR #493, opened 2022-01-09
Named parameters

neworderofjamie:
So, this is the first of a bunch of refactoring-like changes that break compatibility (hence this is being merged into a new ``genn_5`` development branch) but I think make things better going forwards. My general aim is going to be to keep PyGeNN as backward compatible as possible but, break C++ models fairly drastically (I suspect we don't have a lot of C++ users these days).

This PR changes how parameters and variables are configured internally and from the C++ API. Currently you define a model in C++ something like:
```c++
class Recurrent : public NeuronModels::Base
{
public:
    DECLARE_MODEL(Recurrent, 3, 3);

    SET_PARAM_NAMES({
        "TauM",         // Membrane time constant [ms]
        "Vthresh",      // Spiking threshold [mV]
        "TauRefrac"});  // Refractory time constant [ms]

    SET_VARS({{"V", "scalar"}, {"RefracTime", "scalar"}, {"E", "scalar"}});
}
```
where the ``DECLARE_MODEL`` macro does some boilerplate which then allows you to define a model instances parameters like this (if you're being nice and commenting things):
```c++
Recurrent::ParamValues recurrentParamVals(
        20.0,   // Membrane time constant [ms]
        0.61,   // Spiking threshold [mV]
        5.0);   // Refractory time constant [ms]
Recurrent::VarValues recurrentInitVals(
        0.0,    // V
        0.0,    // RefracTime
        0.0);   // E
```

My _intention_ when I first implemented this five years ago was to enable the number of parameters being passed to a model to be checked at compile-time making it harder to make stupid mistakes. However:

1. The compiler errors that result are incomprehensible
2. Building models dynamically requires various hacks on top (see [SpineML ](https://github.com/genn-team/genn/blob/master/include/spineml/generator/modelCommon.h#L38-L110 )and [PyGeNN](https://github.com/genn-team/genn/blob/master/pygenn/genn_wrapper/include/customParamValues.h#L5-L38)) so that some sort of dynamic container for parameters is used in place of the fixed-size ``ParamValues`` containers
3. In order to expose built in models to Python, each of the specialised overrides of ``ModelSpec::addNeuronPopulation``, ``ModelSpec::addSynapsePopulation``, etc need to be wrapped individually - currently this involves a whole load of nasty searching of header files for ``DECLARE_MODEL`` macros and generating SWIG code.

In this PR, we switch from using ``std::vector<double>`` to ``std::unordered_map<std::string, double>`` throughout GeNN for representing parameter values. Now, the previous C++ example looks more like:
```c++
class Recurrent : public NeuronModels::Base
{
public:
    DECLARE_SNIPPET(Recurrent);

    SET_PARAM_NAMES({
        "TauM",         // Membrane time constant [ms]
        "Vthresh",      // Spiking threshold [mV]
        "TauRefrac"});  // Refractory time constant [ms]

    SET_VARS({{"V", "scalar"}, {"RefracTime", "scalar"}, {"E", "scalar"}});
}
...
ParamValues recurrentParamVals{{"TauM", 20.0}, {"Vthresh", 0.61}, {"TauRefrac", 5.0}};

VarValues recurrentInitVals{{"V", 0.0}, {"RefracTime", 0.0}, {"E", 0.0}};
```
This has the advantages of:

1. Being easier for the user - parameters don't need to be specified in order and the code is much more self-documenting
2. Because the parameters are validated at runtime, we can give better errors indicating what variable and population is the cause of the problem
3. There are still templated helper versions of ``ModelSpec::addNeuronPopulation`` etc but, these now do very little aside from pull out built in models - all the heavy-lifting and hence all we need to wrap in python is a single un-templated version
4. This is pretty much what PyGeNN and SpineML already does so we're just moving common logic downwards.
5. I think default parameter values would be a sensible feature at some point but wouldn't have been possible with the old syntax
6. Internally, this has actually tidied up some amount of nastiness (474 more deletions than additions if you believe github)
7. Having to select parameters by index in derived parameter-like code is super-error prone (so many bugs in the convolution work have been due to this). You can now do `` {"ip0", [](const ParamValues &pars, double){ return pars.at("Vspike") * pars.at("Vspike") * pars.at("alpha"); }}`` which is a little verbose but definitely clearer.

If you wanted to make a new C++ model using this branch, it _should_ all work but my plan is to try and use the ``genn_5`` development branch as a way to split up big refactoring type changes rather than keep it in a usable state. Therefore, while I've updated all the unit tests to use the new syntax and added a new one to cover the error handling, I _haven't_ updated PyGeNN to work with this (because PyBind11 is waiting in the wings), updated the feature tests (because I intend to port them all to PyGeNN) or updated the user projects (because exactly what we want to do with them remains a somewhat open question in my mind)


tnowotny (approved):
Yes, I like this. Especially as it brings everything in line with how things are already done in Python and other than breaking backwards compatibility there does not seem a downside to it. We can talk about the user projects another time. It is probably time to do something more systematic with them; we can pickle some in their current state in their own repo and translate some others into python as "tutorials/examples" - but also different repo I suspect.

PR #494, opened 2022-01-10
Check event code for den delay

neworderofjamie:
Until Anindya tried, clearly no one has tried using heterogeneous delays and spike-like events - the event code was never getting searched for ``addToInSynDelay`` so code wasn't compiling. Also added a unit test


tnowotny (approved):
oh boy ... you do need users to find all the nasties ... looks like a straightforward fix, thanks.

PR #495, opened 2022-01-13
Group names missing from hash used for comparing models

neworderofjamie:
This means if you rename a population, the hash doesn't change and the code won't rebuild (unless you force it with ``genn-buildmodel -f`` or equivalent). This PR solves this and adds unit tests to cover this.


tnowotny (approved):
Ok.

PR #496, opened 2022-01-14
Switch to pyBind11

### LINKED ISSUE #310: python setup.py install is broken
You currently need to run this twice to get a functioning build (as Jenkinsfile currently does). The reason is that the ``install`` command first runs ``egg_info``, then ``install_lib``, which runs ``build_py`` then ``build_ext``. ``build_py`` copies all Python sources into build directory (including those built by SWIG) and ``build_ext`` runs SWIG so, if you build in an existing location, bits of the build are out-dated and, if you build in a new location, bits of build are missing.

[Stackoverflow](https://stackoverflow.com/questions/12491328/python-distutils-not-include-the-swig-generated-module) suggests a fix which re-orders the order of ``install`` sub-commands so ``build_ext`` is done first. However that means that SWIG without libgenn (and, I'm guessing, other things) having been copied into the build directory meaning that it fails.

**Note** to reproduce this issue, you need a fresh checkout of GeNN 
**Note** @agarwalmanvi I have deleted our conversation as it was mostly about the confusing of this being in the wrong repository!

### LINKED ISSUE #450: More information from ``SWIG::DirectorMethodException``
These get thrown (and since #433 caught!) when errors happen in Python callbacks called from C++ using SWIG directors e.g. derived parameter calculation code. However, the code generated by SWIG to throw the exceptions looks like:

```c++
PyObject *error = PyErr_Occurred();
if (error) {
  Swig::DirectorMethodException::raise("Error detected when calling 'DerivedParamFunc.__call__'");
}
```

So the _contents_ of the error i.e. the message is totally ignored (meaning that, until I dug into this I was worried something was really wrong with this code but infact I had just forgotten to ``import numpy``...). Extra code can be monkey-patched into the exception throwing code with (from google):

```c++
%feature("director:except") {
    if( $error != NULL ) {
      PyObject *ptype, *pvalue, *ptraceback;
      PyErr_Fetch( &ptype, &pvalue, &ptraceback );
      PyErr_Restore( ptype, pvalue, ptraceback );
      PyErr_Print();
      Py_Exit(1);
    }
  }
```
which prints out a standard Python error message. However I'm not really sure what all the ``PyErr_Fetch`` and ``PyErr_Restore`` business does and I'd rather pipe the errors back into an exception message or something but, that is definitely beyond my Python C API understanding (@alexdewar any thoughts as a more experienced Python C API wrangler?)

### LINKED ISSUE #490: SWIG strips away type safety on typed enums
SWIG converts all typed enums to integers losing type safety. e.g. ``VarAccess_READ_WRITE`` is exported as follows:
```c++
  SWIG_Python_SetConstant(d, "VarAccess_READ_WRITE",SWIG_From_int(static_cast< int >(VarAccess::READ_WRITE)));
```
This makes it very easy to use the wrong type of enum in the wrong place. The "renaming python" approach described in https://stackoverflow.com/a/58841057 could maybe be modified although, perhaps, this is worth leaving until we switch to PyBind11.

neworderofjamie:
This has been something of an engineering effort but I'm pretty happy with the result. In the abstract, the main difference is that SWIG will by default try and wrap **everything** whereas with pybind11 you have to explicitly specify what you want to wrap. While this does involve writing a reasonable amount of boilerplate code (mostly in [pybind11/genn.cc](https://github.com/genn-team/genn/blob/pybind11/pygenn/src/genn.cc) which is around 1000 lines), because you have enough control to make the wrapped code much more Pythonic it doesn't then need to wrapped in _another_ layer of Python and the boilerplate is all in normalish C++ not the weird SWIG language. Overall, github thinks this results in a saving of about 4600 lines of code.

Main advantages to the PyGeNN user are:
- The additional layer of cruft between Python lambda functions and C++ ``std::function`` isn't required so derived parameters can be defined as ``("updateTimesteps", lambda pars, dt: pars["updateTime"] / dt)`` rather than ``("updateTimesteps", genn_model.create_dpf_class(lambda pars, dt: pars[0] / dt)())`` (also using the dictionary syntax from #493)
- Previously, for some lower-level operations, you had to interact with the underlying GeNN objects via e.g. ``model._model.set_default_narrow_sparse_ind_enabled(True)``. You now interact directly with the GeNN objects to which we dynamically 'mix in' an additional Python base classes (https://github.com/genn-team/genn/blob/pybind11/pygenn/genn_model.py#L96-L100) to make things a little friendlier
- The extension builds many times faster because we only wrap what's required and issues like #310 are fixed because there's not the additional 'swigging' phase which setuptools seems to only half understand
- Enums are wrapped in a type-safe manner so issues like #490 can't occur

I've tested this with the microcircuit model and the superspike demo (as well as some simpler tests) and it seems fully functional but, in a subsequent PR, I am going to port all the feature tests to PyGeNN and tidy up some bits and pieces e.g. documentation and improvements to the build process.

Fixes #490, fixes #450 and fixes #310


tnowotny:
Before I try to review anything - Wow, what a fast effort on this non-trivial move. Amazing!

neworderofjamie:
Thank you 😄 was on a bit of a coding roll last week (if you're reviewing, you should probably review #493 first by the way)

pygenn/genn_model.py
```diff
@@ -112,221 +124,154 @@ def __init__(self, precision="float", model_name="GeNNModel",
...
+            "char":             np.int8,
+            "unsigned char":    np.uint8,
+            "uint64_t":         np.uint64,
+            "int64_t":          np.int64,
+            "uint32_t":         np.uint32,
+            "int32_t":          np.int32,
+            "uint16_t":         np.uint16,
+            "int16_t":          np.int16,
+            "uint8_t":          np.uint8,
+            "int8_t":           np.int8}
```
tnowotny, line 191: any reason why not to include bool while wee are at it?

tnowotny (approved):
Looks pretty tidy and based on your description I don't see any red flags.
Any gnarly details will probably only come to light if/when we are using this in anger.
I think there is not much more I can do in terms of review at this stage, so will click approve.

PR #498, opened 2022-01-25
Binomial distribution

neworderofjamie:
Added ``$(gennrand_binomial)`` non-uniform RNG, ``InitVarSnippet::Binomial`` if you ever need to initialise variables with binomial distribution and extended VarInit feature test


tnowotny (approved):
looks good ... I am sure you adopted the code for the binomial right in particular if the test passed.

PR #499, opened 2022-01-26
Missing PyGenn toeplitz

neworderofjamie:
Yet another branch, left behind after the pre-Christmas flurry of ANN conversation...This fixes two issues:
- Totally forgot to expose defining custom Toeplitz models to PyGeNN
- With pooling, you can easily hit the code path where the target population is small enough to fit in shared memory and...this was  broken


tnowotny (approved):
Ok.

PR #502, opened 2022-02-18
Fixed a bug in the spike recording system 

tnowotny:
… where for particular kernel…size and neuron population size combinations too many words were written into the spike recording buffer causing a memory error (but only discovered by compute-sanitize).


neworderofjamie (approved):
This looks good - I had to push some tab->space changes though so could you check your text editor config?

PR #505, opened 2022-02-25
Genn 5 deprecations

neworderofjamie:
Before moving onwards I wanted to deprecate a few things:
1. This should have been deprecated a long time ago but, with ``GLOBALG``, the default was to also attempt to optimise out postsynaptic model variables. To prevent this I had added various ``INDIVIDUAL_PSM`` matrix modes but I can't really think of any situation where this is a good idea
2. The first proto-batching system we added (#323) turned out to be a less than great idea performance wise and required lots of extra logic throughout the code and tests


tnowotny (approved):
I agree with those deprecations. I wonder whether any of this has actually been used.

tnowotny:
Obviously, please make sure the tests pass ;-)

PR #507, opened 2022-03-17
Lookup data structures not taking batching into account

neworderofjamie:
I've been trying to use custom updates to do more of the resetting style stuff in mlGeNN and everything was breaking as batch size increased. Turns out this was totally broken as, for custom updates targetting neurons, the indices within which threads binary search to find their population were generated as if batch size = 1.

None of your models use custom updates on duplicated (i.e. batched) neuron group variables do they @tnowotny?


tnowotny (approved):
Looks good - cannot say yet whether it makes a difference for eventprop. Potentially not (would have been too convenient)

PR #508, opened 2022-03-28
Fix hashing bugs

neworderofjamie:
When calculating the hashes used to merge groups together for **initialization**, _usually_ all that we care about is the names and types of the state variables and how they are initialised _but_ with SIMT backends where additional data structures are required for synapse dynamics/postsynaptic updates this is not enough. This fix simply includes whether weight update models include synapse dynamics and postsynaptic update code in the hash used to determine mergeability - this ignores the backend specific-details so will sometimes split initialization unnecessarily on CPU but fixes this (nasty) issue.

Furthermore, while trying to write a unit test to test this, I found a subtler issue that could also potentially cause issues. Essentially hashing an empty container (``std::vector``, ``std::string`` etc) had no effect on the hash so the following code:
 ```cpp
 Utils::updateHash("", hash);
 Utils::updateHash("test", hash);
 ```
generates the same hash as:
 ```cpp
 Utils::updateHash("test", hash);
 ```

I've fixed this by hashing the size of the container before hashing the contents (which, one way or another, involves looping through the bytes of the contents). This PR fixes both issues.

### DIFF
--- a/include/genn/genn/gennUtils.h
+++ b/include/genn/genn/gennUtils.h
@@ -123,23 +123,25 @@ inline std::string writePreciseString(T value)
     return s.str();
 }
 
+//! Hash arithmetic types and enums
+template<typename T, typename std::enable_if<std::is_arithmetic<T>::value || std::is_enum<T>::value>::type* = nullptr>
+inline void updateHash(const T& value, boost::uuids::detail::sha1& hash)
+{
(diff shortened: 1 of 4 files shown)


tnowotny (approved):
Yes, makes sense. Nasty issues those ...

PR #509, opened 2022-04-14
Fix batch init

neworderofjamie:
When implementing batching I wrote a helper function called ``genVariableFill`` for initialising variables, taking account of delays and batching. However, for some reason, I forgot to actually _use_ this when initializing ``inSyn``, ``denDelay`` (which was subsequently copy-pasted into the initialisation of ``revInSyn``). CUDA seems to zero newly allocated memory so, generally, this has minimal effect but, if you _reinitialise_ your model, you may be left with large residual neural input currents in some batch instances (which, if you're using dendritic delays will continue to appear for max dendritic delay timesteps!). 

This finally fixes the issues we were seeing in Felix Schmitt's model - big thanks to him for testing my various speculative fixes and suggesting that (re-)initialization could be the problem!


tnowotny (approved):
Looks good (I haven't checked what the genVariableFill function does, have to trust you on that to not descend too deeply in limited time)

PR #510, opened 2022-04-14
Fix incompatibility with older Python versions

neworderofjamie:
In #500, we started using [``platform.uname``](https://docs.python.org/3/library/platform.html#platform.uname) to get platform information and thus detect WSL. However, in Python < 3.3, this returns a tuple (rather than a namedtuple) - as GeNN 4 still technically supports Python 2.7, handle that too


tnowotny (approved):
sure ... fine.

PR #511, opened 2022-04-14
Replace synRemap mechanism with much simpler one

neworderofjamie:
In older versions of GeNN which used a plain CSR matrix format for sparse connectivity, figuring out what presynaptic and postsynaptic neuron a synapse dynamics kernel thread was associated with, required a data structure called ``indInG``.  When we switched to the new 'ragged' format, this was no longer necessary but, for _some_ reason, I introduced a new (``synRemap``) lookup structure which maps contiguous (0, numSynapses] thread indices to indices into the ragged matrix. This adds another 32-bits of memory per sparse synapse, significantly slows down synapse dynamics as each synapse has to read the 32-bit from global memory **and** makes doing structural plasticity much harder than it needs to be as you need to keep the ``synRemap`` structure up to date with all the other bits of synapse state. Theoretically this _could_ have saved a few idle threads but, it doesn't because GeNN has no way of knowing how many synapses there are at compile time so it launches  ``sg.getSrcNeuronGroup()->getNumNeurons() * sg.getMaxConnections()`` anyway. The ``synRemap`` structure is an internal detail which users shouldn't ever be messing with so I have totally removed it and, in this PR, just divide the same number of threads into pre and post indices with / and % respectively (duh). On my machine this reduces the time spent in synapse dynamics kernels by around 25%!

### DIFF
--- a/include/genn/backends/single_threaded_cpu/backend.h
+++ b/include/genn/backends/single_threaded_cpu/backend.h
@@ -141,7 +141,6 @@ public:
     //! Different backends seed RNGs in different ways. Does this one initialise population RNGS on device?
     virtual bool isPopulationRNGInitialisedOnDevice() const override { return false; }
 
-    virtual bool isSynRemapRequired(const SynapseGroupInternal&) const override{ return false; }
     virtual bool isPostsynapticRemapRequired() const override{ return true; }
 
     //! Backends which support batch-parallelism might require an additional host reduction phase after reduction kernels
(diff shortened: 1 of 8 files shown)


tnowotny (approved):
Ha ... interesting one. I guess the key is that the launched kernel grid is indeed full size (so to speak) so all the wizardry of remap completely unnecessary ... 
Nice simplification.

PR #512, opened 2022-04-22
Reduce unnecessary strictness in configuration of custom updates

neworderofjamie:
When you built a model with custom updates, GeNN was previously a bit over-zealous is enforcing ``SHARED`` vs ``DUPLICATE`` variable settings even if the model wasn't batched. This PR moves this logic into finalize where batch size is known so the checks can be dis-regarded if batch size is 1. This enables custom updates to _actually_ know whether they're batched rather than just whether they would be if batch size > 1 and allows a bit of tidying.


tnowotny (approved):
yep ... makes sense.

PR #515, opened 2022-04-28
Fix max connections issues

neworderofjamie:
This fixes two issues:
1. Even if you've used a sparse connectivity initialisation snippet to initialise connectivity, as long as you're _increasing_ it, you should still be able to manually increase max connections e.g. to leave a bit more padding at the end of each row for rewiring
2. Max connections wasn't being included in hashes used to determine if model had changed so if you had build a model based on some manually specified connectivity and changed it you ended up in a really broken state


tnowotny (approved):
Looks reasonable ...

PR #518, opened 2022-04-28
Mac debugging with LLDB

### LINKED ISSUE #57: Debugging on Mac
On Mac, debugging doesn't necessarily work as gdb is not automatically on the path. So far GeNN is just aborting when trying to debug - is there somethign more clean that we can do?

neworderofjamie:
While investigating #517, I wanted to use the debugger but no one had been tempted by the "good first issue" to fix #57. I now have and GeNN users on the Mac can now debug code generator bugs to their heart's content :D (this also must win some sort of "oldest bug fixed" prize as it was first raised in 2015!)

Fixes #57


tnowotny (approved):
haha ... the fix isn't that difficult or elaborate - suppose we could have done this a little bit earlier ;-)

PR #520, opened 2022-05-03
Fix rev in syn

### LINKED ISSUE #519: Problems with revInSyn and merged synapse init groups
The ```m_FusedPreOutputSuffix``` for ```revInsyn``` was not initialised upon synapse group creation. Therefore, in runner.cc (and throughout) we get
```C++
float* revInSyn;
float* d_revInSyn;
```
instead of 
```C++
float* revInSyn_NAME;
float* d_revInSyn_NAME;
```
where ```NAME``` is the name of the synapse population (or fused name).

After fixing this (branch ```fix_revInSyn```) a new issue occurred where synapse groups were combined in a ```MergedSynapseDenseInitGroup``` where some of the synapses have ```revInSyn``` and others do not. This then leads to the creation of e.g. 
```C++
pushMergedSynapseDenseInitGroup0ToDevice(2, d_revInSynin_to_hid, d_win_to_hid, d_dwin_to_hid, 256, 700, 256, 2.00000000000000004e-02f, 1.00000000000000002e-02f);
```
for a synapse population ```in_to_hid``` that does not have revInsyn and hence ```d_revInSynin_to_hid``` does not exist.
I suspect there must be some test for mergeability of initGroups where the existence of revInSyn is not considered but should be. Unfortunately, I struggle a little to dig this deep.

neworderofjamie:
Fixes #519

### DIFF
--- a/src/genn/genn/code_generator/groupMerged.cc
+++ b/src/genn/genn/code_generator/groupMerged.cc
@@ -944,10 +944,6 @@ SynapseGroupMergedBase::SynapseGroupMergedBase(size_t index, const std::string &
             addPSPointerField(precision, "inSyn", backend.getDeviceVarPrefix() + "inSyn");
         }
     }
-    // for all types of roles
-    if(getArchetype().isPresynapticOutputRequired()) {
-      addPreOutputPointerField(precision, "revInSyn", backend.getDeviceVarPrefix() + "revInSyn");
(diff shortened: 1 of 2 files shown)


tnowotny (approved):
It seems to work well and makes sense I think. Thanks for sorting that out.

PR #523, opened 2022-05-17
Make sure $(batch) is set to zero on CPU

### LINKED ISSUE #513: $(batch) not present on SingleThreadedCPU backend
This is very annoying as the same models can't be used across backends - this should just be substituted for zero.

neworderofjamie:
Code using ``$(batch)`` previously didn't work on the single-threaded CPU backend. This PR fixes this by substituting zero for ``$(batch)`` in all CPU kernels

Fixes #513


tnowotny:
... does this imply that models with ```model.batch_size = 1``` will now run on the CPU backend (or where batch_size is not set by the user)?

neworderofjamie:
If the reason they didn't run before was because you referenced ``$(batch)`` in your code and it didn't compile, yes they will!

PR #524, opened 2022-05-19
Custom updates on KERNEL variables

neworderofjamie:
I **think** this is the missing piece to enable EventProp on convolutional networks - it allows custom updates to be attached to synapse groups with ``KERNEL`` connectivity. This involves:

1. Custom updates can update kernel entries (e.g. for weight decay in convolutional networks)
2. Reductions from ``DUPLICATE`` into ``SHARED`` kernel variables for batched models
3. Initialisation of variables in custom updates attached to a synapse group with `KERNEL` connectivity (which will themselves be kernels)

The feature tests for all three scenarios have been expanded to cover the new cases. The process of implementing this was also *so* annoying and error prone, I did a bit of tidying (which _almost_  allowed me to delete as much code as I added :smile: )


tnowotny (approved):
As so often, I can't fully digest the details but most certainly the implemented features are welcome/important - thanks Jamie.

Issue #537, opened 2022-08-30
Duplication between github readme files and doxygen content


neworderofjamie:
This is the case both for:
- GeNN installation - https://github.com/genn-team/genn/blob/master/README.md vs https://genn-team.github.io/genn/documentation/4/html/d8/d99/Installation.html and https://genn-team.github.io/genn/documentation/4/html/d7/d98/Quickstart.html
-  PyGeNN installation - https://github.com/genn-team/genn/blob/master/pygenn/README.md vs https://genn-team.github.io/genn/documentation/4/html/d8/d99/Installation.html#pygenn

Not 100% sure what the best solution is but, one option would be to do the following:

1. Convert [Installation ](https://github.com/genn-team/genn/blob/master/doxygen/01_Installation.dox) and [Quickstart ](https://github.com/genn-team/genn/blob/master/doxygen/02_Quickstart.dox) RST source to markdown (Doxygen has supported markdown as well as RST for [some time](https://doxygen.nl/manual/markdown.html)) 
2. Write a little script to concatenate together the markdown files into new readmes (we could trigger this from actions like https://stackoverflow.com/questions/35080160/github-include-md-files-in-readme-md)

PR #538, opened 2022-08-30
Encourage use of Windows path GUI

neworderofjamie:
Based on many users falling into this trap, I replaced suggestions to use ``setx`` with instructions to use the Windows environment editing GUI (which I at least prefer).

While I was at it, I also:
- Removed suggestions to use binary wheels - they just don't work well enough at the moment
- Suggestion to download master as a zip - either download a stable release or use git, don't download an unstable and more-difficult-to-update zip.

This stuff is also explained in the readmes but, as discussed in #537, these have to be updated separately so I will only do that when we're all happy with the changes.


doxygen/01_Installation.dox
```diff
@@ -80,24 +77,23 @@ repository.
     with each other. The latest C++ compiler need not necessarily be
     compatible with the latest CUDA toolkit.
 
-(vi) Set the `CUDA_PATH` variable if it is not already set by the
-    system, by putting
+(vi) GeNN uses the `CUDA_PATH` environment version to determine which
```
tnowotny, line 80: environment variable

doxygen/01_Installation.dox
```diff
@@ -127,7 +114,7 @@ Building PyGeNN is a little involved so we recommend installing a prebuilt binar
  \endcode
  
 \subsection pygenn_source_windows Installing PyGeNN from source on Windows
- - Ensure that you have at least Python 3.5 and Visual Studio 2015 installed (extensions for earlier versions of Python cannot be built using any versions of Visual Studio new enough to support C++11). If you are using Visual Studio 2019, you need at least Python 3.7.5. These instructions assume that the Anaconda platform was used to install Python, but it _should_ be possible to install PyGeNN using suitable versions of Python installed in different way (please let us know if you suceed in doing so!)
+ - Ensure that you have at least Python 3.5 and Visual Studio 2015 installed (extensions for earlier versions of Python cannot be built using any versions of Visual Studio new enough to compile GeNN). If you are using Visual Studio 2019, you need at least Python 3.7.5. These instructions assume that the Anaconda platform was used to install Python, but it _should_ be possible to install PyGeNN using suitable versions of Python installed in different way (please let us know if you succeed in doing so!)
```
tnowotny, line 117: installed in a different way

tnowotny (approved):
Looks good apart from the two typos.

tnowotny (approved):
Didn't spot any other typos and as said previously overall agree with all changes. Hopefully there will be less users struggling with the windows install.

PR #540, opened 2022-10-12
Fixed potential bug where RNG isn't instantiated

neworderofjamie:
If custom update variables were the only place in a model where an RNG was required for initialization, this wouldn't be detected and the model wouldn't compile. Have added extra unit tests and beefed up existing ones to test further up.


tnowotny (approved):
Looks sensible - nice catch.
Somehow I am having a deja-vu of having reviewed this one before but apparently not.

PR #547, opened 2022-10-20
Connectivity hashing bug

neworderofjamie:
If you changed the connectivity initialisation model, this wasn't being correctly picked up in the model hash and thus code wasn't correctly being regenerated. Added a fix and a new unit test


tnowotny (approved):
yep, looks good. One never runs out of bugs ...

Issue #546, opened 2022-10-19
Docker install of GeNN


neworderofjamie:
It would be great to be able to run GeNN models from a docker image. One of our users (@russelljjarvis) kindly started creating the following:

```dockerfile
FROM nvidia/cuda:11.8.0-devel-ubuntu22.04
WORKDIR /usr/local/cuda-11.8/
ENV CUDA_PATH=/usr/local/cuda-11.8 
RUN cuda-memcheck ./has_cuda ; exit 0
RUN nvcc; exit 0
RUN apt update; 
RUN apt install git build-essential lcov wget -y --no-install-recommends 
ENV CONDA_DIR /opt/conda
RUN wget --quiet https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh -O ~/miniconda.sh && \
     /bin/bash ~/miniconda.sh -b -p /opt/conda

ENV PATH=$CONDA_DIR/bin:$PATH
WORKDIR $HOME/jovyan
RUN $(which conda) install -c miniconda swig numpy -y
RUN git clone https://github.com/genn-team/genn.git
RUN mv genn GeNN
WORKDIR GeNN
ENV PATH=$PATH:$HOME/jovyan/GeNN/bin
RUN make DYNAMIC=1 LIBRARY_DIRECTORY=`pwd`/pygenn/genn_wrapper/
RUN python setup.py develop
WORKDIR $HOME/jovyan/GeNN/userproject/PotjansMicrocircuit_project
RUN make
RUN ./generate_run test 
WORKDIR $HOME/jovyan/GeNN/
RUN bash tests/run_tests.sh; exit 0
ENTRYPOINT ["/bin/bash"]
```

It fails at the step
```RUN ./generate_run test ```
but I think it is close.

Maybe I need a different Docker command for building with access to system cuda?

_Originally posted by @russelljjarvis in https://github.com/genn-team/genn/discussions/545#discussioncomment-3911329_

Personally I would like to see:
- [ ] A more flexible entry point that can also launch a jupyter notebook
- [ ] System Python rather than Anaconda (unless someone can convince me why conda improves anything)
- [ ] Figure out where the dockerfile should live (seperate genn-docker repository or in the genn repository) and get this building from Jenkins

russelljjarvis:
>  A more flexible entry point that can also launch a jupyter notebook
>  System Python rather than Anaconda (unless someone can convince me why conda improves anything)
[...]

PR #548, opened 2022-10-20
build: Dockerfile and docker-compose to launch notebook with pygenn resolves genn-team/genn#546

Stevinson:
Run pyGeNN models from a docker image. `docker-compose` is used to launch a jupyter notebook with files from the local filesystem.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,3 +73,7 @@ clean:
 .PHONY docker-build:
 docker-build:
 	@docker build -f ./build_steps/Dockerfile -t pygenn:latest .
+
+.PHONY docker-jupyter:
+docker-jupyter:
+	@docker run -p 127.0.0.1:8888:8888/tcp -v '$(CURDIR)'/pygenn/notebooks:/root/pygenn/notebooks pygenn jupyter notebook --ip 0.0.0.0 --no-browser
--- a/Makefile
+++ b/Makefile
@@ -69,3 +69,7 @@ clean:
 	@# Delete libGeNN
 	@rm -f $(LIBGENN)
 	@rm -f $(BACKEND_LIBS)
+
+.PHONY docker-build:
+docker-build:
+	@docker build -f ./build_steps/Dockerfile -t pygenn:latest .
--- a/build_steps/Dockerfile
+++ b/build_steps/Dockerfile
@@ -18,7 +18,9 @@ WORKDIR ${HOME}
 
 COPY . ${HOME}
 
-RUN adduser --disabled-password --gecos "" $USERNAME
-RUN chown -R ${USERNAME}:${USERNAME} "/root"
+RUN adduser --disabled-password --gecos "" $USERNAME && chown -R ${USERNAME}:${USERNAME} "/root"
+
 RUN make DYNAMIC=1 LIBRARY_DIRECTORY=${HOME}/pygenn/genn_wrapper/
 RUN python3 setup.py develop
+
+USER $USERNAME
--- a/build_steps/Dockerfile
+++ b/build_steps/Dockerfile
@@ -1,0 +1,24 @@
+FROM nvidia/cuda:11.5.0-devel-ubuntu20.04
+
+ARG USERNAME="pygennuser"
+ARG PYGENN_HOME="/root"
+
+RUN apt-get update && \
+    apt-get upgrade -y
+
+RUN apt-get install -yq --no-install-recommends curl vim python3-dev python3-pip
+
+RUN pip install swig numpy jupyter
+
+ENV CUDA_PATH=/usr/local/cuda-11.5
+ENV HOME=${PYGENN_HOME}
+ENV USERNAME=${USERNAME}
+
+WORKDIR ${HOME}
+
+COPY . ${HOME}
+
+RUN adduser --disabled-password --gecos "" $USERNAME
+RUN chown -R ${USERNAME}:${USERNAME} "/root"
+RUN make DYNAMIC=1 LIBRARY_DIRECTORY=${HOME}/pygenn/genn_wrapper/
+RUN python3 setup.py develop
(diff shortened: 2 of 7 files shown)


build_steps/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+ENV CUDA_PATH=/usr/local/cuda-11.5
+ENV HOME=${PYGENN_HOME}
+ENV USERNAME=${USERNAME}
+
+WORKDIR ${HOME}
+
+COPY . ${HOME}
+
+RUN adduser --disabled-password --gecos "" $USERNAME
```
neworderofjamie, line 21: Why not just use root as containers are kinda throwaway? My knowledge of docker is pretty crap so I could very well be missing something...

build_steps/docker-compose.yml
```diff
@@ -0,0 +1,28 @@
+version: "3.4"
```
neworderofjamie, line 1: I don't quite see why docker compose is required....I thought it was intended to combine together multiple containers whereas, couldn't what you do here be achieved with a single ``docker run`` incantation?

Also, more fundamentally, the way I imagine the GeNN docker image being used is that it would be hosted on Dockerhub (probably produced by CI when we make releases/merge master) so users wouldn't have to checkout our git repro and hence wouldn't have access to the docker compose script outside of the container.

Again, I may be missing things....

neworderofjamie (commented):
Thank you so much for your hard work on this - as a docker n00b, there's just a few things I'd like explained

build_steps/Dockerfile
Stevinson: It’s something I was taught was best practice so have got into the habit of doing. I believe this is because:
1. Certain execution environments block containers running as root
2. Force container immutability
3. Limits the processes that can be executed and who can execute them

build_steps/docker-compose.yml
Stevinson: Agreed, it facilitates having multiple containers interacting together. I only used it as I find it helpful to set up port bindings, volumes, health checks, etc in a configurable way that is easy to use (and was used only to launch jupyter).

However, if it is not going to be useful for extendability then maybe, as you say, a command such as `docker run -p 127.0.0.1:8888:8888/tcp -v '$(CURDIR)'/pygenn/notebooks:/root/pygenn/notebooks pygenn jupyter notebook --ip 0.0.0.0 --no-browser` is better suited. I have added this to the makefile (as `make docker-jupyter`) and can delete the `docker-compose.yml`?

build_steps/Dockerfile
neworderofjamie: Ahh, that is really good to know, thanks

build_steps/docker-compose.yml
neworderofjamie: Just for saving such brutally long command lines, docker compose seems like a good idea (especially as it's optional).

neworderofjamie (approved):
I am going to fiddle a bit more (hence the change of target) but, unless my colleague @jamesturner246  has any further concerns, I'm happy to accept this (and will add the relevant hacktoberfest label)

Makefile
```diff
@@ -69,3 +69,11 @@ clean:
 	@# Delete libGeNN
 	@rm -f $(LIBGENN)
 	@rm -f $(BACKEND_LIBS)
+
```
jamesturner246, line 72: It might be cool and convenient to have a makefile rule to setup an interactive pygenn container with  `-it`, and bind host `pwd` to some directory (e.g. 'workspace') in the working directory of the container. Just for people who don't want to use jupyter.

build_steps/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+ENV CUDA_PATH=/usr/local/cuda-11.5
+ENV HOME=${PYGENN_HOME}
+ENV USERNAME=${USERNAME}
+
+WORKDIR ${HOME}
+
+COPY . ${HOME}
+
+RUN adduser --disabled-password --gecos "" $USERNAME && chown -R ${USERNAME}:${USERNAME} "/root"
```
jamesturner246, line 21: Why not root user? What if the user wants to apt install additional programs within the container in an interactive session?

pygenn/README.md
```diff
@@ -13,3 +13,8 @@ PyGeNN wraps the C++ GeNN API using SWIG, allowing GeNN to be used either direct
  - Navigate to the GeNN directory and build GeNN as a dll using ``msbuild genn.sln /t:Build /p:Configuration=Release_DLL`` (if you don't have CUDA installed, building the CUDA backend will fail but it should still build the CPU backend).
  - Copy the newly built DLLs into pygenn using ``copy /Y lib\genn*Release_DLL.* pygenn\genn_wrapper``
  - Build the Python extension with setup tools using ``python setup.py develop`` command
+
```
jamesturner246, line 16: Should probably add pygenn/README.md documentation for the docker-jupyter option, plus for the interactive container rule mentioned above, if implemented.

jamesturner246 (changes requested):
It looks good. A couple of points that might be worth implementing - maybe @neworderofjamie also has preferences on them. Thanks!

build_steps/Dockerfile
jamesturner246: Running the container as root (from the host) could be bad, sure, but I can't see any issue being root *inside* the container. The mutability argument, I'm not sure how relevant that is, since all state is lost once the container is killed anyway. As for limiting processes, sure, it stops the user running e.g. `apt` inside the container as interactive, but is that a good thing?

jamesturner246: Btw, I don't mean to be pedantic. I'm no expert, and I'm very interested to hear yours and others' reasons for and against!

jamesturner246: Ahh.... I see the issue with containers starting as root now..... 

I just pwnt by box with a container that begins as root! :) 

https://flast101.github.io/docker-privesc/

PR #551, opened 2022-10-24
std::uniform_int_distribution operates on closed interval

neworderofjamie:
Serves me right for modernising....


tnowotny (approved):
Yep, always read the manual ;-)
I can believe this fixed the issue ...

PR #554, opened 2022-11-01
More compiler friendly error handling.

neworderofjamie:
Turns out that compiling the code generated by our ``CHECK_CUDA_ERRORS`` macro can cause runner.cc compilation to take a long time and a lot of memory. Switching this out of a simpler macro that just calls ``std::abort`` can reduce compile times by up to 10x on a model with 10000 state variables (100 neuron populations, nearly all-to-all connected):
fig_2_alternate_error
This simpler error handling macro can now be enabled from C++ with:
```c++
void modelDefinition(NNmodel &model)
{
    ...
    GENN_PREFERENCES.generateSimpleErrorHandling = true;
    ...
}
```
or from Python like:
```python
model = GeNNModel("float", "my_model", generateSimpleErrorHandling=True)
```

### DIFF
--- a/include/genn/backends/cuda/backend.h
+++ b/include/genn/backends/cuda/backend.h
@@ -78,7 +78,12 @@ struct Preferences : public PreferencesBase
 
     //! Generate corresponding NCCL batch reductions
     bool enableNCCLReductions = false;
-
+    
+    //! By default, GeNN generates CUDA error-handling code that generates exceptions with line numbers etc.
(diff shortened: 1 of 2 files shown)


tnowotny (approved):
Good work on this. For a moment I thought it might be a good idea to make the simple error handling the default but now I think you have it right as for small models it is not an issue and you would want the more informative output as a default option.

neworderofjamie:
Yeah, the longer term plan is to consider what error information is actually useful and maybe move some of the error handling out of the generated code

PR #555, opened 2022-11-07
Windows multi processor build

neworderofjamie:
No idea what made me realise this but, while we had been passing ``/m`` to ``msbuild`` which allows **projects** to be build in parallel if they have no dependencies, it doesn't actually enable parallel compilation of C++ files. This PR turns the actual option we want on for all projects and, on my laptop, reduces the time taken to fully rebuild GeNN and backends from 48 to 16s.

Sadly, it [seems NVCC doesn't support this option](https://stackoverflow.com/questions/17657488/cuda-parallel-code-generation-in-visual-studio) so this is only really going to help life for developers using Windows (i.e. me) rather than help speed up compilation of generated code on Windows.


tnowotny (approved):
Okaydoke ... love the spaghetti of options.

PR #557, opened 2022-11-10
Custom update merge fix

neworderofjamie:
In the course of my structural plasticity implementing, I found not one but two bugs in the variable reference/custom update stuff!

### Merging
In general, when merging together custom updates, what the variable references _target_ doesn't really matter - they're just pointers. However, whether variable references point to variables  with ``VarAccessDuplication::DUPLICATE``, ``VarAccessDuplication::SHARED`` etc changes the required index calculating code so this needs to be considered. This PR includes the variable duplication mode in the hashes used for merging groups and extends the unit tests to better cover corner cases in this area.

### References to custom update variables
Unlike synapses and neurons which are always batched if the model is, if custom updates are batched or not depends on whether they are attached to ``VarAccessDuplication::DUPLICATE`` variables (and whether the model is batched). This means if you make references to these variables from other custom updates whether these 'downstream' custom updates should be batched should depend on whether the upstream custom update was batched as well as the mode of the variables. This PR adds a ``isBatched`` method to each variable reference and extends the unit tests to cover this.

### DIFF
--- a/include/genn/genn/customUpdate.h
+++ b/include/genn/genn/customUpdate.h
@@ -152,7 +152,10 @@ protected:
         // If target of any variable references is duplicated, custom update should be batched
         if(batchSize > 1) {
             m_Batched = std::any_of(varRefs.cbegin(), varRefs.cend(),
-                                    [](const V& v) { return (v.getVar().access & VarAccessDuplication::DUPLICATE); });
+                                    [](const V& v) 
+                                    {
+                                        return (v.isBatched() && (v.getVar().access & VarAccessDuplication::DUPLICATE)); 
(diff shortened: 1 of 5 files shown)


tnowotny (approved):
Wow, it does get complicated with merging/ hashing/ custom updates ... but it looks like a sensible solution.

neworderofjamie:
Yeah - I feel maybe there might be some extra constraints that could be added which would simplify the whole system somewhat e.g. I'm pretty sure I haven't yet had a use-case for a custom update that simultaneously references variables in different populations of the size (although this is supported)

PR #559, opened 2022-11-17
Fixed bug in ``functionSubstitute``

neworderofjamie:
If you made a mistake in a code string where you didn't close a bracket on a function substitution e.g.
```c++
"$(for_each_synapse, printf(\"%d,\", $(id_post))"
```
The ``functionSubstitute`` function previously got stuck in an infinite loop. This fixes that issue and adds a couple of unit tests.


tnowotny (approved):
Looks like we are going more and more into the parser rabbit hole but that was to be expected...

neworderofjamie:
Yeah, although these adhoc parsers will all be replaced with **the** parser once I integrate it which should be more maintainable

PR #565, opened 2023-02-06
Batched prev spike time

### LINKED ISSUE #564: compilation fails when using batch sizes larger than 1 in combination with is_prev_post_spike_time_required=True
If the model contains a weight update class that uses
`is_prev_post_spike_time_required=True` and model.batch_size is set to any number larger than 1, building the model fails with the error `class "MergedNeuronPrevSpikeTimeUpdateGroup0" has no member "numNeurons"`
A code example (any value of `NBATCH` larger 1 produces the error):
```
import numpy as np
from pygenn.genn_model import (GeNNModel,
                               create_custom_weight_update_class,
                               init_var)

NBATCH = 10
R_TARG = 1.0

T = 100.0
DT = 0.1
NT = int(T/DT)

cs_wu = create_custom_weight_update_class(
        "cs_wu_class",
        param_names=["mu"],
        var_name_types=[("g", "scalar")],
        sim_code="$(addToInSyn, $(g));",
        learn_post_code=f"""
            $(g) += $(mu)*(
                {R_TARG}*($(t)-$(prev_sT_post))
                - 1.0
            );
        """,
        is_prev_post_spike_time_required=True
        )

model = GeNNModel("float", "test")
model.batch_size = NBATCH

poiss_pop = model.add_neuron_population(
        "poiss_pop",
        10, "Poisson",
        {"trefract": 0.0,
         "tspike":0.0,
         "Vspike":1.0,
         "Vrest":0.0},
        {"V":0.0, "spikeTime": 0.0}
        )

lif_pop = model.add_neuron_population(
        "lif_pop",
        10, "LIF",
        {"C": 1.0,
        "Ioffset": 0.0,
        "TauM": 1.0,
        "TauRefrac": 0.0,
        "Vreset": 0.0,
        "Vrest": 0.0,
        "Vthresh": 1.0},
        {"RefracTime": 0.0, "V": 0.0}
        )

syn_pop = model.add_synapse_population(
        "syn_pop", "DENSE_INDIVIDUALG", 0,
        "poiss_pop", "lif_pop",
        cs_wu, {"mu": 0.01},
        {"g": init_var("Uniform",{"min":0.0,"max":0.0})},
        {}, {},
        "DeltaCurr", {}, {})

model.build()

poiss_pop.set_extra_global_param("firingProb", np.random.rand(10))
poiss_pop.set_extra_global_param("offset", 0)

model.load()

for t in range(NT):
    model.step_time()

print("done")
```

neworderofjamie:
Two fixes here:
1. Missing ``numNeurons`` field in ``NeuronPrevSpikeTimeUpdateGroupMerged`` required when model has batch size > 1. Not 100% sure why batch size isn't readily available in this constructor but unconditionally including the number of neurons is fine
2. Indexing when updating previous spike times was incorrect when model is batched but there are no delays (both ``group->prevST`` and ``group->spk`` are ``num_batches * num_neurons`` size arrays so indices into both need to have ``batchOffset`` adding to them)

I have also updated the feature test which _nearly_ covered this to actually do so!

Fixes #564

### DIFF
--- a/src/genn/genn/code_generator/backendSIMT.cc
+++ b/src/genn/genn/code_generator/backendSIMT.cc
@@ -368,11 +368,14 @@ void BackendSIMT::genNeuronPrevSpikeTimeUpdateKernel(CodeStream &os, const Subst
                     os << "if(" << popSubs["id"] << " < group->spkCnt[" << ((batchSize == 1) ? "0" : "batch") << "])";
                     {
                         CodeStream::Scope b(os);
-                        os << "group->prevST[group->spk[";
-                        if(batchSize > 1) {
-                            os << "batchOffset + ";
+                        os << "group->prevST[";
+                        if (batchSize == 1) {
+                            os << "group->spk[" << popSubs["id"] << "]";
                         }
-                        os << popSubs["id"] << "]] = " << popSubs["t"] << " - DT;" << std::endl;
+                        else {
+                            os << "batchOffset + group->spk[batchOffset + " << popSubs["id"] << "]";
+                        }
+                        os << "] = " << popSubs["t"] << " - DT;" << std::endl;
                     }
                 }
                 if(ng.getArchetype().isPrevSpikeEventTimeRequired()) {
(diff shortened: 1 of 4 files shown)


tnowotny (approved):
Yes, that looks like I would have expected. Mysterious that testing for batchsize > 1 is so elusive in the definition of the update group but indeed not much harm done to include numNeurons always.
Also Fabian has confirmed his code now works as he expected.

PR #567, opened 2023-02-20
Fuse PSM rule change

neworderofjamie:
When fusing postsynaptic models, the decision on which can be fused was previously very conservative - any with state variables or extra global parameters were immediately excluded. This makes simulating large models with e.g. alpha synapses pretty much impossible so here I've relaxed the rules to match those used for fusing pre and postsynaptic updates from weight update models:

- Postsynaptic models cannot be merged if they have any extra global parameters that are referenced in their code strings (bit silly but it's consistent with the weight update models)
- Models with variables can be merged as long as they are initialized to a constant and two postsynaptic models will be merged together if these constant values are the same.

I _think_ this makes sense :smile:


tnowotny (approved):
Difficult business this but you have convinced me!

PR #576, opened 2023-03-14
Expose num

### LINKED ISSUE #553: $(num_pre), $(num_post), $(num_batches) etc should be exposed everywhere
This would be helpful to e.g. avoid having to add parameters to neuron models and trivial to implement

neworderofjamie:
Implementing EventProp, I got annoyed enough to implement #553. As well as improving usability, with large models where multiple synapse/neuron groups are actually getting merged together, this will improve performance as otherwise the "special function" unit on GPUs can get saturated by float->int conversions

Fixes #553


tnowotny (approved):
Also looks good - not too many surprises here either.

PR #578, opened 2023-03-24
Fix PyGeNN shared neuron

neworderofjamie:
Clearly #539 left a whole bunch of corner cases unhandled involving custom updates actually operating directly (rather than reducing into) ``SHARED_NEURON`` variables (i.e. duplicated across batches but shared across neurons) which I encountered when trying to use these in mlGeNN:

1. Added ``CustomUpdate::isPerNeuron`` - this is set if any custom update variables or variable reference targets _aren't_ ``SHARED_NEURON``
2. If Custom updates _are_ per-neuron then they can only read from ``SHARED_NEURON`` variables
3. Added SIMT and single-threaded CPU code generation for custom updates which _aren't_ per-neuron.
4. PyGeNN now makes correctly-sized views for accessing ``SHARED_NEURON`` variables

And, as ever, I extended some feature tests to cover this behaviour as well as adding a couple of unit tests to check the  ``CustomUpdate::isPerNeuron`` logic. On a similarly typical vein, there are also a couple of unrelated small fixes:

1. ``SynapseGroup.matrix_type`` can now be set from enumeration as well as by string (useful for mlGeNN connectivity type checking)
2. You can now use ``SynapseGroup.get_var_values`` on ``KERNEL`` connectivity (useful when attempting to learn convolutions)

### DIFF
--- a/include/genn/genn/customUpdate.h
+++ b/include/genn/genn/customUpdate.h
@@ -149,12 +149,12 @@ protected:
     template<typename V>
     void checkVarReferenceBatching(const std::vector<V>& varRefs, unsigned int batchSize)
     {
-        // If target of any variable references is duplicated, custom update should be batched
+        // If target of any variable references is not shared across batches, custom update should be batched
         if(batchSize > 1) {
             m_Batched = std::any_of(varRefs.cbegin(), varRefs.cend(),
                                     [](const V& v) 
                                     {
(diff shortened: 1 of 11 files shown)


tnowotny (approved):
What you say makes sense & I trust you that the code does as you say - also as usual I am afraid.

PR #581, opened 2023-04-25
Ability to unload PyGeNN models

neworderofjamie:
PyGeNN ``GeNNModel`` objects wrap both a ``ModelSpec`` holding the model description and a ``SharedLibraryModel`` for loading the dynamic library created from the generated code and executing it. This PR lets you 'unload' the ``SharedLibraryModel`` component, freeing memory but allowing it to be easily re-loaded if required. This entails:
- In PyGeNN, resetting all views to ``None`` to prevent freed memory being accessed
- Adding a ``close`` method to ``SharedLibraryModel`` which sets all pointers to variables etc to ``nullptr``, empties the cache used to speed up access to functions and unloads the shared library.

### DIFF
--- a/pygenn/genn_groups.py
+++ b/pygenn/genn_groups.py
@@ -321,6 +321,25 @@ class Group(object):
         for var_name, var_data in iteritems(var_dict):
             self._load_egp(var_data.extra_global_params, var_name)
 
+    def _unload_vars(self, var_dict=None):
+        # If no variable dictionary is specified, use standard one
+        if var_dict is None:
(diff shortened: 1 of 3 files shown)


tnowotny (approved):
Ahem, sorry for the delay. This looks reasonable though I do not have the overview of whether everything is cleaned up that needs cleaning up. What would be a good test for this?

neworderofjamie:
Will defer adding new tests until GeNN 5 when I reimplement feature tests in Python - raised #587 to remind me

PR #583, opened 2023-05-01
Extra Global Parameter references

neworderofjamie:
My per-timestep EventProp loss implementation required extra global parameters to be shared between a neuron group and a custom update which, as we have talked for a while, is not an uncommon requirement. However, all my attempts to design some sort of extra global parameter sharing system ended up looking very messy so, instead, here I added the ability to reference extra global parameters in custom updates. This is:
- Simple - the existing system for updating pointers in merged groups can handle it without modification so it's totally transparent to the lower levels of GeNN
- Familiar - the syntax is almost identical to variable references so 
- Safer - by forcing custom updates to be involved, it discourages users from trying to use these to connect neuron groups together in weird ways that will probably fail silently and not-deterministically

A model with EGP refs might look like:
```python

custom_update_model = genn_model.create_custom_custom_update_class(
    "custom_update",
    var_refs=[("x", "scalar")],
    extra_global_param_regs=[("e", "scalar*")],
    update_code="""
    if($(id) == (int)round(fmod($(t), 10.0))) {
        $(e)[$(id)] = 1.0;
    }
    else {
        $(e)[$(id)] = 0.0;
    }
    """)
```
Then you might add it to a model and configure the EGP reference like:
```python
model.add_custom_update("CustomUpdate1", "Group", custom_update_model,
                        {}, {}, {"x": genn_model.create_var_ref(pop, "V"}, 
                        {"e": genn_model.create_egp_ref(pop, "E")})
```
where ``pop`` is some sort of population with a ``V`` state variable and an EGP called ``E``. Note that, in this useless example, the ``x`` var reference is totally unused - it's just there to define the 'size' of the custom update.


tnowotny (approved):
Seems a straightforward addition

PR #592, opened 2023-07-05
Create Docker wheel builder

bdevans:
This adds `Dockerfile.builder` and a Makefile target (`make wheel`) to cleanly build python wheels within a Docker container. The versions of CUDA and Python can be specified by passing build arguments.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,8 +17,8 @@ ENV CUDA_PATH=/usr/local/cuda \
     GENN_PATH=/opt/genn
 
 # Upgrade pip itself and install numpy and jupyter
-RUN pip install --upgrade pip && \
-    pip install numpy jupyter matplotlib
+RUN python -m pip install --upgrade pip && \
+    python -m pip install numpy jupyter matplotlib
 
 # Copy GeNN into /opt
 COPY  . ${GENN_PATH}
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,29 +1,24 @@
 ARG BASE=11.5.0-devel-ubuntu20.04
 FROM nvidia/cuda:${BASE}
 
-LABEL maintainer="J.C.Knight@sussex.ac.uk"
-LABEL version="4.8.0"
-LABEL org.opencontainers.image.documentation="https://genn-team.github.io/"
-LABEL org.opencontainers.image.source="https://github.com/genn-team/genn"
-LABEL org.opencontainers.image.title="GeNN Docker image"
+LABEL maintainer="J.C.Knight@sussex.ac.uk" \
+    version="4.8.0" \
+    org.opencontainers.image.documentation="https://genn-team.github.io/" \
+    org.opencontainers.image.source="https://github.com/genn-team/genn" \
+    org.opencontainers.image.title="GeNN Docker image"
 
-# Update APT database and upgrade any outdated packages
+# Update APT database and upgrade any outdated packages and install Python, pip and swig
 RUN apt-get update && \
-    apt-get upgrade -y
+    apt-get upgrade -y &&\
+    apt-get install -yq --no-install-recommends python3-dev python3-pip swig gosu nano
 
-# Install Python, pip and swig
-RUN apt-get install -yq --no-install-recommends python3-dev python3-pip swig gosu nano
+# Set environment variables
+ENV CUDA_PATH=/usr/local/cuda \
+    GENN_PATH=/opt/genn
 
-# Set CUDA environment variable
-ENV CUDA_PATH=/usr/local/cuda
-
-ENV GENN_PATH=/opt/genn
-
-# Upgrade pip itself
-RUN pip install --upgrade pip
-
-# Install numpy and jupyter
-RUN pip install numpy jupyter matplotlib
+# Upgrade pip itself and install numpy and jupyter
+RUN pip install --upgrade pip && \
+    pip install numpy jupyter matplotlib
 
 # Copy GeNN into /opt
 COPY  . ${GENN_PATH}
@@ -36,9 +31,9 @@ RUN make install -j `lscpu -p | egrep -v '^#' | sort -u -t, -k 2,4 | wc -l`
 RUN make DYNAMIC=1 LIBRARY_DIRECTORY=${GENN_PATH}/pygenn/genn_wrapper/ -j `lscpu -p | egrep -v '^#' | sort -u -t, -k 2,4 | wc -l`
 RUN python3 setup.py develop
 
-# Default command will be to launch bash
-CMD ["/bin/bash"]
-
 # Start entrypoint
 # **NOTE** in 'exec' mode shell arguments aren't expanded so can't use environment variables
 ENTRYPOINT ["/opt/genn/bin/genn-docker-entrypoint.sh"]
+
+# Default command will be to launch bash
+CMD ["/bin/bash"]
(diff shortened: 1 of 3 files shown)


bdevans:
You may also want to consolidate `GENN_DIR` and `GENN_PATH` into one argument throughout the project, as they appear to refer to the same path but have different names in the `Dockerfile` and `Makefile`.

neworderofjamie:
This is very cool - would it be possible to build them on manylinux2014 (https://github.com/pypa/manylinux) for compatibility across linuxes? https://github.com/ameli/manylinux-cuda has some docker files for installing CUDA on top of many linux

bdevans:
I can have a go at basing it on `manylinux2014`. Those docker images don't have the `cuDNN` libraries though - does the wheel need them?

neworderofjamie:
That'd be awesome -  GeNN doesn't need cuDNN at all, just cuRAND

bdevans:
Here you go! I've renamed the previous target (use it with: `make ubuntu_wheel`) and added a new one `make wheels` which will build wheel for each of the installed python versions for a given CUDA version. 

Note that building against python 3.12 currently fails (probably to do with it still being in beta) so I included a clause to skip it which will eventually need to be removed. Also the build process still uses the double `python setup.py bdist_wheel` craziness, so the second one will need to be removed when a new release is made with the fix.

Dockerfile
```diff
@@ -1,44 +1,42 @@
 ARG BASE=11.5.0-devel-ubuntu20.04
 FROM nvidia/cuda:${BASE}
 
-LABEL maintainer="J.C.Knight@sussex.ac.uk"
-LABEL version="4.8.0"
-LABEL org.opencontainers.image.documentation="https://genn-team.github.io/"
-LABEL org.opencontainers.image.source="https://github.com/genn-team/genn"
-LABEL org.opencontainers.image.title="GeNN Docker image"
+LABEL maintainer="J.C.Knight@sussex.ac.uk" \
```
neworderofjamie, line 4: thank you for the tidying here - I didn't realise just how directly dockerfile lines mapped to layers and thus massive disk wastage when I wrote this

Dockerfile.ubuntu_builder
```diff
@@ -0,0 +1,46 @@
+# syntax=docker/dockerfile:1
```
neworderofjamie, line 1: What's the point of keeping (and maintaining) this one? if you have manylinux wheels you can use them on all the ubunti

Dockerfile.ubuntu_builder
```diff
@@ -0,0 +1,46 @@
...
+    python -m pip install numpy swig
+
+# Copy GeNN into /opt
+COPY . ${GENN_PATH}
+
+# Use this as working directory
+WORKDIR ${GENN_PATH}
+
+# Install GeNN and PyGeNN
+RUN make install -j `lscpu -p | egrep -v '^#' | sort -u -t, -k 2,4 | wc -l`
```
neworderofjamie, line 35: there is no need to ``make install`` when all you're doing is build PyGeNN - that's only if you want a system-wide installation to use GeNN from C++

build-wheels.sh
```diff
@@ -0,0 +1,45 @@
...
+# Compile wheels
+for PYBIN in /opt/python/*/bin; do
+    # Exclude Python 3.12
+    if [[ "$PYBIN" == *"cp312"* ]]; then
+        continue
+    fi
+    # "${PYBIN}/pip" install -r /io/dev-requirements.txt
+    "${PYBIN}/pip" install numpy swig
+    # "${PYBIN}/pip" wheel /opt/genn/ --no-deps -w dist/
+    make install -j `lscpu -p | egrep -v '^#' | sort -u -t, -k 2,4 | wc -l`
```
neworderofjamie, line 30: Again, no need to make install

build-wheels.sh
```diff
@@ -0,0 +1,45 @@
...
+    make DYNAMIC=1 LIBRARY_DIRECTORY=${GENN_PATH}/pygenn/genn_wrapper/ -j `lscpu -p | egrep -v '^#' | sort -u -t, -k 2,4 | wc -l`
```
neworderofjamie, line 31: this doesn't link against python so could be done once outside of the loop

Makefile
```diff
@@ -73,3 +73,24 @@ clean:
...
+		--target=output --output type=local,dest=dist/ .
+
+CUDA := 12.2
+.PHONY wheels:
+wheels:
+	@docker build -f Dockerfile.builder \
+		--build-arg CUDA=$(CUDA) \
+		--target=output --output type=local,dest=dist/ .
+
+# TODO: Consider build with docker run instead of docker build
```
neworderofjamie, line 93: What would the advantages of that be?

Dockerfile
```diff
@@ -1,44 +1,42 @@
 ARG BASE=11.5.0-devel-ubuntu20.04
 FROM nvidia/cuda:${BASE}
 
-LABEL maintainer="J.C.Knight@sussex.ac.uk"
-LABEL version="4.8.0"
```
neworderofjamie, line 5: Do you know if there's any way to read this from the version.txt? I've managed to convince setuptools and doxygen to do this but couldn't figure it out for docker

Dockerfile.builder
```diff
@@ -0,0 +1,28 @@
+# syntax=docker/dockerfile:1
+
+ARG CUDA=12.2
+FROM sameli/manylinux2014_x86_64_cuda_${CUDA} AS build
```
neworderofjamie, line 4: Do we want to rely on someone else's dockerfile here or am I being unduly paranoid? All they do is install CUDA from rpm files as I remember, we could just copy-paste

neworderofjamie (commented):
Very cool! manylinux wheels will be an excellent feature for GeNN 5.0.0

Dockerfile.ubuntu_builder
bdevans: Well you can ignore it for now but it might be a more robust alternative for Ubuntu users and it's a little simpler so might also be a useful reference for when the build system changes.

Dockerfile.ubuntu_builder
bdevans: Ok I'll remove it.

build-wheels.sh
bdevans: Ditto.

build-wheels.sh
bdevans: I must have misunderstood something then. The packages need to be installed within the loop still though don't they? Since GeNN uses them for bindings, doesn't that also mean it should be built within the loop?

Makefile
bdevans: I'm not sure yet, beyond saving a bit of time in image building. It was the method used in the example but I'd need to investigate a bit more.

Dockerfile
bdevans: Well passing it as a `build-arg` is simple and explicit but I expect you could also do it with the right `cat` / `sed` incantation.

Dockerfile.builder
bdevans: It seems to be regularly updated at the moment so no harm in using it and it means avoiding having to deal with horrible version strings in the CUDA filenames.

build-wheels.sh
neworderofjamie: the stuff you're installing with pip definitely need to be done within the loop but ``make DYNAMICS=1 etc etc`` isn't Python-specific, it just builds GeNN as shared libraries and sticks them where the python build system expects them to be so can move it outside the loop (it probably won't do anything after the first iteration anyway cos nothing will have changed)

PR #597, opened 2023-09-05
Casting the incoming list-like "post_indices" to numpy array

tnowotny:
Previously, there was an error if "post_indices" was a list because self.synapse_order is a numpy array and therefore cannot be used to index into this list.


neworderofjamie:
Nice! Personally I would do:
```python
pre_indices = np.asarray(pre_indices)
post_indices = np.asarray(post_indices)
```
right at the top of the ``is_ragged`` block as ``np.asarray()`` does nothing if the input's already an array and I suspect it'll be happening repeatedly inside ``np.lexsort`` and ``np.bincount` otherwise.

neworderofjamie (approved):

PR #598, opened 2023-09-26
Step 2 - Variable dimensions

neworderofjamie:
This is a bit of a boring internal one that was originally going to wider reaching but got de-scoped. Basically the ``VarAccessDuplication`` enumeration evolved a bit organically with GeNN 4 as I started adding reductions and custom updates and, in its previous form, didn't really make sense for custom update variables as the meaning of the various dimension flags can get ignored depending on what it's attached to. 

This PR replaces ``VarAccessDuplication`` with the ``VarAccessDim`` bitfield enumeration specifying what dimensions variables have so e.g. 'normal' neuron or synapse variable now have ``VarAccessDim::ELEMENT| VarAccessDim::BATCH``. However, specifying these at this level is really annoying and lots of permutations aren't valid in various locations. This PR redefines ``VarAccess`` in terms of valid combinations of ``VarAccessMode`` and ``VarAccessDim`` and adds ``CustomUpdateVarAccess`` which defines dimensions  _subtractively_ , starting with ORing together the dimensions of all the variables your custom update is referencing and then determine what shape each variable should be e.g. the standard custom update variable access is ``CustomUpdateVarAccess::READ_WRITE``  which is defined as ``VarAccessMode::READ_WRITE`` meaning it inherits the OR'd together dimensionality. However, ``CustomUpdateVarAccess::READ_ONLY_SHARED_ELEMENT`` is defined as ``VarAccessMode::READ_ONLY | VarAccessDim::ELEMENT`` meaning that it's dimensionality is the result of clearing the ``ELEMENT`` bit from the OR'd together dimensionality.

### DIFF
--- a/include/genn/backends/cuda/backend.h
+++ b/include/genn/backends/cuda/backend.h
@@ -383,7 +383,7 @@ private:
             const auto *cm = cg.getArchetype().getCustomUpdateModel();
             for(const auto &v : cm->getVars()) {
                 // If variable is reduction target
-                if(v.access & VarAccessModeAttribute::REDUCE) {
+                if(v.getAccessMode() & VarAccessModeAttribute::REDUCE) {
                     // Add pointer field
                     const auto resolvedType = v.type.resolve(cg.getTypeContext());
                     groupEnv.addField(resolvedType.createPointer(), "_" + v.name, v.name,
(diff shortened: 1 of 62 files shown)


tnowotny (approved):

PR #602, opened 2023-10-31
Step 3 - Dynamic loader

### LINKED ISSUE #408: Linker-imposed model complexity limit on Windows
#407 allows large models to be _compiled_ on Windows if you have sufficient RAM but, trying to _link_ the multi-area model code now fails with "LINK : fatal error LNK1189: library limit of 65535 objects exceeded". As the multi-area model has about 70000 synapse populations and, even with all the runner-size-reduction optimizations in place, there is still more than one symbol exported for each one so this is hardly surprising. Seemingly this isn't even a Visual Studio limitation but rather a limitation of the Windows PE executable format which, somewhere, uses 16-bit indices for symbols (https://developercommunity.visualstudio.com/t/fix-msvc-65535-symbol-limit-in-lib-files-lnk1189/220174). 

Obviously lol Windows but, while #286 has meant that while the size of _kernels_ generated for complex models are kept manageable as long as the model's _relatively_ homogeneous, the runner can still get really big. This means that, even on platforms where this particular limit doesn't apply, the compilation time and memory requirements for huge runners currently limits the maximum model size (in terms of number of populations). Generating runner code that looks more like the following would be one option but quite how it would handle heterogeneous populations is another question:
```c++
enum NeuronPopulations
{
    NeuronPopulationPre,
    NeuronPopulationPost,
    NeuronPopulationMax,
}

scalar *V[NeuronPopulationMax];
scalar *d_V[NeuronPopulationMax];
const size_t popSizes[NeuronPopulationMax] = {100, 100};

void pushVToDevice(NeuronPopulations pop) {
    CHECK_CUDA_ERRORS(cudaMemcpy(d_V[pop], V[pop], popSizes[pop] * sizeof(scalar), cudaMemcpyHostToDevice));
}
```

neworderofjamie:
Since the code generator started merging generated code together, what has stood in the way of simulating complex models is the size of the generated runner.cc which quickly grows to the point where it takes a long time and a lot of memory to compile. Changes like #554 have helped but don't really solve the fundamental issue and my various past attempts to generate less stupid runner code have all proved painful and complex. 

I thought about switching to [NVRTC](https://docs.nvidia.com/cuda/nvrtc/index.html) (a runtime CUDA compiler)  but ended up concluding that while NVRTC itself won't work for us, as it **only** compiles device code and we have quite a lot of generated host code as well, I realised we  _could_ still move in that direction and do almost everything the runner did in generated code at runtime quite a lot more simply and efficiently. This PR does just this within something that looks not-dis-similar to the existing ``SharedLibraryModel`` runtime which is used by PyGeNN and more complex C++ models to look up variables/functions from the shared library built from the generated code  and provides an API for calling them. Main changes are:

- Backends have much more limited functionality for _generating_ code for allocation, pushing and pulling (these are now only used in generated host code). Instead ``BackendBase`` provides a method to create an ``ArrayBase``-derived object which  encapsulates an array of bytes with a given ``VarLocation`` with all the methods to push/pull/allocate/memset this
- New ``Runtime`` object contains similar logic to old generate runner and allocated required arrays at runtime and stores them in a bunch of ``std::unordered_map`` (just like ``SharedLibraryModel`` ended up doing for caching purposes).
- Throughout the code generator, rather than providing a function which returns a string pointing at the variable allocated by the generated code, GeNN now provides functions that take a reference to the runtime and extract a suitable array from it
- Array objects are wrapped with PyBind11 so are accessible from PyGeNN and linked into the existing ``Variable`` objects. These have also been extended to handle correctly re-arranging synapse variables and accessing 'current' value of delayed variables.
- ``Runtime`` provides a C++ implementation of spike recording data decoding which should be much faster and require less memory than the old numpy one.

Calling functions whose signatures are defined at runtime (like those used for building merged groups) is beyond the scope of standard C/C++ so I have had to add a dependency on [libffi](https://github.com/libffi/libffi) to handle this (https://eli.thegreenplace.net/2013/03/04/flexible-runtime-interface-to-shared-libraries-with-libffi describes the problem libffi is trying to solve quite nicely). This is super-battle tested (I believe it's used by both Python and Perl for calling C functions) and available in every package manager under the sun on Linux/Mac. I don't really want to place more hurdles in front of our poor Windows users so, as on Windows we only support the x86-64 architecture, libffi gets built as part of the normal MSBuild-based build process.

Aside from the forementioned changes to the PyGeNN syntax, this doesn't really impact Python users but it does break the old style C++ mode of operation. For small simple models you can now combine model.cc and simulator.cc like:
```c++
#include "parameters.h"

void modelDefinition(ModelSpec &model)
{
    model.setDT(1.0);
    model.setName("va_benchmark");

    //---------------------------------------------------------------------------
    // Build model
    //---------------------------------------------------------------------------
    ParamValues vDist{{"min", Parameters::resetVoltage},
                      {"max", Parameters::thresholdVoltage}};

    // LIF model parameters
    ParamValues lifParams{{"C", 1.0}, {"TauM", 20.0}, {"Vrest", -49.0}, 
                          {"Vreset", Parameters::resetVoltage},
                          {"Vthresh", Parameters::thresholdVoltage}, 
                          {"Ioffset", 0.0}, {"TauRefrac", 5.0}};

    // LIF initial conditions
    VarValues lifInit{{"V", initVar<InitVarSnippet::Uniform>(vDist)},
                      {"RefracTime", 0.0}};
    
    // Create IF_curr neuron
    auto *e = model.addNeuronPopulation<NeuronModels::LIF>("E", Parameters::numExcitatory, lifParams, lifInit);
    auto *i = model.addNeuronPopulation<NeuronModels::LIF>("I", Parameters::numInhibitory, lifParams, lifInit);

    // Enable spike recording
    e->setSpikeRecordingEnabled(true);
    i->setSpikeRecordingEnabled(true);

    auto *ee = model.addSynapsePopulation(
        "EE", SynapseMatrixType::SPARSE, NO_DELAY,
        "E", "E",
        initWeightUpdate<WeightUpdateModels::StaticPulseConstantWeight>({{"g", Parameters::excitatoryWeight}}),
        initPostsynaptic<PostsynapticModels::ExpCurr>({{"tau", 5.0}}),
        initConnectivity<InitSparseConnectivitySnippet::FixedProbabilityNoAutapse>({{"prob", Parameters::probabilityConnection}}));
    auto *ei = model.addSynapsePopulation(
        "EI", SynapseMatrixType::SPARSE, NO_DELAY,
        "E", "I",
        initWeightUpdate<WeightUpdateModels::StaticPulseConstantWeight>({{"g", Parameters::excitatoryWeight}}),
        initPostsynaptic<PostsynapticModels::ExpCurr>({{"tau", 5.0}}),
        initConnectivity<InitSparseConnectivitySnippet::FixedProbability>({{"prob", Parameters::probabilityConnection}}));
    auto *ii = model.addSynapsePopulation(
        "II", SynapseMatrixType::SPARSE, NO_DELAY,
        "I", "I",
        initWeightUpdate<WeightUpdateModels::StaticPulseConstantWeight>({{"g", Parameters::inhibitoryWeight}}),
        initPostsynaptic<PostsynapticModels::ExpCurr>({{"tau", 10.0}}),
        initConnectivity<InitSparseConnectivitySnippet::FixedProbabilityNoAutapse>({{"prob", Parameters::probabilityConnection}}));
    auto *ie = model.addSynapsePopulation(
        "IE", SynapseMatrixType::SPARSE, NO_DELAY,
        "I", "E",
        initWeightUpdate<WeightUpdateModels::StaticPulseConstantWeight>({{"g", Parameters::inhibitoryWeight}}),
        initPostsynaptic<PostsynapticModels::ExpCurr>({{"tau", 10.0}}),
        initConnectivity<InitSparseConnectivitySnippet::FixedProbability>({{"prob", Parameters::probabilityConnection}}));
}

void simulate(const ModelSpec &model, Runtime::Runtime &runtime)
{
    runtime.allocate(Parameters::numTimesteps);
    runtime.initialize();
    runtime.initializeSparse();
    
    const auto startTime = std::chrono::high_resolution_clock::now();
    while(runtime.getTimestep() < Parameters::numTimesteps) {
        runtime.stepTime();
    }
    std::chrono::duration<double> duration = std::chrono::high_resolution_clock::now() - startTime;
    std::cout << "Init time:" << runtime.getInitTime() << std::endl;
    std::cout << "Total simulation time:" << duration.count() << " seconds" << std::endl;
    std::cout << "\tNeuron update time:" << runtime.getNeuronUpdateTime() << std::endl;
    std::cout << "\tPresynaptic update time:" << runtime.getPresynapticUpdateTime() << std::endl;
    
    runtime.pullRecordingBuffersFromDevice();
    
    const auto *e = model.findNeuronGroup("E");
    const auto *i = model.findNeuronGroup("I");
    runtime.writeRecordedSpikes(*e, "spikes_e.csv");
    runtime.writeRecordedSpikes(*i, "spikes_i.csv");

}
```
and build and simulate the whole thing with ``genn-buildmodel.sh``/``genn-buildmodel.bat``. Existing more advanced C++ GeNN integrations like the Bielefeld thing or SpineML all use shared model anyway so could be updated to use runtime quite easily.

To give some idea of the performance savings, this table compares the GeNN implementation of the multi-area model (https://github.com/neworderofjamie/multi-area-model/)

|  | GeNN 4.9.0 |GeNN 5.0.0 
| ------------- |-------------:| -----:|
| Area creation [s] | 0.57 | 1.2 |
| Inter-area creation [s] |  14.2 |  34.9 |
| Compile [s] | 211 [^1]  |   19 |
| Load[s] |  166|  176 |
| Sim (100ms) [s] | 59 | 56 |

[^1]: when using #554, this drops to about 180s

Key points are:
- Model creation is a little slower probably due to scanning tokens whenever a population is created. However, as this model creates 60,000 odd populations, I think it's fine to not optimise this right now
- Compile time is an order of magnitude faster
- Loading the model dynamically is only a _tiny_ bit slower than via the runner suggesting overheads from calling thousands of virtual functions is negligible but some scope for optimisation here

Fixes #408


tnowotny (approved):

PR #604, opened 2023-11-09
Step 4 - Replace implicit neuron variable references with explicit ones

neworderofjamie:
In the previous 3 stages of the great GeNN 5 updating, the ability to reference neuron variables in postsynaptic models, weight update models and current sources _implicitly_ via ``$(X_post)`` syntax was entirely disabled but this PR adds new functionality to do this _explicitly_ via variable references. This means models are not tied together by the names of variables so e.g. you can use the standard ``ExpCond`` model even if you have the gall to call your neuron model's voltage ``u`` or (lower-case) ``v``.

However, this was the straw that broke the camel's back of the addSynapsePopulation function as it ended up with 17 arguments. This is annoying even when using Python where there's keyword arguments so, continuing the approach used to setup connectivity, I added weight update and postsynaptic model inititialisation functions:
```python
excitatory_synapse_params = {"g": EXCITATORY_WEIGHT}

excitatory_post_syn_params = {"tau": 5.0}

model.add_synapse_population("EE", "SPARSE", 0,
    excitatory_pop, excitatory_pop,
    "StaticPulseConstantWeight", excitatory_synapse_params, {}, {}, {},
    "ExpCurr", excitatory_post_syn_params, {},
    init_sparse_connectivity("FixedProbabilityNoAutapse", fixed_prob))
```
becomes something like:
```python
excitatory_weight_init = init_weight_update("StaticPulseConstantWeight", {"g": EXCITATORY_WEIGHT})
excitatory_postsynaptic_init = init_postsynaptic("ExpCurr", {"tau": 5.0})

model.add_synapse_population("EE", "SPARSE", 0,
    excitatory_pop, excitatory_pop,
    excitatory_weight_init, excitatory_postsynaptic_init,
    init_sparse_connectivity("FixedProbabilityNoAutapse", fixed_prob))
```
which I think it actually a lot nicer and lets you reuse fully-configured models much easily. To further ease the pain I have shortened and unified all of the names of the keyword arguments on the ``GeNNModel.add_XXX`` methods and given them defaults of ``{}`` so you can more complex configurations with less ``{}, {}, {}`` nonsense! Clearly this breaks backward compatibility but updating is pretty painless, to the point I was able to update all of the unit tests and feature tests with an (admittedly absolutely nightmarish) regular expression.

With that out of the way, the syntax to connect an ExpCond model back to the presynaptic neuron is simple:
```python
model.add_synapse_population("Pop1self", "SPARSE", 10,
    pop1, pop1,
    init_weight_update("StaticPulseConstantWeight", {"g": -0.2}), 
    init_postsynaptic("ExpCond", {"tau": 1.0, "E": -80.0}, var_refs={"V": create_var_ref(pop1, "V")}),
    init_sparse_connectivity(ring_model))
```
Most of the bits and pieces to implement this internally were already there but some tweaks of environment classes were required to correctly setup the lazy substitutions (e.g. references to neuron variables in per-neuron weight update model code may just be aliases to local variables in the dynamics bit but, on SIMD backends, in the spike-driven bit the indices are different).

This PR also adds the ability to change which neuron input variable a current source targets (in line with postsynaptic model). Finally, this enables a whole swathe of feature tests that rely on delays (which only get applied to neuron variables if they're referenced in synaptic code)  to be re-implemented in PyGeNN which led to a further wave of bug-fixing.


tnowotny (approved):

PR #607, opened 2023-11-24
Step 5 - Dynamic and typed parameters

neworderofjamie:
'Traditionally', in GeNN if you wanted to make parameters changable at runtime you had to change them to extra global parameters. Before GeNN 4.3.0, this used to be pretty efficient as they were all passed through to the device using kernel parameters (which are effectively implemented as a big block of constant cache). However, since adding the whole merging system way back in GeNN 4.3.0, this hasn't been possible so, for backward compatibility, they have been manually copied every timestep as there's no way to know when they change. Because they're very small and there are potentially lots of them this can be very inefficient. This functionality was totally removed in this trail of PRs but here we implement its replacement - dynamic parameters! Key things are:
- You don't need to change your models if you want to fiddle with a parameter at runtime (only caveat is that derived parameters don't get updated)
- You mark parameters as dynamic when you build a model with ``pop.set_param_dynamic("r0")``
- You update them at runtime with ``pop.set_dynamic_param_value("r0", current_r0)``

Another issue with parameters is that they didn't previously have a type. This meant in extreme cases where you have a model with lots of heterogeneous integer parameters (so the conversion couldn't be done at compile time)  e.g. a big convnet with different types of convolution implemented as Toeplitz connectivity init snippets, you could actually overload the GPU special function unit just through converting all the parameters passed as ``scalar`` to ``int``. Also, ``scalar`` is typically a 32-bit float so cannot represent the full range of a 32-bit integer. Fixing this wasn't particularly difficult, just required passing types through from the ``ParamValues`` to the parameters. However you need to be able to specify parameters in some way and, back in #595, I added ``NumericValue`` which is now exposed to python and, in this PR, used for just this. 

Fear not, there is only one step to go until we have something resembling a release candidate!


tnowotny (approved):

Issue #594, opened 2023-08-10
Presynaptic Synapse Variable undefined in Event Threshold Condition


FabianSchubert:
Situation:

- A population has **two or more** outgoing synapse groups using the same weight update model.
- The weight update model has a presynaptic variable that is used in the `event code`.

Error:
`RuntimeError: The variable <presyn. variable name> was undefined in code eventThresholdConditionCode.`
This does not happen for a single synapse population.

Minimal example:
```
from pygenn.genn_model import (GeNNModel, create_custom_neuron_class,
        create_custom_weight_update_class, create_custom_postsynaptic_class)

neuron_model = create_custom_neuron_class(
        "test_neuron",
        param_names=[],
        var_name_types=[("r", "scalar")],
        sim_code=""
)

wu_model = create_custom_weight_update_class(
    "test_wu_model",
    param_names=["th"],
    var_name_types=[("g", "scalar")],
    pre_var_name_types=[("presyn_var", "scalar")],
    event_code="""
        $(presyn_var) = $(r_pre);
    """,
    synapse_dynamics_code="$(addToInSyn, $(g) * $(presyn_var));",
    event_threshold_condition_code="abs($(r_pre) - $(presyn_var)) >= $(th)"
)

model = GeNNModel("float", "testmodel")

p1 = model.add_neuron_population("p1", 1, neuron_model, param_space={}, var_space={"r": 1.0})
p2 = model.add_neuron_population("p2", 1, neuron_model, param_space={}, var_space={"r": 1.0})
p3 = model.add_neuron_population("p3", 1, neuron_model, param_space={}, var_space={"r": 1.0})

syn12 = model.add_synapse_population(
    "syn12", "DENSE_INDIVIDUALG", 0,
    p1, p2, wu_model, {"th": 1e-3}, {"g": 1.0},
    {"presyn_var": 0.0}, {},
    "DeltaCurr", {}, {}
)

syn13 = model.add_synapse_population(
    "syn13", "DENSE_INDIVIDUALG", 0,
    p1, p3, wu_model, {"th": 1e-3}, {"g": 1.0},
    {"presyn_var": 0.0}, {},
    "DeltaCurr", {}, {}
)

model.build()
model.load()
```

neworderofjamie:
Pretty sure this is related to the "event threshold retesting" mechanism - it is on my re-implementation hitlist for GeNN 5

PR #609, opened 2023-12-04
Step 6 - Fused event generation and postsynaptic spike-like events

neworderofjamie:
One feature that has been notably missing from GeNN is postsynaptic spike-like events. These would be very useful for implementing all kinds of event-based learning rules e.g. to add per-MBON event-based dopamine signals to a mushroom body model. However, the spike-like event system was a bit delicate (see #379 and #594) and definitely couldn't take any more complexity!

This PR adds in this functionality and lays the ground-work for enabling the "spike-splitting" optimisation described in  #484 which would hugely improve performance of convolutions. Essentially what this PR does is:

* Remove the (mostly commented out until now in GeNN 5) system for figuring out per-neuron group spike-event thresholds and the threshold re-retesting mechanism in synapse updates (this bit was always problematic)
* Change ``spk`` and ``spkCnt`` from being per-neuron group to per-fused synapse group like ``inSyn`` etc if the appropriate option is enabled (for spikes it always is), reusing the same mechanism used for #201 and #461 along with the more elegant merged group implementation from #595.

However, as ever, the devil is in the details and fusing these synapse groups is rather trickier than e.g. a postsynaptic update because, the neuron code doesn't care whether spike/spike-event thresholds are presynaptic or postsynaptic and should fuse them regardless. However this means that one synapse groups presynaptic end might be fused to another's postsynaptic end if they correspond to the same neuron group. The bugs this caused have been...fun...hence the additional logging functionality I've added to Runtime which should now tell you everything reading the source of runner.cc would have :smile:


tnowotny (commented):
Sounds like something that should be sorted but could be endlessly tricky. Are you confident you have covered all nasty confusions?

neworderofjamie:
It is definitely tricky - I've been fixing bugs in the Final PR (#613) based on new tests/STDP models. However, I think it's the right thing to do and necessary as it lays the groundwork  both for learning convolutions efficiently and more advanced event-driven three-factor learning.

tnowotny (approved):
Approved though detailed code review at this scale is beyond what I can do.

PR #612, opened 2023-12-20
Step 7 - Single PSM code string

neworderofjamie:
This is the last syntax change I want to make. Previously, postsynaptic models had two code strings "decay code" and "apply input code" (the latter of which could also be configured via "custom converter code" in C++ for backward compatibility). This separation doesn't really make any sense and it's not obvious (if you'd asked me before I made this change, I couldn't tell you) what order they're run in. With this change:

```python
exp_curr_model = create_postsynaptic_model(
        "exp_curr",
        decay_code=
        """
        inSyn *= expDecay;
        """,
        apply_input_code=
        """
        Isyn += init * inSyn;
        """,
        params=["tau"],
        derived_params=[....])
```

becomes:

```python
exp_curr_model = create_postsynaptic_model(
        "exp_curr",
        sim_code=
        """
        injectCurrent(init * inSyn);
        inSyn *= expDecay;
        """,
        params=["tau"],
        derived_params=[....])
```
The user is no longer expected to know that ``Isyn`` is a magical thing you can only meaningfully add to (and can be renamed by setting post_target_var on the synapse group) and the injection of current uses the same syntax as current sources.

### DIFF
--- a/include/genn/genn/postsynapticModels.h
+++ b/include/genn/genn/postsynapticModels.h
@@ -9,9 +9,7 @@
 //----------------------------------------------------------------------------
 // Macros
 //----------------------------------------------------------------------------
-#define SET_DECAY_CODE(DECAY_CODE) virtual std::string getDecayCode() const override{ return DECAY_CODE; }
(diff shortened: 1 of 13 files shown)


tnowotny (approved):
Yes, I think this is worth doing and makes things simpler and more sensible.

PR #615, opened 2024-01-26
custom_updates in genn_model.end not defined -> self.custom_updates

FabianSchubert:
### Issue:
`genn_model.end()` throws error because `custom_updates` is not defined.
### Fix:
I assume this should reference the instance variable `self.custom_updates`.

### DIFF
--- a/pygenn/genn_model.py
+++ b/pygenn/genn_model.py
@@ -893,7 +893,7 @@ class GeNNModel(object):
     def end(self):
         """Free memory"""
         for group in [self.neuron_populations, self.synapse_populations,
-                      self.current_sources, custom_updates]:
+                      self.current_sources, self.custom_updates]:
             for g_name, g_dat in iteritems(group):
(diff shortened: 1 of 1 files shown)


neworderofjamie:
Good spot - I don't think I've ever actually called that method! I also think this may be the function @tnowotny was looking for..

neworderofjamie:
Once the checks pass, I will merge this **but** it's lifetime may be short as I'm not planning on making any more GeNN 4.X releases and GeNN 5 automatically frees extra-global parameters when the model is destroyed

FabianSchubert:
`egp_dat.needsAllocation` is also not defined - could this be replaced by `not(egp_dat.is_scalar)` because only pointer type EGPs require allocation/freeing of memory?

FabianSchubert:
I did test it on a Genn model that I set up, and it worked, but I missed to actually include some egps to test the other part, because I assumed this was just a small typo - sorry for that, I'll look into it.

Issue #437, opened 2021-07-15
Improve spike delivery for sparse connections using shared memory and shared atomics


tnowotny:
As detailed in https://arxiv.org/pdf/2107.04092.pdf one can slice the rows of the connectivity matrix into vertical slices where each slice addresses a defined slice of post-synaptic neurons. Pre-loading these neurons' relevant variables into shared memory and using shared atomics for inSyn updates reduces global memory traffic leading to measurable performance gains.

neworderofjamie:
Square blocks also lets you improve the performance of spike processing in all postsynaptically-parallelised strategies by using threads in y to process multiple spikes in parallel. For sparse matrices these can atomically update global/shared memory and for dense matrices they can update registers and use warp shuffles to sum.

PR #618, opened 2024-04-08
Use warp shuffles for addToPre when using postsynaptic parallelism

neworderofjamie:
Profiling confirmed my suspicion that using global memory atomic adds here was very innefficient and becomes more so with large hidden layers. Copy-pasted my trusty CUDA playground (https://github.com/neworderofjamie/dense_out_pre_kernels) I tried a few things but:
- The CUDA compiler is smart enough to 'lift' repeated loads of postsynaptic lambda variables out of the loop/they just sit in L1 cache making them fast to access i.e. https://github.com/neworderofjamie/dense_out_pre_kernels/blob/master/kernel.cu#L213-L240 performs the same as https://github.com/neworderofjamie/dense_out_pre_kernels/blob/master/kernel.cu#L260-L292
- Switching to _shared memory_ atomics, the stalling the profiler shows just moves from global memory "long scoreboard" to shared memory "short scoreboard" and performance actually gets worse (I suspect due to the extra synchronisation required)
- Warp shuffles can massively improve the performance

Figure_2

On _real_ models, the performance gains are sadly somewhat less because, in single hidden layer models, input->hidden (where no gradient transport is required) tends to have the largest number of connections and the cost of making the weight updates/propagating spikes is also significant. Nonetheless, the new kernel is never slower than the old one and with larger batch sizes (where the memory subsystem is being thrashed harder) and larger hidden layers (where more postsynaptic neurons are issuing simultaneous colliding atomic adds) the savings can still be significant (these times are per-epoch of SHD training): 
Figure_1

Note, another factor behind the difference in speedup might well be that the low-level benchmarking was done on my laptop (A500 GPU with 96 GB/s memory bandwidth) and the SHD tests were done on Cortex (3080 GPU with 760 GB/s)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,12 +1,9 @@
 .git
 build
 dist
-documentation
-doxygen
+docs
 lib
 obj*
 tests
-userproject
-!userproject/include
 **/*.so
 **/*.lib
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ RUN apt-get update && \
     apt-get upgrade -y
 
 # Install Python, pip and swig
-RUN apt-get install -yq --no-install-recommends python3-dev python3-pip gosu nano
+RUN apt-get install -yq --no-install-recommends python3-dev python3-pip libffi-dev pkgconf gosu nano
 
 # Set CUDA environment variable
 ENV CUDA_PATH=/usr/local/cuda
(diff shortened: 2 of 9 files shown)


tnowotny (approved):
Nice work on this one! This is the kind of stuff where GeNN keeps the edge if we can make it work well.
Haven't checked the gory details as usual but trust you've seen it working correctly.

neworderofjamie:
Thanks! Totally agree, was planning on focussing a little on optimisations in the near future (e.g. #437  would hopefully give a decent speedup for sparse models)

PR #621, opened 2024-04-19
Rebuild ``remap`` and ``colLength`` after custom connectivity updates

neworderofjamie:
When combining STDP with a postsynaptic spike-driven and custom connectivity, there was previously a problem - the transpose ragged matrix data structure (``remap`` and ``colLength``) wasn't being updated when connectivity changes were made.

Sadly the transpose matrix can't be efficiently updated at the same time as the forward matrix so we simply re-generate it after the custom connectivity update is complete using the same code that's already used to build the transpose matrices at initialisation time. Figuring out exactly which synapse groups transpose matrices need updating is a little fiddly but I have unit tested this reasonably well and added a feature test for the whole system. Before re-generating the transpose matrix, ``colLength`` needs to be zeroed - the most _efficient_ approach would be to add yet another kernel which does all of these in parallel but that's a lot of effort so instead I just implemented this in the runtime.

### DIFF
--- a/include/genn/backends/cuda/backend.h
+++ b/include/genn/backends/cuda/backend.h
@@ -148,7 +148,6 @@ private:
     //------------------------------------------------------------------------
     // Members
     //------------------------------------------------------------------------
-    bool m_EnableNCCLReductions;
     VoidFunction m_NCCLGenerateUniqueID;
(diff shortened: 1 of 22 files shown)


tnowotny (approved):
yeah, I think it makes sense and looks good.

PR #623, opened 2024-05-02
Better fix for clang errors and reduced warnings

neworderofjamie:
The better fix for the actual compiler error was to add trailing return types ``-> uint64_t`` to the lambda functions.

However, the majority of the changes are to fix warnings, mostly:
- Unused captured lambda parameters
- Missing virtual destructors in a few base classes
- A few unused variables of other sorts and a whole unused method


tnowotny (approved):
Looks good and fix of the actual error is general (not Mac-specific) - nice.

neworderofjamie:
I would suspect it will fix compilation with clang on other platforms as well (and reduce the warnings)

PR #629, opened 2024-06-29
Access to postsynaptic variables with heterogeneous delay

neworderofjamie:
This PR enables you to access postsynaptic neuron variable references and postsynaptic weight update model variables in synapse code (pre, post spike/spike-like event triggered or synapse dynamics) using an array subscript expression to specify a potentially heterogeneous delay slot e.g. ``x_post[d]``. This interacts with the existing delay system as follows:

* If any postsynaptic _neuron_ variable references are made with new ``[]`` syntax
    - Number of postsynaptic neuron group delay slots needs to be made at least as large as max dendritic delay
    - Target neuron variables need to be marked as queued using the existing mechanism so they are allocated + initialised with delay slots
* If any postsynaptic _weight update model_ variables are accessed with new ``[]`` syntax
    - Again, the number of postsynaptic neuron group delay slots needs to be made at least as large as max dendritic delay 
    - With the old homogeneous delays, if synapse group was delayed, all weight update model pre and postsynaptic variables were delayed. However, now, some postsynaptic weight update model variables might be delayed and some not. To address this I've added a very similar mechanism to the neuron variable queuing to synapse group which has repercussions for fusing etc etc

One side effect of this change is that neuron groups might now have some variables which require queueing but no axonal/backpropagation delays applied to their spikes. This resulted in unnecessarily complex code being generated for spike handling so I have extended the existing tracking of which variables are queued to also track whether spikes or spike events are queued.

The handling of the actual ``[d]`` syntax is implemented by adding ``x_post`` to the type system as a special sort of function (with signature ``scalar x_post(int index)`` (assuming the type of ``x_post`` is ``scalar``). These functions are tagged with the ``Type::FunctionFlags::ARRAY_SUBSCRIPT_OVERRIDE`` flag. The type checker and pretty printer then treat, ``[]`` as a special sort of function call which enables all the existing substitution magic used to stick code behind functions to be used.

In the process of making this change, I tidied up a few things:

- Added an error if ``addToPostDelay`` is called or the new ``[]`` syntax is used in synapse code but no maximum number of dendritic delay steps has been specified.
- Types in the type system don't really need an enum of qualifiers - the only qualifier that's ever going to be supported is ``const`` so I have simplified a bunch of code by replacing this with a simple bool. Conversely, function types *do* need some more flags so replaced ``variadic`` boolean with some flags (now including ``Type::FunctionFlags::ARRAY_SUBSCRIPT_OVERRIDE`` )
- Because of type promotion, it doesn't _really_ matter but the types of ``addToPre`` functions etc actually vary depending on what type ``scalar`` is so constants like ``Type::AddToPre`` have been replaced by functions like ``Type::getAddToPre(Type::Float)`` which generate the right function type.
- The syntax for ``EnvironmentLocalVarCache`` was unnecessarily complex - the adaptor type being passed in already provides the required information about delay an 99% of the time this is what determines whether variables should be copied between delay slots every timestep.
- Finally got annoyed enough to update some warnings about deprecated kwargs in the tests

### DIFF
--- a/include/genn/genn/code_generator/environment.h
+++ b/include/genn/genn/code_generator/environment.h
@@ -694,6 +694,38 @@ public:
     }
 
     template<typename A>
+    void addVarsHet(GetVarIndexFn<A> getIndexFn, GetVarIndexFn<A> getDelayedIndexFn, 
+                    const std::string &fieldSuffix = "", bool readOnly = false, bool hidden = false)
+    {
+        // Loop through variable references
+        const A archetypeAdaptor(this->getGroup().getArchetype());
+        for(const auto &v : archetypeAdaptor.getDefs()) {
+            // If variable access is read-only, qualify type with const
+            const auto resolvedType = v.type.resolve(this->getGroup().getTypeContext());
+            const auto qualifiedType = (readOnly || (getVarAccessMode(v.access) & VarAccessModeAttribute::READ_ONLY)) ? resolvedType.addConst() : resolvedType;
+            const auto name = hidden ? ("_" + v.name) : v.name;
+            if(archetypeAdaptor.isVarDelayedInSynCode(v.name)) {
+                addField(Type::getArraySubscript(resolvedType), name,
+                         resolvedType.createPointer(), v.name + fieldSuffix,
+                         [v](auto &runtime, const auto &g, size_t) 
+                         { 
+                             return runtime.getArray(A(g).getTarget(), v.name);
+                         },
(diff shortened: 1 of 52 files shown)


tnowotny (approved):
I think it all makes sense. For a moment I thought the test for the continuous reverse delay synapse was a bit weak in that it did not test that reverse add could happen at different times and not test from which synapse/post neuron the reverse input was coming, but on reflection, it seems unlikely one can pass the current test if there was a bug.

PR #630, opened 2024-07-03
Den delay batch fix

neworderofjamie:
Clearly the combination of heterogeneous dendritic delay and batching had never been tested and the indexing code was completely broken. I further parameterised the existing test for dendritic delay to cover this.


tnowotny (approved):
I have to largely trust you on this indexing nightmare.

PR #631, opened 2024-07-15
Ability to add to kernel weights

neworderofjamie:
This PR adds:

1. A backend-agnostic way of generating simple (i.e. not returning previous value or selecting memory spaces) atomic functions (on SIMT backends, this just wraps subset of existing atomic-function generation functionality and, on single-threaded CPU backend, it inserts a += style operator).
2. When making synapse substitutions, create pointer fields for kernel weights and, from these, create read-only variable accessors AND atomic functions to allow modification of READ-WRITE variables.

I am not going to document this as it's more of a quick fix than how I want to implement this in the longer term.


tnowotny (approved):
Seems like a sensible solution for the moment.

PR #634, opened 2024-08-20
Special variable references for zeroing internals

neworderofjamie:
Having to zero things like ``outPost`` (``inSyn`` in GeNN 4 parlance) on the host between trials slows down simulations with extra ``cudaMemcpy`` and Python-side processing. This PR adds some extra functionality that lets you move more of this into custom updates and also improves the handling of delayed variables in custom updates:

* New ``VarAccessMode::BROADCAST`` and ``CustomUpdateVarAccess::BROADCAST_DELAY`` to indicate variables, writes to which are 'broadcast' across all delay slots
* A ``writeOnly`` flag for value types - these can only be assigned with plain ``=`` assignement. Applied to both ``VarAccessMode::BROADCAST`` variables and also reductions to improve error handling (also, while I was in this code, tightened up some type-checking)
* Some new functions to create variable references to ``outPost`` and ``denDelay``.
* Updated MNIST MB classifier user project to use this functionality

Together, this means you can do all resetting required for delayed EventProp on device.


tnowotny (approved):
Very sensible & looks good.

PR #636, opened 2024-08-30
Fix deprecation

neworderofjamie:
Fixed a couple of minor issues:
* ``distutils`` and hence ``distutils.spawn.find_executable`` is deprecated, replace with ``shutil.which``. Only reason we didn't do this before was because GeNN 4.X tried to maintain Python 2.7 compatibility.
* ``setuptools`` 74 gets rid of the module of Visual Studio magic helpers which we use to find the compiler on Windows. For now have just restricted setuptools to < 74 (This also affects Brian https://github.com/brian-team/brian2/issues/1557)


tnowotny (approved):
Sure, one needs to do what needs to be done. Bitrot is *very* annoying.

PR #638, opened 2024-09-19
Fixed merging/fusing issue with multiple spike-like event conditions

neworderofjamie:
Initialisation doesn't care much about what happens in a merged neuron group's outgoing spike-like events but it needs to know how many different conditions there are so it can zero out the correct number of datastructures. This was missing.


tnowotny (approved):
Ah, ok. Makes sense ...

PR #642, opened 2024-11-05
GeNN 5.1.0

neworderofjamie:
Have been meaning to do this for a while as there's been quite a few bug fixes since the release.


tnowotny (approved):
Yes, looks ok. Sorry for delay.

PR #644, opened 2024-11-11
Added assert definition to standard library

neworderofjamie:
``assert`` is supported by CUDA and C++ so a no-brainer


tnowotny (approved):
No problem.

PR #646, opened 2024-12-03
Automate detection of homogeneous parameters

neworderofjamie:
One of the really awkward things with the initial implementation of merging groups was that we had to add a manual mechanism for detecting whether parameters (and derived parameters) were heterogeneous across the merged group or not and handle them separately. With the improved type information present in GeNN 5, it was always on my todo list to simplify and automate it so it could be applied to 'internal' parameters like number of neurons etc.

In ``EnvironmentGroupMergedField::addField``, we now just check that the type is const, numeric and the values match (this is the key thing as, in GeNN 4, all fields basically were just strings) and if their value is the same for all groups that are being merged together, we just substitute a constant. This means for models which don't really benefit from merging i.e. ones with a small number of populations, the code will be as efficient as it was before merging was added **and** this removes a load of ugly boilerplate.

For example, after this change, generated code for a merged neuron update group with one 256 neuron group in looks like:
```c++
if(id >= 768 && id < 1024) {
     const unsigned int lid = id - 768;
     struct MergedNeuronUpdateGroup2 *group = &d_mergedNeuronUpdateGroup2[0]; 
     {
        const unsigned int batchOffset = 256u * batch;
        if(lid < 256u) {
             // **DO STUFF**
        }
    }
}
```
rather than:
```c++
if(id >= 768 && id < 1024) {
    const unsigned int lid = id - 768;
    struct MergedNeuronUpdateGroup1 *group = &d_mergedNeuronUpdateGroup2[0]; 
    {
        const unsigned int batchOffset = group->numNeurons * batch;
        if(lid < group->numNeurons) {
            // **DO STUFF**
        }
    }
}
```

### DIFF
--- a/include/genn/backends/cuda/backend.h
+++ b/include/genn/backends/cuda/backend.h
@@ -345,25 +345,26 @@ private:
 
             // Implement merged group array in previously assigned memory space
             os << g.getMemorySpace() << " Merged" << T::name << "Group" << g.getIndex() << " d_merged" << T::name << "Group" << g.getIndex() << "[" << g.getGroups().size() << "];" << std::endl;
+            if(!g.getFields().empty()) {
+                // Write function to update
+                os << "void pushMerged" << T::name << "Group" << g.getIndex() << "ToDevice(unsigned int idx, ";
+                g.generateStructFieldArgumentDefinitions(os, *this);
+                os << ")";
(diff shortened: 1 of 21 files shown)


tnowotny (approved):
Looks sensible.

PR #647, opened 2025-01-06
HIP

neworderofjamie:
So, HIP is a clone of CUDA with all the functions and types prefixed with ``hip`` instead of ``cuda``. This means it's much closer to CUDA than e.g. OpenCL so I have added another layer of abstraction between ``BackendSIMT`` and the CUDA backend - ``BackendCUDAHIP`` - to reflect this. The CUDA and HIP backends are then  derived off this and provide virtuals which provide the prefixes for the various libraries and do anything that actually involves interfacing with the device (getting ``cudaDeviceProp``). Other than that heavy-lifting, this PR:

* Removes OpenCL - it was never supported well enough by anyone to be worthwhile and it hadn't been updated to GeNN 5 anyway
* Does some fancy pytest stuff to the feature tests so which backends to test is determined automatically (my workstation has HIP installed and all feature tests run correctly)

This is the type of change that is going to be absolutely horrible to merge other stuff onto so, while this isn't really ready for anyone to use until I get to test on an actual AMD card, I'd like to get it merged so I can do other things. It definitely doesn't break CUDA and the HIP backend works fine with the NVIDIA backend which basically just renamed everything back.


tnowotny (approved):
Yes, this makes sense & let's hope that HIP has a future and is not replaced by yet another API ...

Issue #648, opened 2025-01-10
Replace CURAND


neworderofjamie:
As a side result from recent vectorization work, the state size of XORWOW means that it dominates the memory bandwidth requirements of simple point neuron models.

PR #649, opened 2025-01-13
Don't keep box muller transform state between kernel launches

neworderofjamie:
This builds on the  new hybrid HIP/CUDA backend from #647 so review that first!

As described in #648, beyond its (arguably excessive) 192 bit of state, the curand XORWOW RNG used to provide randomness for neuron and custom connectivity updates also stores 160 bits of [box muller transform](https://en.wikipedia.org/wiki/Box%E2%80%93Muller_transform) state in the ``curandState`` struct (BM draws two numbers and produces two normally distributed values so this state is used to cache one of those results for subsequent calls to ``curand_normal``). 

In this PR, when using CUDA and HIP backends, we create our own ``XORWowStateInternal`` struct in definitions.h without the BM state and store this in memory. At the start of the neuron and custom connectivity update kernels, we copy the fields from the  ``XORWowStateInternal`` struct into a local ``curandState`` and, at the end, we copy them back.

Excitingly, because we are very memory bandwidth bound, this makes the neuron kernel on the cortical microcircuit model about 60% faster. On my A5000 (running for 1 second):

|                                | Before[seconds] | After [seconds] |
|--------------------- |-------------------|------------------|
|neuron update        | 0.18                    | 0.11                  |
|presynaptic update | 0.23                    | 0.24                  |

### DIFF
--- a/include/genn/backends/cuda/backend.h
+++ b/include/genn/backends/cuda/backend.h
@@ -174,9 +174,6 @@ public:
                                   AtomicOperation op = AtomicOperation::ADD, 
                                   AtomicMemSpace memSpace = AtomicMemSpace::GLOBAL) const final;
 
-    //! Get type of population RNG
-    virtual Type::ResolvedType getPopulationRNGType() const final;
-
     //--------------------------------------------------------------------------
     // CodeGenerator::BackendBase virtuals
     //--------------------------------------------------------------------------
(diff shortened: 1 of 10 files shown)


tnowotny (approved):
Wow ... what a detail and what an effect. It doesn't necessarily look very elegant but the savings seem worth it!

Issue #622, opened 2024-04-25
GeNNModel constructor accepts any preference kwarg


neworderofjamie:
I think unknown preferences should be an error at https://github.com/genn-team/genn/blob/master/pygenn/genn_model.py#L659

Myrausman:
Hi @neworderofjamie, I want to work on this issue, Could you please assign it to me.

neworderofjamie:
Sure, should be a nice easy fix!

PR #663, opened 2025-03-17
Fix: Raise error for unknown preferences in GeNNModel constructor

Myrausman:
This pull request addresses issue #622 by modifying the GeNNModel constructor to raise a ValueError when an unknown preference keyword is provided.


neworderofjamie (changes requested):
Nice! However, I think it would be slightly tidier to implement this as an ``else`` clause in the identical loop below

neworderofjamie (approved):
Awesome, great work @Myrausman

Issue #669, opened 2025-04-08
Question about SpikeRecorder


neworderofjamie:
This definitely sounds like a bug - could you share your code with me? Either here or with J.C.Knight@sussex.ac.uk (although this email account, helpfully, blocks Python attachments so you'll need to rename the script to .txt or something).

### Discussed in https://github.com/orgs/genn-team/discussions/668

<div type='discussions-op-text'>

<sup>Originally posted by **OOOOOmoi** April  8, 2025</sup>
Hello, I'm using GeNN to simulate a variant of the Potjans model containing three types of inhibitory neurons, with more diverse weights between each neuronal population. When I try to use the `GeNNModel.pull_recording_buffers_from_device()` method to record neuronal spikes in the network like this:
```
model.load(num_recording_timesteps=args.buffer_size)
spike_data = {n: [] for n in neuron_populations.keys()}
while model.t < duration:
    # Advance simulation
    print(model.timestep)
    model.step_time()
    # Indicate every 10%
    if not model.timestep % args.buffer_size:
        model.pull_recording_buffers_from_device()
        for n, pop in neuron_populations.items():
            spike_times, spike_ids = pop.spike_recording_data[0]
            spike_data[n].append(np.column_stack((spike_times, spike_ids)))
    if (model.timestep % ten_percent_timestep) == 0:
        print("%u%%" % (model.timestep / 100))
```
I found the program couldn't run properly and would always get stuck at `pull_recording_buffers_from_device()`, I've already enabled spike recording when creating neurons using `neuron_pop.spike_recording_enabled = True`. I also tried reducing the number of neurons and synapses, but that didn't work. What's puzzling is that when I don't create any synapses, set all synaptic weights to zero, or make the weights between different neuronal populations follow the same mean and variance, the program runs normally. Where do you think the problem might lie?
By the way, I encountered this same issue when using GeNN 4.X before. At that time, the program also got stuck at `pull_recording_buffers_from_device()`, but I didn't pay much attention to it - I directly modified the generated C++ code and makefile to bypass the SpikeRecorder.</div>

OOOOOmoi:
You can clone my code from https://github.com/OOOOOmoi/PyGeNNProject.git. The main script is located at /SingleColumn/SingleColumn.py, which is written following the PyGeNN documentation at https://genn-team.github.io/genn/documentation/5/userproject/potjans_microcircuit.html.

[...]

PR #672, opened 2025-04-21
Fix generation of debug code in PyGeNN on Windows

neworderofjamie:
This probably affects no one aside from me but, while investigating https://github.com/genn-team/genn/issues/669, I realised passing ``debug_code=True`` to the ``GeNNModel`` constructor didn't work properly on Windows. This is because ``runner.vcxproj`` is configured to build either release or debug code but we were always selecting release when we build from PyGeNN. The runtime then fails as _it's_ looking for a debug library. This mirrors the existing logic used in the C++ code generator (https://github.com/genn-team/genn/blob/master/src/genn/generator/generator.cc#L127-L128)

### DIFF
--- a/pygenn/genn_model.py
+++ b/pygenn/genn_model.py
@@ -683,8 +683,12 @@ class GeNNModel(ModelSpec):
         # Build code
         if not never_rebuild:
             if system() == "Windows":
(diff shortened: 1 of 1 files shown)


tnowotny (approved):
That makes perfect sense.

PR #675, opened 2025-05-04
Add support for CUDA Array Interface for interoperability with other Python libraries (resolves #633)

### LINKED ISSUE #633: Support CUDA array protocol for interopability
There is a developing standard to allow python tensor-flinging libraries to exchange data (https://numba.readthedocs.io/en/stable/cuda/cuda_array_interface.html). With the new runtime, this would actually be quite simple. Would just need to expose the device pointers in the CUDA array class and add some python functionality.

JRV7903:
- Added the __cuda_array_interface__ property to the Array class to expose CUDA array data for interoperation.
- Implemented getDevicePointer() in the Array class to retrieve device pointers.
- Updated the cudaBackend.cc and backend.h files to handle the device pointers and support CUDA-specific array operations.
- Modified model_preprocessor.py to enable the inclusion of the __cuda_array_interface__ property for array interoperability.
- Created a new test file test_cuda_interface.py for verifying the changes and ensuring the correct functionality of the CUDA array interface.

### DIFF
--- a/pygenn/model_preprocessor.py
+++ b/pygenn/model_preprocessor.py
@@ -64,6 +64,33 @@ class ArrayBase:
     def _unload(self):
         self._view = None
         self._array = None
+        
+    @property
+    def __cuda_array_interface__(self):
+        """CUDA array interface property for interoperability with other Python libraries.
+        This property follows the CUDA Array Interface standard:
+        https://numba.readthedocs.io/en/stable/cuda/cuda_array_interface.html
+        """
+        if self._array is None:
+            raise RuntimeError("Array is not initialized")
+            
+        if hasattr(self._array, "_device_pointer"):
+            model = self.group._model
+            resolved_type = (self.type if isinstance(self.type, ResolvedType)
+                            else self.type.resolve(model._type_context))
+            dtype = model.genn_types[resolved_type]
+            
+            shape = self._view.shape
+            
+            return {
+                "data": (self._array._device_pointer, False),
+                "shape": shape,
+                "strides": None,
+                "typestr": dtype.str,
+                "version": 3,
+            }
+        else:
+            raise RuntimeError("Array is not a CUDA array")
 
 
 class Array(ArrayBase):
--- a/src/genn/backends/cuda/backend.cc
+++ b/src/genn/backends/cuda/backend.cc
@@ -55,187 +55,170 @@ const EnvironmentLibrary::Library doubleRandomFunctions = {
 const Type::ResolvedType CURandState = Type::ResolvedType::createValue<curandState>("curandState", false, nullptr, true);
 const Type::ResolvedType CURandStatePhilox43210 = Type::ResolvedType::createValue<curandStatePhilox4_32_10_t>("curandStatePhilox4_32_10_t", false, nullptr, true);
 
-//--------------------------------------------------------------------------
-// Array
-//--------------------------------------------------------------------------
-class Array : public Runtime::ArrayBase
-{
-public:
-    Array(const Type::ResolvedType &type, size_t count, 
-          VarLocation location, bool uninitialized)
-    :   ArrayBase(type, count, location, uninitialized), m_DevicePointer(nullptr)
-    {
-        // Allocate if count is specified
-        if(count > 0) {
-            allocate(count);
-        }
-    }
-
-    virtual ~Array()
-    {
-        if(getCount() > 0) {
-            free();
-        }
-    }
-    
-    //------------------------------------------------------------------------
-    // ArrayBase virtuals
-    //------------------------------------------------------------------------
-    //! Allocate array
-    virtual void allocate(size_t count) final
-    {
-        // Set count
-        setCount(count);
-
-        // Malloc host pointer
-        if(getLocation() & VarLocationAttribute::HOST) {
-            const unsigned int flags = (getLocation() & VarLocationAttribute::ZERO_COPY) ? cudaHostAllocMapped : cudaHostAllocPortable;
-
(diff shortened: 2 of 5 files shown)


tests/features/test_cuda_interface.py
```diff
@@ -0,0 +1,64 @@
...
+
+try:
+    import cupy as cp
+except ImportError:
+    pytest.skip("CuPy is required for this test", allow_module_level=True)
+
+from pygenn import genn_model
+
+@pytest.mark.parametrize("precision,np_dtype", [("float", np.float32), ("double", np.float64)])
+def test_cuda_array_interface(make_model, backend, precision, np_dtype):
```
neworderofjamie, line 13: there is a mechanism for selecting backends tests run for in https://github.com/genn-team/genn/blob/master/tests/features/conftest.py#L25-L53 which you could extend with a ``backend_cuda`` fixture

tests/features/test_cuda_interface.py
```diff
@@ -0,0 +1,64 @@
+import numpy as np
+import pytest
+import sys
+
+try:
+    import cupy as cp
```
neworderofjamie, line 6: it would be good to:

1. Add cupy to the list of test dependencies at https://github.com/genn-team/genn/blob/master/setup.py#L316
2. Set the CI to install the dependency at https://github.com/genn-team/genn/blob/master/Jenkinsfile#L196

src/genn/backends/cuda/backend.cc
```diff
@@ -55,187 +55,170 @@ const EnvironmentLibrary::Library doubleRandomFunctions = {
...
-class Array : public Runtime::ArrayBase
+namespace GeNN::CodeGenerator::CUDA
 {
-public:
-    Array(const Type::ResolvedType &type, size_t count, 
+Array::Array(const Type::ResolvedType &type, size_t count, 
           VarLocation location, bool uninitialized)
-    :   ArrayBase(type, count, location, uninitialized), m_DevicePointer(nullptr)
-    {
-        // Allocate if count is specified
```
neworderofjamie, line 68: Can you restore the comments?

src/genn/backends/cuda/backend.cc
```diff
@@ -55,187 +55,170 @@ const EnvironmentLibrary::Library doubleRandomFunctions = {
...
+}
 
-        if(!(getLocation() & VarLocationAttribute::ZERO_COPY)) {
-            // If end of slice overflows array, give error
-            if((offset + count) > getCount()) {
-                throw std::runtime_error("Cannot pull slice that overflows array");
-            }
+//! Copy a 1D slice of elements to device 
+/*! \param offset   Offset in elements to start copying from
+    \param count    Number of elements to copy*/
```
neworderofjamie, line 151: these comments get used to generate API documentation, can you move them to the header?

src/genn/backends/cuda/backend.cc
```diff
@@ -55,187 +55,170 @@ const EnvironmentLibrary::Library doubleRandomFunctions = {
...
+//! Serialise backend-specific host object to bytes
+void Array::serialiseHostObject(std::vector<std::byte>&, bool) const
+{
+    throw std::runtime_error("CUDA arrays have no host objects");
+}
 
+std::byte *Array::getDevicePointer() const
+{
+    return m_DevicePointer;
+}
```
neworderofjamie, line 220: I think this can probably be declared inline in the header

neworderofjamie (changes requested):
This looks really great - nice work! Just a few minor suggestions

tests/features/test_cuda_interface.py
neworderofjamie, line 13: the seperate numpy dtype is unecessary - your range will be cast to ``precision`` anyway
```suggestion
@pytest.mark.parametrize("precision", ["float", "double"])
def test_cuda_array_interface(make_model, backend, precision):
```

tests/features/test_cuda_interface.py
```diff
@@ -0,0 +1,64 @@
...
+            "C": 1.0,
+            "TauM": 20.0,
+            "Vrest": -65.0,
+            "Vreset": -70.0,
+            "Vthresh": -55.0,
+            "Ioffset": 0.0,
+            "TauRefrac": 0.0
+        }, 
+        {
+            "V": -65.0,
```
neworderofjamie, line 40: you can assign the range of initial values directly here:
```suggestion
            "V": np.linspace(-70.0, -60.0, 100),
```

pygenn/model_preprocessor.py
```diff
@@ -64,6 +64,33 @@ def pull_from_device(self):
...
+        https://numba.readthedocs.io/en/stable/cuda/cuda_array_interface.html
+        """
+        if self._array is None:
+            raise RuntimeError("Array is not initialized")
+            
+        if hasattr(self._array, "_device_pointer"):
+            model = self.group._model
+            resolved_type = (self.type if isinstance(self.type, ResolvedType)
+                            else self.type.resolve(model._type_context))
+            dtype = model.genn_types[resolved_type]
```
neworderofjamie, line 81: This is the only actual bug I found - with this it works great 😄 
```suggestion
            dtype = np.type(model.genn_types[resolved_type])
```

JRV7903:
Thank you for reviewing the PR. I’ve fixed the bug so that CuPy now correctly receives a NumPy datatype. I’ve also implemented the other requested changes.

neworderofjamie (changes requested):
Amazing! If you can accept my one suggestion, I will merge

neworderofjamie (approved):

PR #676, opened 2025-05-07
Add pygenn wrappers for get_target_var() and set_target_var() on CurrentSource

clinssen:
Add pygenn wrappers for getting and setting the target variable that current is injected into by a CurrentSource.

<s>I haven't checked that the documentation renders okay, could you please double-check?</s>

Cheers!

### DIFF
--- a/pygenn/genn_model.py
+++ b/pygenn/genn_model.py
@@ -1606,7 +1606,7 @@ def create_current_source_model(class_name: str, params: ModelParamsType = None,
     
     Finally, the function ``injectCurrent(x)`` can be used to inject a current
     ``x`` into the attached neuron. The variable it goes into can be
-    configured using the :meth:`.CurrentSource.set_target_var` method. It defaults to ``Isyn``.
+    configured using the :attr:`CurrentSource.target_var`. It defaults to ``Isyn``.
 
     Args:
         class_name:             name of the new class (only for debugging)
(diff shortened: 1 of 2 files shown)


pygenn/src/genn.cc
```diff
@@ -508,8 +508,10 @@ PYBIND11_MODULE(_genn, m)
              pybind11::arg("param_name"), pybind11::arg("dynamic") = true,
              DOC(CurrentSource, setParamDynamic))
         WRAP_METHOD("set_var_location", CurrentSource, setVarLocation)
-        WRAP_METHOD("get_var_location", CurrentSource, getVarLocation);
-    
+        WRAP_METHOD("get_var_location", CurrentSource, getVarLocation)
+        WRAP_METHOD("set_target_var", CurrentSource, setTargetVar)
+        WRAP_METHOD("get_target_var", CurrentSource, getTargetVar);
```
neworderofjamie, line 513:
```suggestion
        WRAP_PROPERTY("target_var", CurrentSource, TargetVar)
```

pygenn/genn_model.py
```diff
@@ -1606,7 +1606,7 @@ def create_current_source_model(class_name: str, params: ModelParamsType = None,
     
     Finally, the function ``injectCurrent(x)`` can be used to inject a current
     ``x`` into the attached neuron. The variable it goes into can be
-    configured using the :attr:`CurrentSource.target_var`. It defaults to ``Isyn``.
+    configured using the :meth:`.CurrentSource.set_target_var` method. It defaults to ``Isyn``.
```
neworderofjamie, line 1609:
```suggestion
    configured using :attr:`.CurrentSource.target_var`. It defaults to ``Isyn``.
```

neworderofjamie (changes requested):
I would prefer this to be exposed as properties (when in Python, make it Pythonic!) but, otherwise, excellent spot!

PR #678, opened 2025-05-10
Allow variable names to start with an underscore

clinssen:
In C variable naming, variable names may start with an underscore (see https://en.cppreference.com/w/cpp/language/identifiers). Currently, variable names starting with an underscore are not lexed correctly and result in an error during code generation. This PR fixes the issue.


neworderofjamie:
This is intentional to prevent clashes with internal code

PR #680, opened 2025-05-18
Add NVTX profiling support to CUDA backend (resolves #635)

### LINKED ISSUE #635: NVTX support for easier profiling
Currently, when profiling with NVIDIA Nsight Systems, you have to figure out what kernel launches relate to what yourself. However, it would be simple to add a flag to the CUDA backend preferences to generate NVTX ranges (https://developer.nvidia.com/blog/cuda-pro-tip-generate-custom-application-profile-timelines-nvtx/) e.g. for each simulation timestep, custom update etc

JRV7903:
- Added `enableNVTX` preference flag to the CUDA backend
- Added linking against nvToolsExt library when NVTX is enabled
- I've implemented NVTX range markers around key simulation components:
  - Overall simulation steps
  - Neuron updates
  - Synapse updates
  - Custom updates

### DIFF
--- a/include/genn/genn/code_generator/backendBase.h
+++ b/include/genn/genn/code_generator/backendBase.h
@@ -326,6 +326,15 @@ public:
     //! Get hash digest of this backends identification and the preferences it has been configured with
     virtual boost::uuids::detail::sha1::digest_type getHashDigest() const = 0;
 
+    //! Check if NVTX markers are enabled for this backend
+    virtual bool isNVTXMarkerEnabled() const { return false; }
+    
+    //! Push a named NVTX range marker for profiling
+    virtual void pushNVTXRange(const std::string &) const {}
+    
+    //! Pop the current NVTX range marker
+    virtual void popNVTXRange() const {}
+
     //--------------------------------------------------------------------------
     // Public API
     //--------------------------------------------------------------------------
--- a/include/genn/genn/code_generator/backendBase.h
+++ b/include/genn/genn/code_generator/backendBase.h
@@ -330,7 +330,7 @@ public:
     virtual void genPushProfilerRange(CodeStream&, const std::string&) const {}
     
     //! Generate code to pop the current profiler range marker
-    virtual void genPopProfilerRange(CodeStream &os) const {}
+    virtual void genPopProfilerRange(CodeStream&) const {}
 
     //--------------------------------------------------------------------------
     // Public API
--- a/include/genn/genn/code_generator/backendBase.h
+++ b/include/genn/genn/code_generator/backendBase.h
@@ -327,7 +327,7 @@ public:
     virtual boost::uuids::detail::sha1::digest_type getHashDigest() const = 0;
 
     //! Generate code to push a profiler range marker
-    virtual void genPushProfilerRange(CodeStream &os, const std::string &name) const {}
+    virtual void genPushProfilerRange(CodeStream&, const std::string&) const {}
     
     //! Generate code to pop the current profiler range marker
     virtual void genPopProfilerRange(CodeStream &os) const {}
--- a/include/genn/genn/code_generator/backendBase.h
+++ b/include/genn/genn/code_generator/backendBase.h
@@ -326,14 +326,11 @@ public:
     //! Get hash digest of this backends identification and the preferences it has been configured with
     virtual boost::uuids::detail::sha1::digest_type getHashDigest() const = 0;
 
-    //! Check if NVTX markers are enabled for this backend
-    virtual bool isNVTXMarkerEnabled() const { return false; }
+    //! Generate code to push a profiler range marker
+    virtual void genPushProfilerRange(CodeStream &os, const std::string &name) const {}
     
-    //! Push a named NVTX range marker for profiling
-    virtual void pushNVTXRange(const std::string &) const {}
-    
-    //! Pop the current NVTX range marker
-    virtual void popNVTXRange() const {}
+    //! Generate code to pop the current profiler range marker
+    virtual void genPopProfilerRange(CodeStream &os) const {}
 
     //--------------------------------------------------------------------------
     // Public API
--- a/src/genn/backends/cuda/backend.cc
+++ b/src/genn/backends/cuda/backend.cc
@@ -507,6 +507,9 @@ std::string Backend::getNVCCFlags() const
     if(getPreferences<Preferences>().generateLineInfo) {
         nvccFlags += " --generate-line-info";
     }
+    if(getPreferences<Preferences>().enableNVTX) {
+        nvccFlags += " -lnvToolsExt";
+    }
 
     return nvccFlags;
 }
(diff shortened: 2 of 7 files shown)


src/genn/backends/cuda/backend.cc
```diff
@@ -534,6 +537,18 @@ void Backend::genDefinitionsPreambleInternal(CodeStream &os, const ModelSpecMerg
...
+    
+    // If NVTX profiling is enabled, include nvToolsExt header
+    if(getPreferences<Preferences>().enableNVTX) {
+        os << "#include <nvToolsExt.h>" << std::endl;
+        
+        // Add helper macros for NVTX ranges
+        os << std::endl;
+        os << "// NVTX profiling macros" << std::endl;
+        os << "#define NVTX_PUSH(name) nvtxRangePushA(name)" << std::endl;
+        os << "#define NVTX_POP() nvtxRangePop()" << std::endl;
```
neworderofjamie, line 549: These macros don't really seem to be necessary - they don't save any typing and code generation is handling turning off profiling i.e. you don't just disable it by:
```c++
#define NVTX_PUSH(name)
#define NVTX_POP()
```

include/genn/genn/code_generator/backendBase.h
```diff
@@ -326,6 +326,15 @@ class GENN_EXPORT BackendBase
...
 
+    //! Check if NVTX markers are enabled for this backend
+    virtual bool isNVTXMarkerEnabled() const { return false; }
+    
+    //! Push a named NVTX range marker for profiling
+    virtual void pushNVTXRange(const std::string &) const {}
+    
+    //! Pop the current NVTX range marker
+    virtual void popNVTXRange() const {}
+
```
neworderofjamie, line 337: I think you are right that this needs adding to the ``BackendBase`` interface _but_ NVTX is a CUDA-specific technology. I think the virtuals you should define are:
```c++
virtual void genPushProfilerRange(CodeStream &os, const std::string &name) const {}
virtual void genPopProfilerRange(CodeStream &os) const {}
```
which, in the CUDA backend, will check the preferences flag and then write ``nvtxRangePushA(name)`` to ``os`` (I don't see you need to expose the test for whether profiling ranges are enabled)

src/genn/genn/code_generator/generateRunner.cc
```diff
@@ -269,25 +269,47 @@ void GeNN::CodeGenerator::generateRunner(const filesystem::path &outputPath, Mod
...
+    // Function to step time in the simulation
     runner << "void stepTime(unsigned long long timestep, unsigned long long numRecordingTimesteps)";
     {
         CodeStream::Scope b(runner);
 
+        // Add NVTX range if NVTX is enabled
+        if(dynamic_cast<const BackendCUDAHIP&>(backend).getPreferences<BackendCUDAHIP::PreferencesCUDAHIP>().enableNVTX) {
+            runner << "NVTX_PUSH(\"stepTime\");" << std::endl;
+        }
+
```
neworderofjamie, line 281: ``dynamic_cast`` is _often_ a sign of poor design in C++ and this one of those times. The code generator is intended to be totally backend-agnostic so we really don't want to introduce dependencies on ``BackendCUDAHIP``. With my suggested API, this would just be replaced by:

```c++
backend.genPushProfilerRange(runner, "stepTime (" + std::to_string(timestep) + ")");
```
(this applies to all the instances below)

src/genn/genn/runtime/runtime.cc
```diff
@@ -700,6 +700,11 @@ void Runtime::stepTime()
 //----------------------------------------------------------------------------
 void Runtime::customUpdate(const std::string &name)
 {
+    // If NVTX markers are enabled, push range
+    if(m_Backend.get().isNVTXMarkerEnabled()) {
+        m_Backend.get().pushNVTXRange("customUpdate:" + name);
+    }
+
```
neworderofjamie, line 707: I think profiling ranges should be inserted in the generated code as otherwise GeNN itself will need to be linked against NVTX. This functions to push and pop ranges will need calling in https://github.com/genn-team/genn/blob/master/src/genn/backends/single_threaded_cpu/backend.cc#L666 and https://github.com/genn-team/genn/blob/master/src/genn/genn/code_generator/backendCUDAHIP.cc#L756

src/genn/backends/cuda/backend.cc
```diff
@@ -507,6 +507,9 @@ std::string Backend::getNVCCFlags() const
     if(getPreferences<Preferences>().generateLineInfo) {
         nvccFlags += " --generate-line-info";
     }
+    if(getPreferences<Preferences>().enableNVTX) {
+        nvccFlags += " -lnvToolsExt";
+    }
```
neworderofjamie, line 512: According to the documentation, NVTX has been header-only since CUDA 10 so this is unnecessary (which is good because I can't figure out the windows equivalent!). Would be good to add an error check for CUDA < 10 and this being enabled in the CUDA backend constructor though.

src/genn/backends/cuda/backend.cc
```diff
@@ -534,6 +537,18 @@ void Backend::genDefinitionsPreambleInternal(CodeStream &os, const ModelSpecMerg
     if(getRuntimeVersion() >= 9000) {
         os <<"#include <cuda_fp16.h>" << std::endl;
     }
+    
+    // If NVTX profiling is enabled, include nvToolsExt header
+    if(getPreferences<Preferences>().enableNVTX) {
+        os << "#include <nvToolsExt.h>" << std::endl;
```
neworderofjamie, line 543: As it's header only and thus easier to build, you should switch the NVTX3 which is included with:
```c++
#include "nvtx3/nvToolsExt.h"
```

neworderofjamie (changes requested):
Nice work! This all seems sensible but I have some suggestions to tidy it up. Also, you should expose your new preference to PyGeNN in https://github.com/genn-team/genn/blob/master/pygenn/src/cudaBackend.cc#L73

src/genn/genn/code_generator/generateRunner.cc
JRV7903: I have implemented the backend-agnostic approach here using `genPushProfilerRange` as you suggested. However, I noticed that including the timestep in the profiling label `"stepTime (" + std::to_string(timestep) + ")"` causes a build error since timestep is not available in that scope. The build works well with just the static string `stepTime`. Would you like me to keep the static string implementation or should i move the profiling calls to a scope where timestep is available?

neworderofjamie: This is a very good point - what you would actually want to do is include the value of the timestep variable in the generated code. i.e. generate code that does the ``std::to_string(timestep) `` bit. However, I can't think of a sensible API for this right now so maybe just switch it back to the static string for now.

neworderofjamie (changes requested):
This looks great - well done! If you could just switch back to the static string for stepTime then I'll test and merge

JRV7903:
Thankyou! I've reverted to the static string and the build tests are working fine. I noticed you have resolved the profiling consistency improvement suggestion so that's the only one that hasn't been implemented.

include/genn/genn/code_generator/backendBase.h
```diff
@@ -326,6 +326,12 @@ class GENN_EXPORT BackendBase
     //! Get hash digest of this backends identification and the preferences it has been configured with
     virtual boost::uuids::detail::sha1::digest_type getHashDigest() const = 0;
 
+    //! Generate code to push a profiler range marker
+    virtual void genPushProfilerRange(CodeStream &os, const std::string &name) const {}
```
neworderofjamie, line 330:
```suggestion
    virtual void genPushProfilerRange(CodeStream&, const std::string&) const {}
```
to fix warnings in GCC

include/genn/genn/code_generator/backendBase.h
```diff
@@ -326,6 +326,12 @@ class GENN_EXPORT BackendBase
...
+    
+    //! Generate code to pop the current profiler range marker
+    virtual void genPopProfilerRange(CodeStream &os) const {}
```
neworderofjamie, line 333:
```suggestion
    virtual void genPopProfilerRange(CodeStream&) const {}
```
to fix warnings in GCC

neworderofjamie (changes requested):
Just a couple of suggestions to fix warnings in GCC and _then_ I will definitely merge :)

PR #693, opened 2025-07-07
More flexible file generation

neworderofjamie:
Previous Backends could only make one (.cc) file for each module (neuron update, synapse update, init, custom update) they generated. However, the ISPC compiler _only_ compiles parallel ISPC code so generated ISPC code needs to go into a seperate .cc file and  the C++ helper functions to push and pull variable needs to remain in the .cc. In this PR, ``generateAll`` (and the CUDA optimiser) provide functions to create new streams with associated filenames to write them to and these get passed to the backends. In ``generateAll``, this function encompasses the logic which only actually writes the streams if model hashes change whereas in the CUDA optimiser (which generates temporary modules for testing different block sizes), files are created directly.


tnowotny (approved):
Ok, makes sense in the context of the ISPC backend development.

PR #696, opened 2025-07-09
Scanner fix

### LINKED ISSUE #692: Badly formed code strings can crash scanner
``scanString`` definitely has this problem. In the following (https://github.com/genn-team/genn/blob/master/src/genn/genn/transpiler/scanner.cc#L255-L257):

```c++
 while(scanState.peek() != '"') {
    scanState.advance();
}
```

``ScanState::peek`` will return ``'\0'`` if EOF is encountered which ``!= '"'`` so the scanner will try and advance past EOF

neworderofjamie:
The scanner is the bit of the transpiler which turns a code string into a sequence of token objects (which is a much nicer thing to work with). It's pretty much a C++ port of the one from https://craftinginterpreters.com/scanning.html but I failed to implement scanning of string literals properly. This means un-terminated strings would crash the scanner. This PR adds a unit test and fixes the issue.

Fixes #692

### DIFF
--- a/src/genn/genn/transpiler/scanner.cc
+++ b/src/genn/genn/transpiler/scanner.cc
@@ -250,11 +250,25 @@ void scanNumber(char c, ScanState &scanState, std::vector<Token> &tokens)
 //---------------------------------------------------------------------------
 void scanString(ScanState &scanState, std::vector<Token> &tokens)
 {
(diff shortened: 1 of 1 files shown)


tnowotny (approved):
Such a simple thing ... fix looks good.

PR #700, opened 2025-08-06
Support NMake build system for building generated code

neworderofjamie:
We switched from using NMake to MSBuild to build generated code a long time ago as it's a better build system (supports dependency management and parallel building) **but**:

1. If you install the CUDA toolkit using Conda, the plugin required to build CUDA does not get installed
2. Neither ISPC or HIP provides MSBuild plugins.

Therefore, in this PR, I have resurrected support for building generated code using the archaic NMAKE build system.


tnowotny (approved):
I wish this could be a little less messy, but I appreciate the trade-off between the better build system (MSBuild) and the arguably more universal one (NMake) ... in terms of details, I need to trust you that this works.

neworderofjamie:
It's a very annoying trade-off - I would just say to get rid of MSBuild for building generated code but NMake doesn't seem to track dependencies and that makes it way slower to build generated code

PR #713, opened 2025-09-15
HIP AMD

neworderofjamie:
The git branches/forks got a bit messed up but this is a joint effort by me and @HengyeZhu to turn our _theoretical_ HIP support (#647) into the actual ability to run on AMD cards! @HengyeZhu has fixed up the build system and I've made various RNG fixes.

HIP provides various libraries ``hipRand``, ``hipRand`` etc which aim to emulate the CUDA equivalents. However, the actual implementations they use behind the CUDA-like API can be very different. We have hit this with cuRand vs hipRand where, in cuRand, the state of an RNG is a plain struct e.g. for XORWOW:
```c++
struct curandStateXORWOW {
    unsigned int d, v[5];
    int boxmuller_flag;
    int boxmuller_flag_double;
    float boxmuller_extra;
    double boxmuller_extra_double;
};
```
whereas, in rocRand, the lower level library hipRand calls gets dispatched to on AMD,  the implementation is a lot more C++ (https://github.com/ROCm/rocm-libraries/blob/develop/projects/rocrand/library/include/rocrand/rocrand_xorwow.h#L69). This has two connotations for things we do in GeNN:

1. You can't create classes with ``__device__ hiprandState d_rng;`` as that would mean the constructor needs to be run on device when the kernel is loaded which is understandably not possible
2. You can't freely hack with the internals of the RNG state to save memory bandwidth (#649) as they don't even provide a getter for the state struct which, irritatingly, is sitting there in ``m_state`` protected member.

This PR solves the first issue by turning the global RNG into a pointer, cuda/hip mallocing it and copying it to the device symbol and the second by disabling the optimisation when using the HIP backend on AMD.


src/genn/genn/code_generator/backendCUDAHIP.cc
```diff
@@ -255,8 +255,15 @@ void BackendCUDAHIP::genWarpReduction(CodeStream& os, const std::string& variabl
                                       VarAccessMode access, const Type::ResolvedType& type) const
 {
     for (unsigned int i = 16; i > 0; i /= 2) {
+#if defined(__HIP_PLATFORM_NVIDIA__)
```
neworderofjamie, line 258: I don't think this is the correct fix. Can you try something like:

```c++
std::string mask;
if(getNumLanes() == 32) {
    mask = "0xFFFFFFFF";
}
else {
    assert(getNumLanes() == 64);
    mask = "0xFFFFFFFFFFFFFFFFull";
}
for (unsigned int i = (getNumLanes() / 2); i > 0; i /= 2) {
        os << getReductionOperation(variable, "__shfl_down_sync(" + mask + ", " + variable + ", " + std::to_string(i) + ")",
                                    access, type);
        os <<  ";" << std::endl;
    }
```

src/genn/genn/code_generator/backendCUDAHIP.cc
```diff
@@ -1515,7 +1519,14 @@ void BackendCUDAHIP::genReturnFreeDeviceMemoryBytes(CodeStream &os) const
 //--------------------------------------------------------------------------
 void BackendCUDAHIP::genAssert(CodeStream &os, const std::string &condition) const
 {
+#if defined(__HIP_PLATFORM_NVIDIA__)
+    // Use standard assert for CUDA backend (HIP on NVIDIA)
     os << "assert(" << condition << ");" << std::endl;
+#else
+    // Use HIP-compatible assertion for AMD backend that doesn't cause runtime state issues
+    // Simply return early instead of using problematic assert/abort
+    os << "if(!(" << condition << ")) { return; }" << std::endl;
```
neworderofjamie, line 1528: Asserts are used for signalling serious errors so just disabling them isn't a good solution - I would rather the test fails because the kernels doesn't exit cleanly than asserts be ignored.

src/genn/backends/hip/backend.cc
```diff
@@ -382,8 +443,15 @@ void Backend::genLazyVariableDynamicAllocation(CodeStream &os, const Type::Resol
...
-    std::string linkFlags = "--shared -arch " + architecture;
+    std::string architecture;
+#if defined(__HIP_PLATFORM_NVIDIA__)
+    architecture = "sm_" + std::to_string(getChosenHIPDevice().major) + std::to_string(getChosenHIPDevice().minor);
+#elif defined(__HIP_PLATFORM_AMD__)
+    // Get AMD GPU architecture directly from device properties
+    architecture = std::string(getChosenHIPDevice().gcnArchName);
+#endif
+    
+    std::string linkFlags = "--shared --offload-arch=" + architecture + " -fgpu-rdc";
```
neworderofjamie, line 454: --offload-arch is getting passed to NVCC when running on NVIDIA which doesn't compile:
```suggestion
    std::string linkFlags = "--shared -fgpu-rdc";
#if defined(__HIP_PLATFORM_NVIDIA__)
    architecture = "sm_" + std::to_string(getChosenHIPDevice().major) + std::to_string(getChosenHIPDevice().minor);
    linkFlags += " -arch " + architecture;
#elif defined(__HIP_PLATFORM_AMD__)
    // Get AMD GPU architecture directly from device properties
    architecture = std::string(getChosenHIPDevice().gcnArchName);
    linkFlags += " --offload-arch=" + architecture;
#endif
```

src/genn/genn/code_generator/backendCUDAHIP.cc
HengyeZhu: Okay. But HIP does have the [`__shfl_down_sync` function](https://rocm.docs.amd.com/projects/HIP/en/latest/how-to/hip_cpp_language_extensions.html).

HengyeZhu: This is likely the root cause of the error: **All mask values returned or accepted by these built-ins are 64-bit unsigned integer values, even when compiled for a device with 32 threads per warp. On such devices the higher bits are unused. CUDA code ported to HIP requires changes to ensure that the correct type is used.**

PR #718, opened 2025-10-07
AMD HIP support

### LINKED ISSUE #674: HIP on AMD hardware
As of #647 GeNN has a HIP backend but we do not have any suitable AMD hardware so this currently has only been tested and built on the NVIDIA HIP passthrough. Any help testing or working on this would be much appreciated!

neworderofjamie:
The git branches/forks got a bit messed up but this is a joint effort by @HengyeZhu and I to turn our _theoretical_ HIP support (#647) into the actual ability to run on AMD cards! @HengyeZhu has fixed up the build system and there were a couple of more 'interesting' issues:

### Random number generators
HIP provides various libraries ``hipRand``, ``hipRand`` etc which aim to emulate the CUDA equivalents. However, the actual _implementations_ they use behind the CUDA-like API can be very different. We have hit this with cuRand vs hipRand where, in cuRand, the state of an RNG is a plain struct e.g. for XORWOW:
```c++
struct curandStateXORWOW {
    unsigned int d, v[5];
    int boxmuller_flag;
    int boxmuller_flag_double;
    float boxmuller_extra;
    double boxmuller_extra_double;
};
```
whereas, in rocRand, the lower level library hipRand calls gets dispatched to on AMD,  the implementation is a lot more C++ (https://github.com/ROCm/rocm-libraries/blob/develop/projects/rocrand/library/include/rocrand/rocrand_xorwow.h#L69). This has two connotations for things we do in GeNN:

1. You can't create classes with ``__device__ hiprandState d_rng;`` as that would mean the constructor needs to be run on device when the kernel is loaded which is understandably not possible
2. You can't freely hack with the internals of the RNG state to save memory bandwidth (#649) as they don't even provide a getter for the state struct which, irritatingly, is sitting there in ``m_state`` protected member.

This PR solves the first issue by turning the global RNG into a pointer, cuda/hip mallocing it and copying it to the device symbol and the second by disabling the optimisation when using the HIP backend on AMD.

### Warp sizes
NVIDIA GPUs have always had a warp size of 32 (and, at this point, presumably always will). However, AMD GPUs have either 32 (@HengyeZhu 's AMD Radeon RX 7900 XTX) or 64 (MI300A). GeNN only descends to warp level programming in a few places but there were numerous hard-coded 32s and some annoyance about the shuffle functions in HIP taking 64-bit rather than 32-bit masks. However, more fundamentally, because we haven't yet implemented all the fancy block size optimisers, block sizes are set to the manual default which used to be 32. On devices with a warp size of 64, this is inefficient **and** breaks warp shuffles so the preferences now default these to zero and the optimiser replaces zeros with the actual warp size reported by the device.

This is now all working with the GeNN test suite and various models running successfully on the MI300A node very kindly provided by the [HPC hardware lab at Durham](https://durham.readthedocs.io/en/latest/hardwarelab/index.html)

Fixes #674

### DIFF
--- a/include/genn/backends/hip/backend.h
+++ b/include/genn/backends/hip/backend.h
@@ -76,6 +76,7 @@ struct Preferences : public PreferencesCUDAHIP
     //BlockSizeSelect blockSizeSelectMethod = BlockSizeSelect::OCCUPANCY;
 
     //! If block size select method is set to BlockSizeSelect::MANUAL, block size to use for each kernel
+    /*! These default to zero which signals the HIP backend to replace them with the device warp size. */
     KernelBlockSize manualBlockSizes;
 
     //! How much constant cache is already used and therefore can't be used by GeNN?


tnowotny (approved):
Looks good. Somewhat annoying issues to contend with, but in the end, no huge code changes were needed to fix it!

PR #733, opened 2026-01-21
Fix incomplete CUDA_PATH error message

Atharv-K-979:
This PR fixes a truncated error message thrown when the CUDA_PATH environment
variable is not set.

Previously, the message ended mid-sentence and did not provide actionable
guidance. This change completes the message and clarifies how users can
resolve the issue.

No functional or behavioral changes.


neworderofjamie (changes requested):
Thanks! Just some suggested white-space changes

PR #734, opened 2026-01-21
tests: add regression test for presynaptic output with procedural connectivity

Atharv-K-979:
### Summary
Add a basic regression test for procedural sparse connectivity used together
with kernel-based presynaptic output (`addToPre`).

### Details
This test sets up a minimal model using procedural connectivity and a simple
weight update emitting presynaptic output. The model is built, loaded, and
stepped once to make sure this configuration continues to work.

### Testing
- Ran locally with `pytest`
- Checked for both `Float` and `Double` precisions

### Scope
Test-only change. No functional or backend changes.

### DIFF
--- a/tests/features/test_presynaptic_kernel_procedural.py
+++ b/tests/features/test_presynaptic_kernel_procedural.py
@@ -1,0 +1,153 @@
+import numpy as np
+import pytest
+from pygenn import types
+from pygenn import VarAccess, VarAccessMode
+from pygenn import (
+    create_neuron_model,
+    create_var_ref,
+    create_weight_update_model,
+    init_postsynaptic,
+    init_weight_update,
+)
+
+# Neuron model which does nothing
+empty_neuron_model = create_neuron_model("empty")
+
+spike_event_source_array_model = create_neuron_model(
+    "spike_event_source_array",
+    sim_code=
+    """
+    if(startSpike != endSpike && t >= spikeTimes[startSpike]) {
+        output = true;
+        startSpike++;
+    }
+    else {
+        output = false;
+    }
+    """,
+    vars=[
+        ("startSpike", "unsigned int"),
+        ("endSpike", "unsigned int", VarAccess.READ_ONLY_DUPLICATE),
+        ("output", "bool"),
+    ],
+    extra_global_params=[("spikeTimes", "scalar*")],
+)
+
+static_event_pulse_model = create_weight_update_model(
+    "static_event_pulse",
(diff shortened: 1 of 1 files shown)


neworderofjamie:
What exactly are you trying to test here? I don't see any procedural connectvity or any presynaptic output!

Atharv-K-979:
Thanks for pointing this out — that’s fair.

The intent of the test is fairly minimal and probably not obvious from a quick read. The procedural connectivity is coming from the use of `create_sparse_connect_init_snippet` with an explicit loop over `num_post` in the row build code (creating diagonal connections based on `id_pre`). There’s no fixed connectivity initialiser involved.

The presynaptic output is exercised via the weight update model using `addToPre(g)` in `pre_spike_syn_code`. The test doesn’t currently assert on the accumulated value — it just builds, loads, and steps once to make sure this combination of procedural connectivity and presynaptic output continues to compile and run.

If you think this would be clearer with an explicit comment or a small assertion to make the intent more obvious, I’m happy to update the test accordingly.

neworderofjamie:
Why do you think this needs testing? Coverage suggests that procedural + presynaptic output it is already covered (https://app.codecov.io/github/genn-team/genn/commit/c448253463c913f3a116dcdb85d45cf9244df21b/blob/src/genn/genn/code_generator/presynapticUpdateStrategySIMT.cc?dropdown=coverage#L557)

But, more fundamentally:

> The procedural connectivity is coming from the use of create_sparse_connect_init_snippet with an explicit loop over num_post in the row build code (creating diagonal connections based on id_pre). There’s no fixed connectivity initialiser involved.

I don't see this anywhere in your PR. You are using a "DENSE" connection

> The presynaptic output is exercised via the weight update model using addToPre(g) in pre_spike_syn_code. The test doesn’t currently assert on the accumulated value — it just builds, loads, and steps once to make sure this combination of procedural connectivity and presynaptic output continues to compile and run.

It doesn't though. the ``pre_spike_syn_code`` calls ``addToPost``.

Atharv-K-979:
Thanks for the detailed clarification — that makes sense.

You’re right: the test as written doesn’t actually exercise procedural sparse
connectivity or presynaptic output, and the coverage link you shared shows that
the intended code path is already covered.

I’ll take this as a lesson to be more careful about aligning test intent with
the exact backend paths being exercised.

neworderofjamie:
Great, there are several issues about testing you could look at but also, I can believe there are gaps in coverage it would be great to cover with tests (either unit or feature)

PR #737, opened 2026-01-22
tests: check error when pulling recording buffers without recording enabled

Atharv-K-979:
This change extends the existing recording tests to explicitly cover the
case where recording is not enabled.

Currently, calling `pull_recording_buffers_from_device()` without enabling
spike or spike-event recording raises an error. This behaviour is intentional,
but was not previously asserted by the test suite.

The new test verifies that this error is raised, ensuring the API contract
remains clear and does not silently change in future.

Test-only change.


Atharv-K-979:
Hi, just a gentle follow-up on this PR whenever you have time. Please let me know if any changes are needed from my side. Thanks!

PR #739, opened 2026-02-08
Tests: cover event recording disabled behavior

Atharv-K-979:
This updates the existing recording tests to explicitly cover the case
where spike-event recording is not enabled.

Attempting to access event recording buffers without enabling recording
currently raises an error; this test asserts that behavior so it does not
regress in future.

Test-only change.


Atharv-K-979:
Hi @neworderofjamie ,
I just wanted to check in on this PR when convenient.
It adds explicit test coverage for the disabled event-recording case to prevent regressions.
Please let me know if you’d like any changes or additional coverage. Thanks!

PR #741, opened 2026-02-08
tests: simplify assertions in test_num

Atharv-K-979:
This cleans up assertions in test_num by replacing conditional
assert False patterns with direct asserts.

No behavior changes — tests still validate the same conditions,
but failures are clearer and more idiomatic.


neworderofjamie (approved):
Much tidier, thanks!

PR #742, opened 2026-02-20
Spike count reset

neworderofjamie:
#634 provided a pretty generic framework for adding variable references to GeNN internals which made this quite easy. There are some quirks:

* You need to create the reference using the synapse group the spikes go into
* At the point where variable reference types are checked, the ability to fully resolve types is not available so you have to use "uint32_t" as the type (aliases like "unsigned int" or ``types.Uint32`` won't work)


tnowotny (approved):
As much as I understand it, it looks good. Was instructive to read the test in detail.

PR #743, opened 2026-03-13
NCCL fixes

neworderofjamie:
I left a stray "N" in place when making some macro names HIP/CUDA agnostic but, otherwise, NCCL still works. While tidying this I also removed a copy of the CHECK_NCCL_ERRORs macro in utils.h because the backend does not directly make any NCCL calls - they all happen in generated code.


tnowotny (approved):
Seems pretty obvious - approved.

PR #744, opened 2026-03-19
Tweaked error message

neworderofjamie:
It's fine to attach custom updates to synapse groups with procedural _connectivity_, just not procedural _weights_ i.e. they should work fine with ``SynapseMatrixType::PROCEDURAL_KERNELG`` as these have a perfectly valid kernel of weights


tnowotny (approved):
Sure ... makes sense, I think.

Issue #745, opened 2026-03-24
[Suggestion] Update installation instructions


danilo-bc:
First of all, thanks for the work on the repository! I got curious to try it after the recent RISC-V presentation, but I found some problems and wanted to share them and how I'd change the README to better guide new users.

Apparently, CUDA and HIP stacks have recently been updated and somethings may have become incompatible. GeNN 5.3 and 5.4 also seem to have important changes, but the README still points to the 5.3 release.

System info:
- Windows 11
- RTX 3050 mobile
- Python 3.12
- Both PowerShell and Ubuntu 24.04 WSL-2 shells tried

List of things that failed, and then solutions:
- CUDA needs to be version 12.x, there's no mention in the README.
    - I tried with 13.2, which failed (segfaults), then went with 12.9 and all worked fine.
    - From "nvidia-smi" in WSL: NVIDIA-SMI 595.54, Driver Version: 595.79, CUDA Version: 13.2 (support from Driver, toolkit installed is 12.9.1)
- VisualStudio compilers are not put into Windows' PATH by default, so you need to open "Developer PowerShell for VS" to make cl.exe easily available.
    - I couldn't get this environment to work. Visual Studio 2026 PowerShell 18.4.1, CUDA Toolkit 12.9.1. 
    - The pip install build fails with linker errors ("unresolved external symbol") I couldn't decipher
- I tried the HIP path with another PC with an AMD card. Apparently, ROCm changed the file structure and, while trying to use ROCm 7.2.0, I got an error about not finding a .h file.
- I haven't faced problems with NumPy, but I believe it's use in the project. By default, pip will download numpy 2.x, but the latest stable numpy 1 version is 1.26.4.

Here's my review of the README markdown up to "Creating an editable install with pip" with a few of these points:

```markdown
# GPU-enhanced Neuronal Networks (GeNN)

GeNN is a GPU-enhanced Neuronal Network simulation environment based on code generation for NVIDIA CUDA and AMD HIP.

## Installation

### Pre-installation

#### C/C++ Compiler and other minor system dependencies

Install the C++ compiler on the machine, if not already present.

##### Windows with PowerShell

- Download Visual Studio 2019 or above. The Microsoft Visual Studio 
    Community Edition can be downloaded from
    https://www.visualstudio.com/en-us/downloads/download-visual-studio-vs.aspx.
- Select the 'Desktop development with C++' configuration during the installation process.

##### Ubuntu Linux or Windows with Ubuntu 24.02 in WSL 2

[...]

PR #746, opened 2026-03-27
Readme improvement

neworderofjamie:
Some readme improvements suggested by @danilo-bc in #745. I think a bit more structure is a good thing! @tnowotny could be a good opportunity to write some Mac instructions.


tnowotny (approved):
Looks a bit tidier. Thank you.

PR #755, opened 2026-05-29
Fix accessing presynaptic variable references in weight update models with Toeplitz connectivity

neworderofjamie:
The two stage code generation for this is a bit gnarly but the weight update model bit was expecting a variable called ``ipre`` to be provided by the connectivity generation bit which wasn't doing that. This PR adds this intermediate variable - this probably improves quality of generated code as well as otherwise the shared memory array will be de-referenced numerous times. _Theoretically_ it would be nice to do this lazily but:

1. I don't think this is possible in the two stage code generation process used here
2. By their nature, all topelitz connectivity updates need to use ``id_pre`` so it's kinda irrelevant


tnowotny (approved):
Looks like a straightforward issue and fix. Thanks.

PR #757, opened 2026-06-03
Toeplitz fixes

neworderofjamie:
@tnowotny found that the indices being used when accessing batched postsynaptic and kernel variables from weight update models on synapse groups with toeplitz connectivity were incorrect. All this indexing is handled in the same place for all types of connectivity so this was very strange until I realised a batch size of 1 was always being passed to the function that generated this code :face_with_open_eyes_and_hand_over_mouth: 

In the course of looking at the generated code, also realised that the derived parameters of the built-in Toeplitz models don't have their type configured correctly so fixed that.


tnowotny (approved):
Small but extremely important fix! Seems to make convolutional networks functional for Eventprop learning!

PR #758, opened 2026-06-08
Fixed bug in initialisation of sparse connectivity from convolutions

neworderofjamie:
Code generation was failing when sparse connectivity was initialised from some sort of kernel-based initialiser (e.g. ``Conv2D``) and the resultant kernels merged multiple groups with different kernel sizes. This was due to a missing ``_kernel_size``. 

The code for handling kernel sizes was also very manual and thus overly complex so tidied this up using the (relatively new) functionality from #646 (as past me suggested in **TODO** comments 😉 )


tnowotny (approved):
Looks good and seems to work.

